PR #1988, opened 2014-03-26
resolves #1904, fixes #1901

### LINKED ISSUE #1901: Symphony does not work with Cyrillic domains
When I try to sort the order of pages in the admin menu, displaying the error sorting. The logs shows the following entry:

17.12.2013 09:57 > UNKNOWN: SymphonyErrorPage 0 - Запрашиваемая страница не существует. on line 650 of /var/www/motochel/data/www/xn--90ahkicelda9ae5d0f.xn--p1ai/symphony/lib/core/class.symphony.php

jensscherbl:
Doesn't look like much, but I actually spent hours debugging and researching this... :/

Apparently, webkit and non-webkit browsers [behave differently](https://bugzilla.mozilla.org/show_bug.cgi?id=743488) when it comes to internationalized domain names, so we shouldn't rely on `window.location.href` in javascript for routing etc.

Relying on string length is also a bad idea when dealing with utf8 encoded strings.

Regarding cyrillic domains (#1901): Shouldn't be a problem now for Symphony (works in Firefox), but are not supported in Safari for [security reasons](http://support.apple.com/kb/TA22996), so you shouldn't use cyrillic domains if you want to avoid much bigger problems.


CarAuto34:
Ty

PR #2015, opened 2014-04-03
PDO implementation

brendo:
This PR adds PDO support to Symphony, replacing the `mysqli` driver with a PDO driver.

This is based on work that @nils-werner did years ago and I updated a couple of months ago (I've squashed all the commits down into a single commit so for easy merging).

@jensscherbl, @designermonkey, @lewiswharf, @iwyg, @nitriques


nitriques:
:+1:

PR #2106, opened 2014-06-22
ensure sane handles

jensscherbl:
All kinds of special characters that are not explicitly replaced as transliterations are currently allowed in handles, which sometimes leads to strange side effects. Proposing a saner approach here.

### DIFF
--- a/symphony/lib/toolkit/class.general.php
+++ b/symphony/lib/toolkit/class.general.php
@@ -1263,65 +1263,6 @@ class General
     }
 
     /**
-     * Truncate a string to a given length. Newlines are replaced with `<br />`
-     * html elements and html tags are removed from the string. If the resulting
-     * string contains only spaces then null is returned. If the resulting string
-     * is less than the input length then it is returned. If the option to
-     * truncate the string to a space character is provided then the string is
-     * truncated to the character prior to the last space in the string. Words
-     * (contiguous non-' ' characters) are then removed from the end of the string
-     * until the length of resulting string is within the input bound. Initial
-     * and trailing spaces are removed. Provided the user requested an
-     * ellipsis suffix and the resulting string is shorter than the input string
-     * then the ellipses are appended to the result which is then returned.
-     *
-     * @param string $string
-     *  the string to truncate.
-     * @param integer maxChars (optional)
-     *  the maximum length of the string to truncate the input string to. this
-     *  defaults to 200 characters.
-     * @param boolean $appendHellip (optional)
-     *  true if the ellipses should be appended to the result in circumstances
-     *  where the result is shorter than the input string. false otherwise. this
-     *  defaults to false.
-     * @return null|string
-     *  if the resulting string contains only spaces then null is returned. otherwise
-     *  a string that satisfies the input constraints.
-     */
-    public static function limitWords($string, $maxChars = 200, $appendHellip = false)
-    {
-        if ($appendHellip) {
-            $maxChars -= 1;
-        }
-
-        $string = trim(strip_tags(nl2br($string)));
-        $original_length = strlen($string);
-
-        if ($original_length == 0) {
-            return null;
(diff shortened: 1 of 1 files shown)


brendo:
This can be rebased and resubmitted for merging now

jensscherbl:
> the new handle function is createHandleOptimised and it uses the new function truncateHandle.

I found the `truncateHandle` function a bit hard to understand, so I went with a simpler approach.

> it now includes code to allow the handlizer to be configured

Since we're now aiming for a `3.0.0` release, this is no longer necessary and I simply changed the existing function.

> I created a simple benchmark to compare the old and new versions

Could you please update and re-run your benchmark with my new version?

jensscherbl:
Regarding migration, I guess we need a way to update existing handles in the database with the new function. For the most part, there won't be any differences between old and new handles. But some might  still contain nasty characters that are no longer allowed for new handles.

brendo:
The only way to update handles in the database is to blanket assume that every single column in the Symphony schema (`tbl_`) that is named `handle` was produced by Symphony. Are we ok with that?

Do we actually have any concrete examples of handles before and after this change? Or is this 'in theory the code allows it'?

nitriques:
>  Are we ok with that?

I am!

jensscherbl:
> Are we ok with that?

I'm pretty much in the dark regarding possible side effects, so I really can't make an informed decision here. I trust you with this.

brendo:
> I'm pretty much in the dark regarding possible side effects, so I really can't make an informed decision here. I trust you with this.

It means if a field or extension also has a `handle` column, but it's not actually been generated by Symphony, then potentially we will erase/modify data. It also depends on the answer to my previous question

> Do we actually have any concrete examples of handles before and after this change? Or is this 'in theory the code allows it'?

Do we have a clean cut example to say, if you're handle was `big-red-dog`, it will now be `big-red-cat`. This is important for the release notes, because usually handles form URL's, so if there is someone who is using a URL, such as `example.org/big-red-dog/`, it would be great to warn them and say, that link is actually now going to be `example.org/big-red-cat/`. Particularly important if the link is being referenced elsewhere (other websites, or worse, in print).

So, do the benefits of this change outweigh the risks?

jensscherbl:
> if you're handle was big-red-dog, it will now be big-red-cat

All non-word characters `^/w` except `-` will be stripped out and truncation of very long handles might be slightly different. So I don't expect too many changed handles except a few edge cases, which might have caused problems anyway.

> So, do the benefits of this change outweigh the risks?

Definitely. I actually ran into issues a couple of times when clients copy pasted text with certain characters that weren't replaced by transliterations and remained in the handle.

symphony/lib/toolkit/class.general.php
```diff
@@ -312,46 +312,53 @@ public static function createHandle($string, $max_length = 255, $delim = '-', $u
...
+
+        // strip out tags
+
         $string = strip_tags($string);
 
-        // Remove punctuation
-        $string = preg_replace('/[\\.\'"]+/', null, $string);
+        // remove unwanted characters
+
+        $string = preg_replace('/[^\w- ]+/', null, $string);
```
brendo: Instead of `null`, we should use `' '`.

jensscherbl: > This new createHandle returns dog-cat32lf-dfajl without this change.

You mean invalid characters should better be replaced by `-` instead of just being stripped out?

brendo: In a round about way, yeah. Code later on already handles this by replacing all spaces with the `$delim`.

jensscherbl:
Rebased and edited [as suggested][1].

[1]: https://github.com/symphonycms/symphony-2/pull/2106#discussion_r26902137

animaux:
Oh, this would make my [last issue](https://github.com/symphonycms/symphony-2/issues/2396) pointless of course.

PR #2205, opened 2014-09-19
Remove Pragma HTTP response header

DavidOliver:
The Pragma header is a request header, not a response header.

> When the no-cache directive is present in a request message, an application SHOULD forward the request toward the origin server even if it has a cached copy of what is being requested. [...] Note: because the meaning of "Pragma: no-cache as a response header field is not actually specified, it does not provide a reliable replacement for "Cache-Control: no-cache" in a response”

http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

> Many people believe that assigning a Pragma: no-cache HTTP header to a representation will make it uncacheable. This is not necessarily true; the HTTP specification does not set any guidelines for Pragma response headers; instead, Pragma request headers (the headers that a browser sends to a server) are discussed. Although a few caches may honor this header, the majority won’t, and it won’t have any effect. Use the headers below instead.”

https://www.mnot.net/cache_docs/#PRAGMA

The Cache-Control header already does what we want.

Any weird old Internet Explorer issues related to the Pragma header not being added to HTTP responses should have ceased to be of concern to Symphony long ago.

### DIFF
--- a/install/lib/class.installerpage.php
+++ b/install/lib/class.installerpage.php
@@ -30,7 +30,6 @@
 			$this->addHeaderToPage('Cache-Control', 'no-cache, must-revalidate, max-age=0');
 			$this->addHeaderToPage('Expires', 'Mon, 12 Dec 1982 06:14:00 GMT');
 			$this->addHeaderToPage('Last-Modified', gmdate('D, d M Y H:i:s') . ' GMT');
-			$this->addHeaderToPage('Pragma', 'no-cache');
 
 			$this->setTitle($this->_page_title);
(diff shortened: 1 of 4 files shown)


brendo:
Implemented Option 1. If the request is made using HTTP 1.0 and the Cache Control header is set to `no-cache`, we will add Pragma and Expires headers. Otherwise they will be omitted. This results in a small saving with reduced overhead sent to the browser which is nice :)

DavidOliver:
Nice! Thanks @brendo.

PR #2258, opened 2014-11-18
Change autocomplete value to new-password

nitriques:
This was a breaking change in Chrome M34 that has been fixed in M37: The password manager would completely ignore the autocomplete="off" directive and would still fill up the first input[type=password] it could find.

This was making me nervous since I spotted my password getting saved in plain text in the config file.

The only was to fix (without any hacks) is to use the new value "new-password" (See https://code.google.com/p/chromium/issues/detail?id=370363#c8 and http://lists.whatwg.org/htdig.cgi/whatwg-whatwg.org/2014-March/253895.html)

I've only changed the password field and left the username field untouched.

I've tested it in Chrome and Firefox (both do well).

This would need testing in Safari.

### DIFF
--- a/symphony/lib/toolkit/email-gateways/email.smtp.php
+++ b/symphony/lib/toolkit/email-gateways/email.smtp.php
@@ -396,7 +396,7 @@ class SMTPGateway extends EmailGateway
 
         $label = Widget::Label(__('Password'));
         $label->setAttribute('class', 'column');
-        $label->appendChild(Widget::Input('settings[email_smtp][password]', $this->_pass, 'password', array('autocomplete' => 'off')));
+        $label->appendChild(Widget::Input('settings[email_smtp][password]', $this->_pass, 'password', array('autocomplete' => 'new-password')));
         $div->appendChild($label);
         $group->appendChild($div);
 


jensscherbl:
@nitriques I still think we should stick to the [HTML5 spec][1] and don't merge this. What do you think?

[1]: http://www.w3.org/TR/2014/REC-html5-20141028/forms.html#autofilling-form-controls:-the-autocomplete-attribute

brendo:
Agree, Chrome 37 is a thing of the past now as well.

nitriques:
I agree, but 

> This was making me nervous since I spotted my password getting saved in plain text in the config file.

I still get this. The SMTP fields are auto-completed even when not visible, which saves my symphony password into it.

I agree that including non-standard stuff sucks. But I still feels that this is a problem.

jensscherbl:
> I still get this. The SMTP fields are auto-completed even when not visible, which saves my symphony password into it.

Maybe you should file a bug report on the chromium issue tracker then...

nitriques:
>  If the field has autocompletion turned off and the browser simply ignores it, it's clearly a browser problem.

No. It's "by design"

https://code.google.com/p/chromium/issues/detail?id=352347 
https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/zhhj7hCip5c/discussion

And it turns out I could run chrome with a special flag... Which may not what I want.... I like to have auto-completion... But not everywhere.

jensscherbl:
> No. It's "by design"

A browser problem "by design" is still a browser problem.

nitriques:
> A browser problem "by design" is still a browser problem.

I agree. But this cancels all my arguments.

> Maybe you should use a browser where it works as expected then

Yeah, I'll use Spartan :smile:

PR #2267, opened 2014-11-24
Fix some scrutinizer issues

nitriques:
content.publish.php was the file with the most issue, so I've started with that.

It's mostly missing variable declaration.

I've also touched the FieldManager, as many issues in content.publish.php was that `FieldManager::fetch` 

I know this process is error prone and I've reviewed/tested it carefully. Please, double check and read my comments in the code!


nitriques:
@brendo Why did you closed this ? You fixed it elsewhere ?

PR #2298, opened 2014-12-22
Template path was hardcoded

jensscherbl:
Symphony allows for changing the template path via the `PAGES` constant, but the path was still hardcoded for XSL imports.

I was wondering why this is implemented as an empty template importing the actual template instead of using the actual template directly, so here's a little background...

First XSL import is always relative to the executing script. In our case this is `index.php`, so our imports would be relative to the `DOCROOT`. Importing the page template itself changes the context for further imports, so we can import relative to the page template instead of the `DOCROOT` in our templates.


nitriques:
>  Maybe we can change this to a more generic rule like RewriteRule ^(.*)\.xsl$ - [F]?

XSL files can be used for client-side... So I would say no since it could break things.

PR #2301, opened 2014-12-28
Do not allow mixture of forward and backslashes in install path match

jurajkapsz:
If this idea is correct. Just an idea review.

Also, I was not sure about the `\\\\` pattern part, in which environments does it match, and if its ok also like `\\install\\index.php` . If not, a `\\{1,2}` pattern would be fine I think. Well, initially I was thinking about to use only a `\\`, but this syntax was throwing errors in my local tests (I don't understand why), while on online testers it was fine.

### DIFF
--- a/symphony/lib/boot/bundle.php
+++ b/symphony/lib/boot/bundle.php
@@ -13,7 +13,7 @@
     // Redirect to installer if it exists
     if (!file_exists(CONFIG))
     {
-        $bInsideInstaller = (bool)preg_match('%(/|\\\\)install(/|\\\\)index.php$%', $_SERVER['SCRIPT_FILENAME']);
+        $bInsideInstaller = (bool)preg_match('/(\/|\\\\)install\1index\.php$/', $_SERVER['SCRIPT_FILENAME']);
 
         if (!$bInsideInstaller && Symphony::isInstallerAvailable()) {
             header(sprintf('Location: %s/install/', URL));


nitriques:
What are you trying to fix here ? Is it security related ?

jurajkapsz:
It's not a fix, more a tighter syntax of the regexp at most.

So we have either a `/path/to/php` or `\\path\\to\\php` , and not a `/path\\like/this\\php` match or `\\path/like/this\\php` etc.

I don't know if this is an useful commit, just wanted to share this little idea. It can be discarded and if it was also wrong on top, I can spank myself.

nitriques:
Ah! No I think that allowing only one version of the directory separator (no able to mix them) is ok.

And I was not aware of the \1 trick. Good thing!

@brendo do you agree ?

nitriques:
One thing though: Symphony mostly uses / as the regexp delimiter (instead of %). Could you change this ?

nitriques:
Also the dot (.) before php should be escaped (\. instead of only . which matches any characters)

jurajkapsz:
> And I was not aware of the \1 trick. Good thing!

Yes, I've also learned this recently, so I wanted to share. The `\N` e.g. `\1` syntax is a backreference to a previous capturing `(match)`, in `N` order of the `()` appearence in the expression, holding the actual matched value.

> Also the dot (.) before php should be escaped (. instead of only . which matches any characters)

A good point. Will do.

> One thing though: Symphony mostly uses / as the regexp delimiter (instead of %). Could you change this ?

Ok, can do. Maybe all these changes will have meaning when @brendo will want to incorporate them.

And then I should probably create a single commit for all this (a rebase I guess?).

I was also puzzled about the reason for `\\\\`, does it match a `\\` starting path? Or just a backslash path `\like\this\php`.

nitriques:
> (a rebase I guess?).

```
git commit -a --amend
git push origin branch --force
````

It's easier to amend your last commit in your case (rebase + squash would also work but it's more error prone)

> I was also puzzled about the reason for \\\\,

This is common in lots of programming language: you escape the string literal AND the regexp char.

'\\'  == one \ in the string. \ is an escape char (for \n\t\r)

and then \ is also special in regexp so it needs another one (actually 2) the match one single \ char.

C#, Java, Javascript kind of have the same behavior :)


As for regexp^, I would recommand this resource https://regex101.com/#pcre

That's how I've test your regexp :)

brendo:
Is this ok to merge (it looks like it?). Has this been tested on a Windows box?

brendo:
> Ok, can do. Maybe all these changes will have meaning when @brendo will want to incorporate them.

In this scenario where the expression has a million slashes already, I'm fine with using a different delimiter for readability.

nitriques:
Ok. But not it is not ready, still missing the fix for the litteral dot in index.php

jurajkapsz:
> Ok. But not it is not ready, still missing the fix for the litteral dot in index.php

Will ship everything mentioned by tomorrow.

> Has this been tested on a Windows box?

I haven't. But if Windows does not have a `\\path/likethis` than it should be ok.

jurajkapsz:
Ok, I've made the forced update with the updated syntax, also made the commit comment easier, but did not manage to test it under Windows in this run, only undex Linux. Will return to the win test in the evening, which is in couple of hours here.

Thanks @nitriques and @brendo for your suggestions and tips!

andrewminton:
This breaks local WAMP and XAMPP installs. Just wont run the XSLT processes
due to invalid URI reference..

jensscherbl:
Wouldn't it make more sense to just get rid of that entire regex/path checking and simply set a flag inside `/install/index.php` before `bundle.php` is included? oO

nitriques:
> Wouldn't it make more sense to just get rid of that entire regex/path checking and simply set a flag inside /install/index.php before bundle.php is included? oO

Maybe, but I think it's easier to break than check the file path...

jensscherbl:
> Maybe, but I think it's easier to break than check the file path...

How?

nitriques:
> How?

I had some setup where variables declared globally where not readable by other functions...

jensscherbl:
> I had some setup where variables declared globally where not readable by other functions...

How does this affect the situation we have here? This is not a function.

The change I suggest is simply set `$bInsideInstaller` in `/install/index.php` before `bundle.php` is included, and check if `$bInsideInstaller` is set inside `bundle.php`.

What's happening right now is that `$bInsideInstaller` is set inside `bundle.php` based on the file path and a cumbersome regex.

nitriques:
`$bInsideInstaller` may not be readable from bundle.php

But this happend to me only once...

jensscherbl:
> $bInsideInstaller may not be readable from bundle.php

This seems pretty weird.

nitriques:
Yeah it was. @brendo ever encounter similar situation ?

brendo:
Based off @andrewminton comments on [this commit](https://github.com/jensscherbl/symphony-2/commit/dc211a791af9082f753fbd95edf9bcd1a23c0808#commitcomment-9263909) this PR should be closed. It appears that on a Windows environment you can have a mixed path.

nitriques:
> It appears that on a Windows environment you can have a mixed path

Yeah but it's not "right". And on linux, it breaks.

I would like to test the change proposed by @jensscherbl

jensscherbl:
> Yeah but it's not "right". And on linux, it breaks.

On linux, it doesn't have mixed paths... Mixed paths exist because we define (Unix) paths for certain things (workspace, templates, etc.) and get other paths, like the document root, from the system (where Windows returns backslash paths). So if you concatenate `DOCROOT` and `/workspace/pages`, you get a mixed paths on Windows. Good thing is, Windows can handle this fine, so no problem here.

nitriques:
2.6.x ? why not 2.6.0 ?

jensscherbl:
> 2.6.x ?

Sorry, my fault. 2.7.0 would be more appropriate, since 2.6.x should be hotfixes only.

> why not 2.6.0

Because every change made should be well tested and could potentially create new issues that need to be fixed before a release, or even worse, remain unnoticed.

If we want to get 2.6.0 out soon, we shouldn't introduce new changes that aren't really fixing anything and are only nice to have at this point.

Same for other, pre-2.6.0 issues popping up recently. You have to decide when it's time to stop adding stuff and start stabilizing a release.

nitriques:
Ok though that your proposed change was a bug fix ;)

But I understand your reasoning and I agree with it.

PR #2304, opened 2015-01-05
gzopen vs. gzopen64

Petertron:
When PHP is compiled with large file support, gzopen is replaced by gzopen64. This affects the archiving of log files.


brendo:
Thank you!

nitriques:
Thanks indeed!!

Issue #1821, opened 2013-09-17
Email attachment encodings are not set properly


nitriques:
When you add an attachment to an email, some information are missing, such as the proper encoding to use.

Right now I am getting this:

```
--=_mix_a24bb7793fe5cd57b623d7f06fafa6ad
Content-Type: text/csv; name="filename.txt"
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="filename.txt"
```

My file contains utf-8 characters, but the utf-8 information is missing and email client are decoding the attachment as plain ASCII.

If I add charset=utf-8, everything gets decoded properly, i.e.

```
--=_mix_a24bb7793fe5cd57b623d7f06fafa6ad
Content-Type: text/csv; charset=UTF-8; name="filename.txt"
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="filename.txt"
```

Right now I added it on this line: https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/toolkit/class.emailgateway.php#L513

But it now makes ALL attachment to be utf-8. Maybe we should use the `mb_detect_encoding` function in order to try and guess what's the orginial encoding of the attached file.

PR #2305, opened 2015-01-05
Allow email attachment mime type definition

jonmifsud:
Update email gateway to allow the attachment mime type to be pre defined.

The php `finfo_file` is not always accurate, and is at times unable to determine csv types, and similar text types, giving them plain text types. For remote files such as those generated using [Content Type Mappings](http://symphonyextensions.com/extensions/content_type_mappings/) there is also no extension. The only way to properly attach these files is by allowing the developer to set a mime type attribute.

PS. before merging kindly confirm that it does not have any adverse effect on anything else, I don't think so as the other pre-defined mime types should trigger the description.


nitriques:
I do not see why it could break. Also very aligned with discussion in https://github.com/symphonycms/symphony-2/issues/1821

michael-e:
Nor do I. It looks pretty straightforward, so I am not afraid of merging it.

PR #2310, opened 2015-01-08
Use execute instead of grab

nitriques:
grab has been deprecated for sometime now. Time to move on.

@brendo is it ok to directly commit those kind of things in integration ?

### DIFF
--- a/symphony/lib/toolkit/class.frontendpage.php
+++ b/symphony/lib/toolkit/class.frontendpage.php
@@ -894,7 +894,7 @@ class FrontendPage extends XSLTPage
 
             // if the XML is still null, an extension has not run the data source, so run normally
             if (is_null($xml)) {
-                $xml = $ds->grab($this->_env['pool']);
+                $xml = $ds->execute($this->_env['pool']);
             }
 
             if ($xml) {


nitriques:
@brendo bump! :smile: I really think it's 2.6 scoped... cc @nilshoerrmann @jensscherbl

brendo:
I'd like to keep this for 2.6.0. Removing would be a breaking change to any old installations, and we really shouldn't do that without a bump to 3.0.0

nitriques:
Ok then!

jensscherbl:
> Removing would be a breaking change to any old installations, and we really shouldn't do that without a bump to 3.0.0

Correct, removing has to wait for 3.0.0. But this commit should be fine for 2.6.0.

jensscherbl:
> But this commit should be fine for 2.6.0.

Oh wait, this would affect existing datasources. My mistake.

brendo:
This commit is only part of the puzzle:

- The base `DataSource` class still has a `grab()` function reference
-

nitriques:
> The base DataSource class still has a grab() function reference

Was it fixed ? If not, leave the issue open I'll fix it.

Issue #2312, opened 2015-01-12
Date Field calendar: week starts with sunday


animaux:
In germany it’s uncommon to have the week starting on sunday. Is there a way to select monday as well? Can this be reflected by the language settings somehow?

nilshoerrmann:
Sunday is the default start of the week in JS (don't ask!). CLNDR offers an offset setting, so yes this can be customised.

animaux:
Should we have this as an option of the date field then, or rather have a language extension trigger the right behaviour (if possible)?

nilshoerrmann:
If we allow an offset, it should be a system preference in my eyes because the calendar is used for date fields and publish filtering – relying on localisations is restrictive because you cannot just set your week start to Wednesday and might also be confusing conceptually. I didn't want to introduce such feature until we know the calendar itself is stable.

animaux:
I agree. Still, a week starting on sunday is quite a usability problem in german.

nilshoerrmann:
Calling @nitriques and @brendo for some thoughts on this.

nitriques:
We have the same problem here: Canada uses Sunday while US use Monday.

I have a couple of ideas on how to solve this:

1. A new setting in config.php
2. Using the php's locale to automatically get it
3. Use the author's locale to try and guess it

But I would vote for number one. @brendo ?

nilshoerrmann:
Only number 1 is predictable across users. So I'd vote for that.

animaux:
I’ll have number 1 too. Without onions and garlic.

nilshoerrmann:
From the interface perspective, we'd just have to pass CLNDR a value for `weekOffset` (`0` for sunday etc.).

nitriques:
CLNDR seems to represent 0 as Sunday (and so does php w formatter) so the setting could be a simple integer.

I would also add this setting in the installer. But would you add it to the preference page too ?

nilshoerrmann:
> But would you add it to the preference page too ?

No, if it's the only setting.
Yes, if the date and time formats can we updated as well.

Personally, I'd prefer the latter – but for some reason this has never been editable in Symphony 2 (it was editable in version 1).

nitriques:
> Personally, I'd prefer the latter 

Yeah I agree with you. And this change would be a nice kick in the ass to add them all on the preference page.

nilshoerrmann:
[...]

PR #2325, opened 2015-01-20
Symphony 2.6: Remove calendar widget

nilshoerrmann:
Due to timezone problems (see #2312), this pull request removes the new calendar widget again.
It should be fixed and reapplied in https://github.com/symphonycms/symphony-2/tree/calendar (Symphony 2.7, if possible).


brendo:
Am I right in saying that the only reason we are removing this is because of the Feature Request to adjust the start of the week? This FR led to incorrect calendar rendering when combined with a system language that is not English.

Is it possible to just revert the FR and leave the calendar as is? Do the timezone issues affect the calendar otherwise?

nilshoerrmann:
No, the feature request has been added and is working as intended. The problem is that the calendar shifts days one weekday back depending on your timezone.

This issue was found while adding the week offset but it exists indepently.

brendo:
Can you provide a exact set of steps to replicate?

nilshoerrmann:
There aren't any specific steps, it's more the general setup:

- German OS (in my case Mac OS X Yosemite)
- German locales in the browser (in my case Firefox 35 with `Deutsch [de]` as primary language)
- German date and time (in my case CET which is `UTC+1`)

In this given setup, using the calendar will result in the first January starting on the wrong day (Wednesday instead of Thursday).

brendo:
Ok, well I think this may be fixed then:

- German Timezone in OS (Europe/Berlin)
- German Timezone in Symphony (Europe/Berlin)
- German Language set in Firefox/Chrome (`de`)

Creating a new entry, and flipping the calendar back to Januar 2015 results in the 1st being on Donnerstag, which is correct. Care to give `integration` a try and see if you can replicate?

nilshoerrmann:
I'll have a look later today (having a sick child around might delay that a bit).

brendo:
Ah that's no good! Hope you're still getting some sleep :)

nilshoerrmann:
I just checked and I can confirm that the calendar displays date/weekday connections correctly. 

> Ah that's no good! Hope you're still getting some sleep :)

Max is sleeping most of the time, so we get sleep as well. Nothing serious.

nilshoerrmann:
Revisiting this issue, I ask myself [how your changes to the to the calendar script might have fixed the issue](https://github.com/symphonycms/symphony-2/commit/57fda1f7e7b3ccd8a472f56747d60b82b03e29ec#diff-5c17bd4c69c458c61cfae7b3cc4c4984R152): it should work even without these changes.

I've got an idea (and I'd like @animaux to check this as well): Is it possible that the turn of the year fooled us and that we looked as January 20**14** which of course started on a Wednesday.

nilshoerrmann:
This pull request has been rendered obsolete by https://github.com/symphonycms/symphony-2/commit/57fda1f7e7b3ccd8a472f56747d60b82b03e29ec#commitcomment-9540637.

animaux:
@nilshoerrmann 

> Is it possible that the turn of the year fooled us and that we looked as January 2014 which of course started on a Wednesday.

Definitely not. I double- and triplechecked.

PR #2326, opened 2015-01-20
Calendar widget

nilshoerrmann:
<del>This pull request tracks the progress of fixing the calendar widget initially introduced in Symphony 2.6.0-beta1 but removed due to timezone isses (see #2325).</del> <ins>[See notes below](https://github.com/symphonycms/symphony-2/pull/2326#issuecomment-72442554).</ins>

- [ ] Date field: Check if the date field has to store the date twice in the database (local date + GMT). If possible only store once.
- [x] Date field: Make the field default to the system timezone.
- [x] Date field: Make the field display the used timezone in the interface (maybe as a floating label).
- [x] <del>Date field: Allow the user to change the date display to his local time (optional).</del>
- [x] <del>Date field: Make the field UI respect the correct timezone.</del>
- [x] Calendar: Add widget.
- [x] <del>Calendar: Fix day display, see #2312.</del>


brendo:
Just a heads up with Timezones that I'm not sure we should start displaying the time relative to current user. For example, think of music festival across a country with different timezones. The 9pm headliner should always appear as 9pm. Just because I'm on the east coast doesn't mean I want to see it in the backend as starting at 11:30pm. I might think this is wrong and change it to 9pm (which has actually changed it to 6:30pm).

> Date field: Make the field default to the system timezone.
> Date field: Make the field UI respect the correct timezone.

We do respect the timezone that is set in the Symphony configuration. Every time a date is formatted, that timezone [is applied](https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/core/class.datetimeobj.php#L196-L204). The whole Symphony instance runs in that timezone [actually](https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/core/class.datetimeobj.php#L106-L111).

nilshoerrmann:
> Just a heads up with Timezones that I'm not sure we should start displaying the time relative to current user. For example, think of music festival across a country with different timezones. The 9pm headliner should always appear as 9pm. Just because I'm on the east coast doesn't mean I want to see it in the backend as starting at 11:30pm. I might think this is wrong and change it to 9pm (which has actually changed it to 6:30pm).

Ah, no, this isn't what I meant because this is – as you pointed out – quite confusing for the user.
My idea was to show a floating label that displays the timezone the field is using (the one from the system preferences):

sym_timezone-label

Of course, this is only needed, _if my local timezone is different from the one the system uses_.

Especially if you think of cross-timezone scenarios, it would be a nice **optional feature**, if the user could just click that label to convert it to his local time. This wouldn't change the way the date is stored, it would just be converted to the system timezone on save. 

> We do respect the timezone that is set in the Symphony configuration. Every time a date is formatted, that timezone is applied. The whole Symphony instance runs in that timezone actually.

I understand that this is true for the PHP/DB part – it's just that the UI doesn't take care of timezones at all yet. We just didn't notice until we added date visualisations with the calendar widget.

nilshoerrmann:
From the other thread:

> Symphony supports timezones and the date field saves two values for as long as I can remember. The value is just a string which includes the timezone set in the Symphony configuration. The date is the UTC timestamp. We use this for filtering. We could exclusively use it. There was a reason why we don't exclusively use it, but it alludes me at the moment.

It doesn't make sense to store the value with a hardcoded timezone in my eyes, because if the system timezone is changed all displayed dates need to change and this will not be reflected in the DB. 

My guess is that Symphony needed to store this twice because of the poor date and time support in earlier PHP versions.

nitriques:
> that I'm not sure we should start displaying the time relative to current user.

We should *never* do this in the frontend. Dates and times must be timezone-less, unless opt-in by the user.

> My guess is that Symphony needed to store this twice because of the poor date and time support in earlier PHP versions.

Yeah that's true.

nilshoerrmann:
> Dates and times must be timezone-less, unless opt-in by the user.

That's not possible because Symphony stores them with a timezone _and adjust them_ if the system timezone changes.

nitriques:
> That's not possible because Symphony stores them with a timezone and adjust them if the system timezone changes.

Yeah I meant the display part. It should never uses the javascript (client) timezone, only the global one in Symphony's config.

nilshoerrmann:
The underlying issue has been solved by https://github.com/symphonycms/symphony-2/commit/57fda1f7e7b3ccd8a472f56747d60b82b03e29ec#commitcomment-9540637. The only thing we should consider is to highlight the system timezone in a label, if the user is located in a different timezone. Symphony will always use the system locale and not the user one.

nilshoerrmann:
@brendo: Is it worth removing the double date storage in `2.x` or is this something you'd like to handle for `3.0`?

nilshoerrmann:
The date field now displays a lable, if the user's timezone differs from the system timezone:

### Input

bildschirmfoto 2015-02-02 um 13 40 09

### Calendar widget

bildschirmfoto 2015-02-02 um 13 41 41

nilshoerrmann:
Brendan, I'll close this pull request now that it is obsolete. 
Feel free to decide how and when to handle the two date fields in the database.

PR #2329, opened 2015-01-21
ASSETS_URL constant

designermonkey:
I've done a thorough search through the code, but may have missed something.

As and when I find parts of the core that fail in my custom structure, I will send more PRs.

### DIFF
--- a/install/lib/class.installer.php
+++ b/install/lib/class.installer.php
@@ -486,28 +486,28 @@
 
             // Create manifest folder structure
             Symphony::Log()->pushToLog('WRITING: Creating ‘manifest’ folder (/manifest)', E_NOTICE, true, true);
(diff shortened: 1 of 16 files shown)


designermonkey:
PS, this is untested as I've just copied across what I've done in my custom structure, which works fine with these changes.

nitriques:
All my ad-hoc tests seems to run fine with your modifications. I think this is good to merged.

Thanks John!

brendo:
I won't pull just yet as I want to test 0e47535 in particular. I have a bad feeling that those constants are not available during install (or didn't use to be, autoloading might of changed this)

nitriques:
Ok Brendan, feel free to ask me anything if you need help.

designermonkey:
When I say not tested, I meant this particular commit, the installer works fine, as I've installed using it.

brendo:
Thanks John :)

PR #2333, opened 2015-01-23
Add count of entries (unfiltered). Fixes #2129.

### LINKED ISSUE #2129: Publish filtering option misses the count of entries
The new publish filtering is cool, but it doesn't display the count of entries. It would be useful to know

* how many entries are there if no filters are active
* how many entries have been found with the current filter(s)

michael-e:


symphony/content/content.publish.php
```diff
@@ -495,7 +495,10 @@ public function __viewIndex()
 
         // Flag filtering
         if (isset($_REQUEST['filter'])) {
-            $filter_stats = new XMLElement('p', '<span>– ' . __('filtered (%d of %d)', array($entries['total-entries'], EntryManager::fetchCount($section_id))) . '</span>', array('class' => 'inactive'));
+            $filter_stats = new XMLElement('p', '<span>– ' . __('%d of %d entries (filtered)', array($entries['total-entries'], EntryManager::fetchCount($section_id))) . '</span>', array('class' => 'inactive'));
+            $this->Breadcrumbs->appendChild($filter_stats);
```
nilshoerrmann: Can't we just have this line once after the condition? We'll always be adding the count, aren't we?

michael-e: You mean line 499? I just thought that öater one might add more conditions, which theoroetically might not append `$filter_stats` (but something different, or nothing). Maybe I over-thought this?.

nilshoerrmann: Haha, you just voted for always having a count and now you are already planning the pull-out :grinning:
I don't mind, I just noticed …

michael-e: You are right, again I optimized for a "future scenario" that doesn't exist. I will add the simplification to the pull request.

michael-e: https://github.com/michael-e/symphony-2/commit/0a084d774de18c06a3b3dee2314a1f5fc0d6d032

nilshoerrmann:
Danke, Michael!

nitriques:
Danke, Michael!

michael-e:
No prob!

PR #2334, opened 2015-01-23
Fixing array syntax for PHP 5.3

nitriques:
This bug was introduce in 4d18be1ab1b3af0e91ce694210bae81c8ec03478 and uses a PHP 5.4+ compatible only syntax.

This reverts to the old array() syntax.

//cc @michael-e


brendo:
Bah sorry guys, working heavily on PHP5.6 at the moment so just used to it now! Thanks for fixing!

nitriques:
> Bah sorry guys,

No problemo :P

PR #2335, opened 2015-01-26
A small fix for the tag list prepareImportValue

psychoticmeow:
The tag list field was preparing a value for import as an array, then passing it to preg_split which expects a string.

Also, I get the feeling nobody really understands this import/export stuff and that I'm the only one using it. Anyone want to do a Google hangout to talk about it?

### DIFF
--- a/symphony/lib/toolkit/fields/field.taglist.php
+++ b/symphony/lib/toolkit/fields/field.taglist.php
@@ -404,12 +404,12 @@ class FieldTagList extends Field implements ExportableField, ImportableField
         $message = $status = null;
         $modes = (object)$this->getImportModes();
 
-        if (!is_array($data)) {
-            $data = array($data);
+        if (is_array($data)) {
+            $data = implode(', ', $data);
         }
 
         if ($mode === $modes->getValue) {
-            return implode(', ', $data);
+            return $data;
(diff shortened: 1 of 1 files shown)


michael-e:
> I get the feeling nobody really understands this import/export stuff

That is true, at least for me. But instead of a hangout, I would prefer a comprehensive list of all possible field values and their intended usage. Or have I missed that?

psychoticmeow:
I'm not sure even I know that any more, the original idea was that you'd be able to have fields understand more than just post data. So you would export a DateTime from a date field, and then import it again, however the import modes of fields don't really match the output modes so it's kind of confusing.

Which is why I'd rather discuss it, and how I'm using it to come up with a plan to improve it and make it understandable.

michael-e:
So I will rather wait for the outcome. I am not a big help in programmers' discussions (since I am so bad in programming).

jonmifsud:
@psychoticmeow most of the extensions/fields have incorrect import/export stuff set had to change quite a few. Open to discuss but I guess we'll need @brendo to chip in maybe.

nitriques:
> Also, I get the feeling nobody really understands this import/export stuff and that I'm the only one using it. 

I am not for the mather. It always seemed to complicated.

Issue #2362, opened 2015-03-02
Class 'Symphony' not found


jdsimcoe:
I keep getting this error:

```
Fatal error:  Class 'Symphony' not found in /Users/jdsimcoe/Sites/atheycreek/symphony/lib/boot/bundle.php on line 18, referer: http://atheycreek.dev/
```

Not sure what is driving it.

brendo:
Is this with `2.6.0.rc-1` or with `2.5.3`? If it's the former, can you check if you have an [`autoload.php`](https://github.com/symphonycms/symphony-2/blob/integration/symphony/lib/boot/autoload.php) file and your [index.php matches](https://github.com/symphonycms/symphony-2/blob/integration/index.php)?

jdsimcoe:
It's with `2.6.0.rc-1`. I do have an `autoload.php` file and the `index.php` file matches.

brendo:
And running PHP 5.3+ I assume? When does this error occur? Just accessing your homepage?

jdsimcoe:
Yeah PHP 5.5+. But I’m not seeing the error anymore.


--

simko.io

jdsimcoe:
What is a `/vendor/autoload.php` file? Why would I need that? Is this a new thing to Symphony `2.6.0`?

brendo:
> What is a /vendor/autoload.php file? Why would I need that? Is this a new thing to Symphony 2.6.0?

Symphony 2.6 allows for installation via Composer, so there are two different possible autoload routes.

> Yeah PHP 5.5+. But I’m not seeing the error anymore.

That's strange/scary?

jdsimcoe:
@brendo Well I have since updated `JIT` and `s3upload_field` so either of those could have fixed it.

> Symphony 2.6 allows for installation via Composer, so there are two different possible autoload routes.

Does this mean it would be possible to deploy Symphony via Heroku?

brendo:
> Does this mean it would be possible to deploy Symphony via Heroku?

No. Unfortunately as I just discovered. The other hurdle here is Postgres and the parsing of database information. That should be a little easier with PDO support.

jdsimcoe:
> The other hurdle here is Postgres and the parsing of database information.

Why is Postgres a hurdle? I am running several instances of Craft CMS on Heroku, using ClearDB:

https://medium.com/@aj1215/craft-cms-on-heroku-79b991665b0b

They parse the Clear DB config settings in Heroku like so:

```php
[...]

Issue #2330, opened 2015-01-21
Autoloading & Composer


jensscherbl:
@brendo 

Never thought of this before, but...

Does it really make sense to have a custom autoloader and an optional Composer autoloader? I mean, if we treat Symphony as standalone software and not as a PHP package that requires Composer to be installed, shouldn't we include the vendor folder in our repository anyway and only use Composer internally for development? Following this approach, we could just always include the autoloader created by Composer.

brendo:
Implemented.

PR #2363, opened 2015-03-03
simplify autoloader, embrace composer

jensscherbl:
Basically what I suggested in #2330.

Could reduce confusion and redundancy when it comes to autoloading (#2362).


nitriques:
@jensscherbl We need to still support people who don't use it, am I right ?

jensscherbl:
> @jensscherbl We need to still support people who don't use it, am I right ?

This is not about people using composer, this is about us using composer. I included the vendor folder in Symphony's versioning, so the composer autoloader is now just part of Symphony's code.

nitriques:
Ok great. Let see what @brendo thinks. I am not against it.

brendo:
If someone is already using Composer in their installation, or continues to use it after they do a base install, would committing the autoloader cause issues when future Symphony versions are released? My thoughts is the autoload files will have progressed with the additional classes and then our cut down autoload will result in a merge conflict?

jensscherbl:
> If someone is already using Composer in their installation, or continues to use it after they do a base install, would committing the autoloader cause issues when future Symphony versions are released?

You mean if someone adds their own dependencies to Symphony's composer.json and vendor folder? 

Wouldn't be any different from changing any other of Symphony's core files. If you modify something and there are changes to these files in a new Symphony version, you need to readd your modifications after updating.

To be more specific, if you add dependencies via Composer, you need to delete your vendor folder, readd your custom dependencies and rerun `composer install` after updating.

The workflow doesn't change because of this pull-request.

jensscherbl:
The only good argument I can think of against including the vendor folder in our repository is that it's generally not recommended because of redundancy. But on the other hand, Symphony is a standalone product, not a library, package or framework, so running composer locally to install dependencies shouldn't be a requirement to use the product.

nitriques:
I agree with you Jens.

PR #2376, opened 2015-03-17
Fixed #2375

### LINKED ISSUE #2375: XSL import fails when Windows path contains a space
There is an issue with this commit https://github.com/symphonycms/symphony-2/commit/dc211a791af9082f753fbd95edf9bcd1a23c0808

Windows fails to import and xsl when there is a space in the path(i.e. "C:\Program Files (x86)\EasyPHP-DevServer-14.1VC11\data\localweb\projects\")

I have to revert the line back to `<xsl:import href="./workspace/pages/' . basename($page['filelocation']).'"/>` in order to get 2.6.0 to work on windows

jensscherbl:
Not a windows problem. XSLT has issues with spaces in URLs.

### DIFF
--- a/symphony/lib/toolkit/class.frontendpage.php
+++ b/symphony/lib/toolkit/class.frontendpage.php
@@ -537,7 +537,7 @@ class FrontendPage extends XSLTPage
 
         $this->setXML($xml);
         $xsl = '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' .
-               '    <xsl:import href="' . str_replace('\\', '/', str_replace(' ', '%20', $page['filelocation'])) . '"/>' .
+               '    <xsl:import href="' . str_replace(array('\\', ' '), array('/', '%20'), $page['filelocation']) . '"/>' .
                '</xsl:stylesheet>';
 
         $this->setXSL($xsl, false);
--- a/symphony/lib/toolkit/class.frontendpage.php
+++ b/symphony/lib/toolkit/class.frontendpage.php
@@ -536,10 +536,9 @@ class FrontendPage extends XSLTPage
         $xml->prependChild($params);
 
         $this->setXML($xml);
-        $xsl = '<?xml version="1.0" encoding="UTF-8"?>
-        <xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
-            <xsl:import href="' . str_replace('\\', '/', $page['filelocation']) . '"/>
-        </xsl:stylesheet>';
+        $xsl = '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' .
+               '    <xsl:import href="' . str_replace('\\', '/', str_replace(' ', '%20', $page['filelocation'])) . '"/>' .
+               '</xsl:stylesheet>';
 
         $this->setXSL($xsl, false);
         $this->setRuntimeParam($this->_param);
@@ -1035,4 +1034,3 @@ class FrontendPage extends XSLTPage
         return XMLElement::stripInvalidXMLCharacters($parameter);
     }
 }
-
--- a/symphony/lib/toolkit/class.frontendpage.php
+++ b/symphony/lib/toolkit/class.frontendpage.php
@@ -536,8 +536,9 @@ class FrontendPage extends XSLTPage
         $xml->prependChild($params);
 
         $this->setXML($xml);
-        $xsl = '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' .
-               '    <xsl:import href="' . str_replace(array('\\', ' '), array('/', '%20'), $page['filelocation']) . '"/>' .
+        $xsl = '<?xml version="1.0" encoding="UTF-8"?>' .
+               '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' .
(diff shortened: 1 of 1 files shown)


nitriques:
As soon as you have confirmation that this is fixed, I'll merge it.

Should we release 2.6.1 right now ?

And should we merge this into integration right now ? Or wait until 2.6.1 is released ?

jensscherbl:
>  Or wait until 2.6.1 is released ?

Wait until it's released.

nitriques:
Ok thanks!

symphony/lib/toolkit/class.frontendpage.php
```diff
@@ -536,10 +536,9 @@ private function __buildPage()
         $xml->prependChild($params);
 
         $this->setXML($xml);
-        $xsl = '<?xml version="1.0" encoding="UTF-8"?>
-        <xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
-            <xsl:import href="' . str_replace('\\', '/', $page['filelocation']) . '"/>
-        </xsl:stylesheet>';
+        $xsl = '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' .
+               '    <xsl:import href="' . str_replace('\\', '/', str_replace(' ', '%20', $page['filelocation'])) . '"/>' .
```
brendo: This can be simplified further as `str_replace(array('\\', ' '), array('/', '%20'), $page['filelocation'])` will save invoking the function twice.

nitriques: +1

jensscherbl: Done, thanks.

nitriques: Great!

jensscherbl:
@brendo No Scrutinizer feedback for the `2.6.x` branch?

koolaid1551:
I still ran into issues, but I got it working by adding the file protocol 

See https://github.com/symphonycms/symphony-2/pull/2380

brendo:
Let's go back to the start for a second, and figure out exactly what we are trying to accomplish here because we've gone from something that has had no issues in 2.5.3, to something that's buggy in 2.6.0 and still not right on 2.6.x.

In 2.5.3 the code for this was:

```php
        $xsl = '<?xml version="1.0" encoding="UTF-8"?>
        <xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
            <xsl:import href="./workspace/pages/' . basename($page['filelocation']).'"/>
        </xsl:stylesheet>';
```

In 2.6.0 it currently stands as:

```php
        $xsl = '<?xml version="1.0" encoding="UTF-8"?>
        <xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
            <xsl:import href="' . str_replace('\\', '/', $page['filelocation']) . '"/>
        </xsl:stylesheet>';
```

On `2.6.x` branch it's now:

```php
        $xsl = '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' .
               '    <xsl:import href="' . str_replace(array('\\', ' '), array('/', '%20'), $page['filelocation']) . '"/>' .
               '</xsl:stylesheet>';
```

What problem are we trying to solve? The theoretical situation where page utilities don't live in `./workspace/pages`?

koolaid1551:
@brendo the issue is, in 2.6 everything was converted to absolute paths, if we want to keep this consistent than we will probably want to take this pull request https://github.com/symphonycms/symphony-2/pull/2380  otherwise if we are okay leaving this particular a relative path then we should revert it back to what it was in 2.5.3

nitriques:
>  in 2.6 everything was converted to absolute paths

When did that occurred ? What do you mean by everything ? Did I miss something ?

jensscherbl:
> What problem are we trying to solve? The theoretical situation where page utilities don't live in ./workspace/pages?

The issue was that folders for `workspace`and `pages` are configurable in `defines.php`, but hardcoded in `class.frontendpage.php`.

Some might wonder why the mentioned piece of code is necessary at all, so I checked that and it tuns out it's the only way to have xsl:imports relative to `pages` in your own templates.

I'm sorry this causes so much trouble in certain environments, will investigate today.

jensscherbl:
Hope this finally works now (tested on Windows 10 and Yosemite).

As @koolaid1551 pointed out, windows paths aren't starting with a `/` and were therefore treated as relative instead of absolute paths. Still not sure why it worked when I tested on Windows the first time when introducing the change.

Certain non-URI characters (like spaces and backslashes) caused issues as well and are now `rawurlencodeed`.

jensscherbl:
> indentation makes no sense

Other than improved (PHP) code readability, probably not.

> I'd prefer a single string including line breaks

What for? This string never sees the light of day, you won't ever see or touch it. It's just a workaround to allow relative (to the`pages` folder) import paths. PHP readability ranks higher in this case.

jensscherbl:
> that has had no issues in 2.5.3

Just for the record, it only had no issues as long as no one changed the `workspace` and `pages` paths.

I stumbled across the issue by accident because I had to change these in a project recently and Symphony stopped working.

I agree it's an edge case, but since paths are configurable in `defines.php` and that was the only place still using hardcoded paths, it had to be addressed.

nitriques:
Thanks Jens!

nitriques:
Could not find any problem running on Ubuntu.

koolaid1551:
This works on windows now as well. Thanks @jensscherbl !

nitriques:
Thanks @koolaid1551 for testing on, uuuuhhhggg, Windows ;)

PR #2377, opened 2015-03-18
Remove handles for parameters. Fixes #2332

### LINKED ISSUE #2332: Why are handles created for array values of page parameters?
If a page parameter is an array (and only then), Symphony creates additional handles for the values (instead of simply using `item` elements).

So, for example, Symphony creates:

```xml
<page-types>
	<item handle="index">index</item>
</page-types>
```

Wouldn't the following be enough?

```xml
<page-types>
	<item>index</item>
</page-types>
```

The handles are generated in https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/toolkit/class.frontendpage.php#L525.

If anybody thinks that these handles are useful, please tell us. Otherwise I suggest to remove them.

michael-e:


brendo:
Thank you!

Issue #2383, opened 2015-03-20
Add ctype support in General::intval


nitriques:
I've found those to be quite useful.

http://php.net/manual/en/function.ctype-digit.php

Maybe we should consider using it here: https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/toolkit/class.general.php#L1622-L1628 ?

brendo:
> Checks if all of the characters in the provided string

This is potentially the only quarm here, passing an integer to `General::intval` wouldn't work like it currently does :)

nitriques:
Well, what I meant was:

Check ctype_digit before doing the regexp ;)

jensscherbl:
> passing an integer to General::intval wouldn't work like it currently does

Current function works with strings and integers, `ctype_digit` would only work with strings.

>  This method makes sure that the value is a valid positive integer representation

Shouldn't the regex be `'/^[0-9]+$/i'` then?

nitriques:
> Still wouldn't work.

Why ?

> Shouldn't the regex be '/^[0-9]+$/i' then?

Yes it should. The change was made in fdb39fb

jensscherbl:
> Why ?

`is_numeric` checks if the value, string or integer, is numeric (only contains digits and `+` or `-`). The regex then checks if the value is positive (only contains digits).

Thinking about it, `is_numeric` seems a bit redundant here, regex alone should handle everything just fine.

`ctype_digit` can't be used because it only works with strings and fails if your original value already is an integer.

nitriques:
> Thinking about it, is_numeric seems a bit redundant here, regex alone should handle everything just fine.

You're right. The weird part: how does php handle doing regexp on numeric types ?

> ctype_digit can't be used because it only works with strings and fails if your original value already is an integer.

Like I said:  "Check ctype_digit before doing the regexp ;)"

I had perf in minde, but I now think it's an overhead...

jensscherbl:
> Like I said: "Check ctype_digit before doing the regexp ;)"

I'll try again... If the value is an integer, `ctype_digit` will return false, even if the value is numeric and positive. In such a case, the regex wouldn't even be executed.

If the value is a string on the other hand, `ctype_digit` would already check if the value is numeric and positive, so we wouldn't even need the regex in that situation.

Makes sense now?

nitriques:
[...]

PR #2384, opened 2015-03-21
Fixed issue where intval function returned invalid values

jensscherbl:
Not entirely sure about this, but as mentioned [here][1], the `intval` function didn't quite work as described. Please have a look at this.

[1]: https://github.com/symphonycms/symphony-2/issues/2383#issuecomment-84202035

### DIFF
--- a/symphony/lib/toolkit/class.general.php
+++ b/symphony/lib/toolkit/class.general.php
@@ -1621,9 +1621,11 @@ class General
      */
     public static function intval($value)
     {
-        if (is_numeric($value) && preg_match('/[0-9]+/i', $value) === 1) {
+        if (is_numeric($value) && preg_match('/^[0-9]+$/i', $value) === 1) {
+
             return intval($value);
         }
+
         return -1;
     }
 }


nitriques:
Damn, I fixed it in fdb39fb before seeing this PR. Is it ok like this ?

jensscherbl:
>  Is it ok like this ?

Sure, I'll close my pull request then.

Are you pushing commits directly to the repository? I think it might be better if we all (except @brendo maybe) use pull requests instead. Would make it easier for others to review and discuss our changes.

nitriques:
> Oh wait, looks like you pushed your change to integration? I think this should be a patch for 2.6.2.

Yeah maybe that's why PRs are a better thing. But since it was a such a small change, 

>  I think this should be a patch for 2.6.2.

On the other hand, should this be treated as a breaking change ? But it's also a fix, since we align the code with the spec....

jensscherbl:
> On the other hand, should this be treated as a breaking change?

No, it's a fix that makes sure the function does what it says it does.

nitriques:
> No, it's a fix that makes sure the function does what it says it does.

Yeah I agree with you. Changing the function's spec would be a breaking change.

So, should I remove my commit in integration and push it into 2.6.x ?

jensscherbl:
> So, should I remove my commit in integration and push it into 2.6.x ?

I guess you can leave it as is in `integration` and we should merge my pull request into `2.6.x`.

nitriques:
>  Should be fine when merging 2.6.x back to integration after 2.6.2 is released.

Yeah, you'll run into a conflict, but it's an easy one to fix.

jensscherbl:
@brendo Ok for me to merge this into `2.6.x` for an upcoming `2.6.2` release?

nitriques:
@jensscherbl I would say yes ;) but wait for @brendo approval :)

Issue #2386, opened 2015-03-21
is_session_empty() - $_SESSION isn't always an array.


pointybeard:
I think this is pretty obscure, but on a couple of custom 2.6.x installs, I get the following error:

`Symphony Warning: Invalid argument supplied for foreach()
An error occurred in /.../symphony/lib/boot/func.utilities.php around line 193`

Issue #2107 introduced some new code: `is_session_empty()` which loops through `$_SESSION` to check if it's populated, however there is no sanity check. On my particular install, `$_SESSION` is sometimes `NULL`.

I doubt this is a widespread problem, and likely relegated to some weird edge case, but I feel it's worth handling.

Simple fix is to wrap the `foreach()` in an if statement to check that `$_SESSION` is an array. I'll issue a pull request for it shortly.

pointybeard:
Eh. Looks like this was fixed in Integration branch. Sorry for duplicate. https://github.com/symphonycms/symphony-2/issues/2107

Issue #2107, opened 2014-06-22
Session cookie set on every frontend request again.


jensscherbl:
This was already solved in one of the 2.3 releases if I remember correctly, but 2.4 seems to fall back into old habits and creates a session cookie on every request to the frontend.

My guess is that this has something to do with the XSRF related changes, but I couldn't figure out the relevant parts myself.

PR #2387, opened 2015-03-21
Check if SESSION is actually an array before trying to loop over it

pointybeard:
See issue #2386 for details

### DIFF
--- a/symphony/lib/boot/func.utilities.php
+++ b/symphony/lib/boot/func.utilities.php
@@ -190,12 +190,15 @@ function cleanup_session_cookies()
 function is_session_empty()
 {
     $session_is_empty = true;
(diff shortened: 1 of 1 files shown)


pointybeard:
yeah, this was already fixed. Good idea to check integration first eh? ;)

https://github.com/symphonycms/symphony-2/issues/2107

michael-e:
> Good idea to check integration first eh?

AFAIK, the fix is in the 2.6.x branch. Am I right?

pointybeard:
Hmm, [nope](https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/boot/func.utilities.php#L193). It's only in integration, hence why I missed it.

pointybeard:
Ah yes, you are right. I've been tracking master.

Issue #2331, opened 2015-01-22
Display Publish Panel Pre-populate Bug


jonmifsud:
I'm not entirely sure if this is exclusively a "Core" issue. It's most certainly linked to the way Filters in the backend work. The screenshot below is the error I get when trying to create a new item from a Filtered view of an Association Field.

![screen shot 2015-01-22 at 09 10 05](https://cloud.githubusercontent.com/assets/859775/5852427/d39c23c8-a216-11e4-8d6e-7e62200f1c01.png)

As you might assume, the Association Field accepts/allows me to search using a text string which matches the "Associated Value" yet once using pre-populate it's expecting to have an entry id. Initially I thought that this was an issue with the Association Field which can be fixed by checking if the values are strings rather than integers and converting them. However (the documentation)[http://www.getsymphony.com/learn/api/2.5.2/toolkit/field/#displayPublishPanel] for `displayPublishPanel` reads the following:

> $data array(optional) any existing data that has been supplied for this field instance. this is encoded as an array of columns, each column maps to an array of row indexes to the contents of that column. this defaults to null.

So as a developer I should only cater for the fields which I support - thus a relation id should be an integer correct? As the dump within the screenshot shows this is passed "News". I'm not sure what would be the correct way to fix this, but at the very least we need to update the documentation to state that any supported filtering value can be passed.

brendo:
The Association Field deals with ID's mostly, so pre-population has to follow the same rules. The data that `displayPublishPanel` expects is `relation_id`'s. 

Are you able to generate a pre-populate link using an entry ID instead?

jonmifsud:
I can generate a pre-populate link using ID's *id* I search using ID's which the common user won't do. It also works fine If I use the association interface link (on the right of an article) to create a new entry.

The below is not really intuitive for the user, it's something specifically linked to filtering I believe. Is there a way we can get the field to give back the pre-populate value having used a filter? If not at the moment is it worth looking into or is there another way to go around it?

![screen shot 2015-01-22 at 10 33 17](https://cloud.githubusercontent.com/assets/859775/5853277/6a7c6504-a222-11e4-919a-439388b648f0.png)

brendo:
[...]

PR #2390, opened 2015-03-23
filter string to show string value not id

jonmifsud:
This is the inverse to the use of `fetchIDFromValue` when creating a filter string, it does not make sense to show the entry ID when the filtering UI is showing the value. So the ID should be converted back to a string value where applicable.

### DIFF
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -1790,6 +1790,15 @@ class contentPublish extends AdministrationPage
         if (isset($_REQUEST['prepopulate'])) {
             foreach ($_REQUEST['prepopulate'] as $field_id => $value) {
                 $handle = FieldManager::fetchHandleFromID($field_id);
+                
+                //This is in case it is an Association so the filter reads the text value instead of the ID
+                $field = FieldManager::fetch($field_id);
+                if ($field instanceof Field) {
+                    if (method_exists($field, 'fetchValueFromID')) {
+                        $value = $field->fetchValueFromID($value);
+                    }
+                }
+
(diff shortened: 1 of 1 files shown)


jensscherbl:
> It does not make sense to show the entry ID when the filtering UI is showing the value.

+1

nitriques:
I would rather use the standard `prepareReadableValue` instead of the non-standard `fetchIDFromValue`. Maybe use only if `prepareReadableValue` does not return any value... ?

jonmifsud:
Ah if `prepareReadableValue` serves the same function, seems like it does. We can use that, however I would think it should only be use if `fetchValueFromID` exists. Just to make sure we are not trying to convert unsupported fields.

Also `prepareReadableValue` has a data part, so I think it will require filling up before using the function.

nitriques:
> We can use that, however I would think it should only be use if fetchValueFromID exists. Just to make sure we are not trying to convert unsupported fields.

I think a better way to do this is by using `canPublishFilter()` method on the field.

> Also prepareReadableValue has a data part, so I think it will require filling up before using the function.

Yup. A new, optional parameter could be added. In the edit part, `$existingEntry` variable can be used.

If no entry is passed, then no conversions needs to be done.

---

Also, I think it might be worth adding a way to get which data fields are used in the filtering.

cc @brendo  ?

brendo:
I'm in favour of the pull request as is. While I'm very aware that associations, filtering, importing and exporting are a total mess on the fields (it feels like there are several functions that do *very* similar things), the addition of this new function will allow us to fix issues outside of publish filtering, like #2331

Issue #2311, opened 2015-01-09
DS named "index" breaks styling


nilshoerrmann:
I don't remember why, but as it seems, Symphony adds the handle of a Data Source as a class to the `body` of the DS Editor. If you have a DS named `index`, this results in the following broken styling:

![bildschirmfoto 2015-01-09 um 13 55 08](https://cloud.githubusercontent.com/assets/25466/5680029/a8d96b00-9807-11e4-9f88-759e8b618792.png)

Symphony uses `.index` to identify the table views in the backend so I'm not sure how to fix this without breaking half of the system.

michael-e:
LOL, a DS named "index"? Do you try to supersede me in doing crazy things?

nilshoerrmann:
Shouldn't have worked with you that long last year ;)

michael-e:
That was my first thought as well. :-))

nilshoerrmann:
:ambulance: (Going to the doctor know.) :fire_engine: :police_car:

michael-e:
My doctor said it's irreversible.

nilshoerrmann:
:disappointed:

nitriques:
You guys are funny :P

brendo:
Given it's 2.6.0, we could potentially add a `ds-` prefix to the classname. I imagine this flaw is also with Events as well.

brendo:
Not an easy one to fix. The context of the page (derived from the URL path) is automatically added to the body via the `AdministrationPage` class. At this point, it has no idea what each of the context values represent so it's pretty difficult to append `ds` or `event` to the start of them.

A possible solution would be to change the DS/Event indexes to generate URL's like `datasources/edit/ds-handle` and then have the `ds-` stripped. I don't like this too much though. Other solution is to go the other way and prefix all our classes with `sym-` and hope that people don't start creating DS's called Sym Index ;)

michael-e:
So what shall we do? I'd vote for prefixed classes (i.e. no "hacking"). A DS named "Sym Index" is really unlikely to occur, compared to "Index". @nilshoerrmann?

nilshoerrmann:
Conceptually, it would be best to use classes for styling purposes only and add context information in dedicated `data` attributes. But that's a breaking change.

nitriques:
> Conceptually, it would be best to use classes for styling purposes only and add context information in dedicated data attributes.

[...]

Issue #1585, opened 2012-12-25
Inconsistent values in admin getPageCallback()


designermonkey:
I've been looking at this for ages trying to figure out what I was doing wrong while writing an extension, and after a var_dump, I saw this inconsistency...

From `Administration::instance()->getPageCallback()`, on an `instanceof contentPublish` we are returned:

```
array
  'driver' => string 'publish' (length=7)
  'driver_location' => string '/Users/designermonkey/Sites/symphony.local/htdocs/symphony/content/content.publish.php' (length=86)
  'context' => 
    array
      'section_handle' => string 'parent-one' (length=10)
      'page' => string 'edit' (length=4)
      'entry_id' => int 2
      'flag' => null
  'pageroot' => string '/publish/parent-one/' (length=20)
  'classname' => string 'contentPublish' (length=14)
```

From an `instanceof contentBlueprintsSections we are returned:

```
array
  'driver' => string 'blueprintssections' (length=18)
  'driver_location' => string '/Users/designermonkey/Sites/symphony.local/htdocs/symphony/content/content.blueprintssections.php' (length=97)
  'context' => 
    array
      0 => string 'edit' (length=4)
      1 => string '1' (length=1)
  'classname' => string 'contentBlueprintsSections' (length=25)
  'pageroot' => string '/blueprints/sections/' (length=21)
```

Notice the keys inconsistency in the context array. This needs updating to be `page` and `section_id` respectively IMO. This is from the `integration` branch.

Issue #2367, opened 2015-03-06
Content pages can be rendered event if the extension is not enabled/installed


nitriques:
I really do not think this is a feature, hence it's a bug.

Is it a security issue/concern ?

brendo:
It's definitely not a feature, that's for sure.

It's the `Administration` class that should get a bit smarter and check if an Extension is installed during the route resolution (somewhere in [`getPageCallback()`](https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/core/class.administration.php#L309) I'd imagine). The `ExtensionManager` class is pretty lazy here, it doesn't really have a suitable method to call at this stage. While we can call `fetchStatus`, that unfortunately requires an associative array with handle and version keys. Alternatively we can call `listInstalledHandles` and check the array. Both will work, but aren't particularly nice to use. Probably worthwhile introducing a new accessor for `isInstalled($handle)`.

nitriques:
> Probably worthwhile introducing a new accessor for isInstalled($handle).

Will do. Thanks.

nitriques:
I can't decide which one we should use....

Version 1, strait DB call..

```php
    /**
     * Returns true if extension $handle is installed
     * @since Symphony 2.6.0
     * @return boolean
     */
    public static function isInstalled($handle)
    {
        if (Symphony::Database()->isConnected()) {
            $handle = MySQL::cleanValue($handle);
            $count = Symphony::Database()->fetchCol(
                'c',
                "SELECT count(`id`) as `c` FROM `tbl_extensions` WHERE `status` = 'enabled'
                 AND `name` = '$handle'"
            );
            return intval($count) == 1;
        }
        return false;
    }
```

Works ok, simple enough, makes sure we have the latest value...

---

Version 2

```php
    /**
     * Returns true if extension $handle is installed
     * @since Symphony 2.6.0
     *
     * @return boolean
     */
    public static function isInstalled($handle)
    {
        return in_array($handle, self::listInstalledHandles());
    }
```

Even simpler, but load extra data in some cases (returning ALL extensions name)

nitriques:
I woul also add

```php
$callback['extension'] = $extension_name;
```

nitriques:
Here's a temporary commit for review https://github.com/DeuxHuitHuit/symphony-2/commit/ae6f51b94ba5afda672e25529f13176fb064b3cf

brendo:
This will not make 2.6.0 final.

nitriques:
[...]

PR #2392, opened 2015-03-24
Move to data-attributes in the backend admin pages instead of using classes

brendo:
This pull request changes the way we interact with the Administration `$context`. Instead of the context (which is the URL path items) being added blindly as classes to the `<body>`, they are now added as `data-attributes` as suggested in #2311. If the value is an array, it will be JSON encoded (which fixes an existing bug with associations appearing as `associations-Array`).

To further address #1585, the `Administration->getPageCallback()` has been educated about:

- `/blueprints/pages/`
- `/blueprints/events/`
- `/blueprints/datasources/`
- `/blueprints/pages/`

These pages now reference the following path: `/blueprints/events/edit/add_date/saved/` with named parameters instead of just indexed:

- `action`: edit
- `handle`: add_date (for pages & sections, this will be named `id`)
- `flag`: saved

As part of this change, the CSS classes have been cleaned up and the references to `$this->context[0]` (etc) have been updated to reflect the new named parameters.

This change will have an impact on any extension that:

- Relies on `$this->_context` for any of the blueprint pages (PHP)
- Relies on any of those context items being available as a class (either in PHP or JS)

I think it's a worthwhile break, as it's now less magical for extensions!

CC in original bug reporters: @nilshoerrmann, @designermonkey


designermonkey:
:+1:

nitriques:
:+1:

nilshoerrmann:
:+1:

jensscherbl:
:+1:

brendo:
The logic for actually parsing the 'extras' of a URL shouldn't really sit in the `Administration` class. The `AdministrationPage` class is a much better place for this logic as pages can then parse the URL however they wish and create their owned named context.

So acting on this, I've made dc189a0. The logic for the Publish, Authors, Events, Datasources, Pages and Sections context has moved to the relevant classes into a `parseContext` method and out of the `Administration` class.

This is a funny example of how extensions have had more power than the core. Extensions already have access to the `AdminPagePostCallback` delegate, which allows the context to be overridden as they see fit. The addition of `parseContext` means that extensions which provide multiple pages can override the context in a more convenient fashion instead of having to do conditional logic on `$page` inside the delegate callback.

There is still some extra cleanup that can be done for the`/login` route as well as enforcing the fix for #2367 before I'd merge this change.

Anyone spot anything odd yet?

brendo:
This now includes @nitriques fix to check that Extensions are enabled before routing to them. I'm in a position to merge this into `integration` now as long as everyone is happy :)

nitriques:
Yeah!!!

PR #2394, opened 2015-03-26
Support for internationalized domain names

jensscherbl:
Support for internationalized domain names was [introduced in 2.4][1], but broke again with the `index.php` reshuffling in [2.5.0][2].

Here's a fix to bring it back.

[1]: http://www.getsymphony.com/download/releases/version/2.4/
[2]: http://www.getsymphony.com/download/releases/version/2.5.0/

PR #2395, opened 2015-03-26
Improved redirect function

jensscherbl:
- Use `SymphonyErrorPage ` exception for nicer error message if headers already sent.
- Simplified idn to ascii conversion.
- Introduced optional `$status` parameter, so developers can use other status codes like [`303 See Other`][1] or `307 Temporary Redirect`.
- Removed caching headers. For redirects, the [status code already indicates the appropriate caching behavior][2].

[1]: http://en.wikipedia.org/wiki/HTTP_303
[2]: https://tools.ietf.org/html/rfc2616#section-10.3

### DIFF
--- a/symphony/lib/boot/func.utilities.php
+++ b/symphony/lib/boot/func.utilities.php
@@ -8,7 +8,7 @@
  * Redirects the browser to a specified location. Safer than using a
  * direct header() call
  *
- *  @param string $url
+ *  @param string $location
  *  @param string $status (optional)
  */
 function redirect($location, $status = '302 Found')
--- a/symphony/lib/boot/func.utilities.php
+++ b/symphony/lib/boot/func.utilities.php
@@ -23,7 +23,8 @@ function redirect($location, $status = '302 Found')
 
     // convert idn back to ascii for redirect
     if (function_exists('idn_to_ascii')) {
-        $location = str_replace(HTTP_HOST, idn_to_ascii(HTTP_HOST), $location);
+        $host     = parse_url($location, PHP_URL_HOST);
+        $location = str_replace($host, idn_to_ascii($host), $location);
     }
 
     header('Status: '   . $status);
--- a/symphony/lib/boot/func.utilities.php
+++ b/symphony/lib/boot/func.utilities.php
@@ -13,22 +13,16 @@
  */
 function redirect($location, $status = '302 Found')
 {
+    // throw exception if headers already sent
     if (headers_sent($filename, $line)) {
-
-        // throw exception if headers already sent
-
         throw new SymphonyErrorPage(sprintf(
-
             'Cannot redirect to <a href="%s">%s</a>. Output has already started in %s on line %s.',
-
             $location, $location, $filename, $line
         ));
     }
 
+    // convert idn back to ascii for redirect
     if (function_exists('idn_to_ascii')) {
-
-        // convert idn back to ascii for redirect
-
         $location = str_replace(HTTP_HOST, idn_to_ascii(HTTP_HOST), $location);
     }
 
--- a/symphony/lib/boot/func.utilities.php
+++ b/symphony/lib/boot/func.utilities.php
@@ -9,31 +9,31 @@
  * direct header() call
  *
  *  @param string $url
+ *  @param string $status (optional)
  */
-function redirect ($url)
+function redirect($location, $status = '302 Found')
 {
-    // Just make sure.
-    $url = str_replace('Location:', null, $url);
-
     if (headers_sent($filename, $line)) {
-        echo "<h1>Error: Cannot redirect to <a href=\"$url\">$url</a></h1><p>Output has already started in $filename on line $line</p>";
-        exit;
-    }
 
-    // convert idn back to ascii for redirect
+        // throw exception if headers already sent
+
+        throw new SymphonyErrorPage(sprintf(
+
+            'Cannot redirect to <a href="%s">%s</a>. Output has already started in %s on line %s.',
+
+            $location, $location, $filename, $line
+        ));
+    }
 
     if (function_exists('idn_to_ascii')) {
-        $root = parse_url(URL);
-        $host = $root['host'];
-        $url  = str_replace($host, idn_to_ascii($host), $url);
+
+        // convert idn back to ascii for redirect
+
+        $location = str_replace(HTTP_HOST, idn_to_ascii(HTTP_HOST), $location);
     }
 
-    header('Status: 302 Found');
-    header('Expires: Mon, 12 Dec 1982 06:00:00 GMT');
-    header('Last-Modified: ' . gmdate('D, d M Y H:i:s') . ' GMT');
-    header('Cache-Control: no-cache, must-revalidate, max-age=0');
-    header('Pragma: no-cache');
-    header("Location: $url");
+    header('Status: '   . $status);
+    header('Location: ' . $location);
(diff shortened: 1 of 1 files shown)


symphony/lib/boot/func.utilities.php
```diff
@@ -8,32 +8,32 @@
...
+    }
 
     if (function_exists('idn_to_ascii')) {
-        $root = parse_url(URL);
-        $host = $root['host'];
-        $url  = str_replace($host, idn_to_ascii($host), $url);
+
+        // convert idn back to ascii for redirect
+
+        $location = str_replace(HTTP_HOST, idn_to_ascii(HTTP_HOST), $location);
```
brendo: This is an existing flaw with the `redirect` function, but we may as well fix it now (especially with #2394). This line assumes that the `$location` is always going to be hosted on the Symphony domain, when in fact the `redirect()` function accepts any `$location` to redirect to.

symphony/lib/boot/func.utilities.php
```diff
@@ -8,32 +8,32 @@
...
     }
 
-    header('Status: 302 Found');
-    header('Expires: Mon, 12 Dec 1982 06:00:00 GMT');
```
brendo: I'm fairly sure these lines are required to prevent browsers from caching the redirect.

symphony/lib/boot/func.utilities.php
```diff
@@ -8,32 +8,32 @@
  * Redirects the browser to a specified location. Safer than using a
  * direct header() call
  *
- *  @param string $url
+ *  @param string $location
+ *  @param string $status (optional)
```
brendo: This may be problematic as the input is not sanitised or checked to be a valid redirect HTTP code. I like the dumbness and simplicity of the redirect function so it's probably ok to say as is. Just something to highlight.

symphony/lib/boot/func.utilities.php
```diff
@@ -8,32 +8,32 @@
...
-    // Just make sure.
-    $url = str_replace('Location:', null, $url);
-
     if (headers_sent($filename, $line)) {
-        echo "<h1>Error: Cannot redirect to <a href=\"$url\">$url</a></h1><p>Output has already started in $filename on line $line</p>";
-        exit;
-    }
 
-    // convert idn back to ascii for redirect
+        // throw exception if headers already sent
```
brendo: Just curious on this code style, what's it derived from?

symphony/lib/boot/func.utilities.php
jensscherbl: Well, if the host in `$location` is different from `HTTP_HOST`, there's nothing to replace, so the line shouldn't have any effect on external redirects, I guess?

symphony/lib/boot/func.utilities.php
jensscherbl: According to [the spec][1], only if you want to change the default caching behavior for your status code.

> **302 Found**
> This response is only cacheable if indicated by a Cache-Control or Expires header field.

> **303 See Other**
> The 303 response MUST NOT be cached, but the response to the second (redirected) request might be cacheable.

> **307 Temporary Redirect**
> This response is only cacheable if indicated by a Cache-Control or Expires header field.

[1]: https://tools.ietf.org/html/rfc2616#section-10.3

symphony/lib/boot/func.utilities.php
jensscherbl: I guess it's ok to keep it simple. Developers using this should know what they're doing (and will notice pretty fast if something went wrong).

symphony/lib/boot/func.utilities.php
jensscherbl: What do you mean exactly? I try to stick with `PSR-2` wherever possible, otherwise aiming for better code readability.

symphony/lib/boot/func.utilities.php
michael-e: I think that these specs are pretty clear, and omitting the cache headers could only be an issue in case of _mis_configuration. I wonder if we should care about servers or proxies that don't follow the specs. (You can not win that battle.)

jensscherbl: > left up to the configuration of the developers environment

From my understanding, it would be up to the client to behave correctly (or not) given a certain status code.

jensscherbl:
@brendo Thanks for the code review, by the way, really appreciate it.

symphony/lib/boot/func.utilities.php
brendo: Fair enough. I find it the opposite. Splitting a 4 line function over multiple lines with the extra whitespace messes with the grouping in my head. For example I find the top more readable then the bottom, especially when scanning through code quickly (either by scrolling or with Sublime's preview)

```php
// Comment
if (condition) {
 // do condition
}
```

Instead of

```php
// Comment

if (condition)

// do comment

}
```

symphony/lib/boot/func.utilities.php
brendo: Unless the `$location` is also using the idn style domain?

symphony/lib/boot/func.utilities.php
michael-e: I agree with @brendo here. Empty lines are for grouping, which doesn't work if you use too many of them.

symphony/lib/boot/func.utilities.php
jensscherbl: Oh, you mean if the `$location` uses another internationalized domain name? You're right, will fix that.

symphony/lib/boot/func.utilities.php
jensscherbl: Are we talking about my use of whitespace in general here, or only around comments?

michael-e: Both green blocks, I guess.

jensscherbl: Just to make sure I got it right, you prefer

    if (headers_sent($filename, $line)) {
        // throw exception if headers already sent
        throw new SymphonyErrorPage(sprintf(
            'Cannot redirect to <a href="%s">%s</a>. Output has already started in %s on line %s.',
            $location, $location, $filename, $line
        ));
    }

    if (function_exists('idn_to_ascii')) {
        // convert idn back to ascii for redirect
        $location = str_replace(HTTP_HOST, idn_to_ascii(HTTP_HOST), $location);
    }

over

    if (headers_sent($filename, $line)) {

        // throw exception if headers already sent

        throw new SymphonyErrorPage(sprintf(

            'Cannot redirect to <a href="%s">%s</a>. Output has already started in %s on line %s.',

            $location, $location, $filename, $line
        ));
    }

    if (function_exists('idn_to_ascii')) {

        // convert idn back to ascii for redirect

        $location = str_replace(HTTP_HOST, idn_to_ascii(HTTP_HOST), $location);
    }

in terms of readability?

symphony/lib/boot/func.utilities.php
michael-e: @jensscherbl: Do I remember correctly? It looks to me as if GitHub lost some comments here (probably because of the recent attack). Am I right?

michael-e: Ah, I see now, my email link has "expired"… Line numbers have changed.

jensscherbl: I pushed new commits to address your feedback.

PR #2411, opened 2015-04-29
make custom 500 pages have a 500 error message

jonmifsud:
When there is a template file available, Symphony did not automatically set the 500 Error message. I'm not sure if this was intended for the developer to set the error header within the file overrides but I think it should be handled within the core before applying any custom templates.


brendo:
I think we should actually fix this by removing the header line altogether and instead let the [errorhandler](https://github.com/jonmifsud/symphony-2/blob/custom500fix/symphony/lib/core/class.errorhandler.php#L119-L133) deal with it. What do you think?

nitriques:
@brendo Those line were added because the HTTP status code was 200. If the user wants another code, he should use the status code property of the exception.

jonmifsud:
@brendo / @nitriques not sure what happened with this PR seems like it got messed up (probably needs a rebase / to be redone)

@brendo not sure I'm following.. maybe the 200 status code would already be sent when using a custom template within workspace? All our custom 500 error pages were being served as 200's maybe there's actually a bug elsewhere?

brendo:
I guess the only way to know for sure is to retest the current build and see if the issue persists

nitriques:
> not sure what happened with this PR seems like it got messed up (probably needs a rebase / to be redone)

Yup. @brendo rebased integration and now your branch is out-of-date.

What I would do is:

1. Checkout a fresh copy of the repo.
1. Checkout the integration branch.
1. git checkout -b new-branch-name
1. git cherry-pick individual commits, starting from the oldest.
1. create a new PR.

Thanks!

nitriques:
https://github.com/DeuxHuitHuit/symphony-2/tree/errorhandler should take care of this.

Issue #2398, opened 2015-04-05
(Sort of) memory leak with entry objects


michael-e:
I am importing 5000 entries using the PHP CLI. The script works with Symphony and is supposed to build the core of a web-based (frontend) importer later, so I care about memory consumption.

The initial memory consumption of the script is around 47 MB. (This is caused by parsing a rather big Excel XML worksheet, so that is ok.)

Then I start saving entries to Symphony basically using the following code:

```php
foreach ($data as $index => $fields) {
    $entry = EntryManager::create();
    $entry->set('section_id', $section_id);
    $entry->set('author_id',  $author_id);
    if (!empty($fields['id'])) {
        $entry->set('id', $fields['id']);
    }

    // Initial checks to see if the Entry is ok
    if (__ENTRY_FIELD_ERROR__ == $entry->checkPostData($fields, $errors)) {
        foreach ($errors as $field_id => $message) {
            echo 'ERROR  : entry ' . ($index + 1) . ' field ' . $field_id . ': ' . $message . "\n";
        }
    // Secondary checks, this will actually process the data
    } elseif (__ENTRY_OK__ != $entry->setDataFromPost($fields, $errors)) {
        foreach ($errors as $field_id => $message) {
            echo 'ERROR  : entry ' . ($index + 1) . ' field ' . $field_id . ': ' . $message . "\n";
        }
    } else {
        // attempt to save
        if (!$entry->commit()) {
            echo 'ERROR  : entry ' . ($index + 1) . "\n";
        } else {
            Symphony::ExtensionManager()->notifyMembers('EntryPostEdit', '/publish/edit/', array(
                'section' => SectionManager::fetch($section_id),
                'entry' => $entry,
                'fields' => $fields
            ));
            echo 'Success: entry ' . ($index + 1) . "\n";
        }
    }

    unset($entry);
    unset($fields);

    echo number_format(memory_get_peak_usage(true), 0, '.', ',') . " bytes\n";
}

```

Memory consumption stays at the initial level of **47 MB** for around 150 entries, then starts growing slowly. It reaches **437 MB** (!) for entry number 5000.

I did some more tests:

* Using `$entry->checkPostData` and `$entry->setDataFromPost` only (i.e. without `$entry->commit()`), memory still grows, but much slower. It reaches **55 MB** for entry number 5000.
[...]

PR #2414, opened 2015-04-30
Added XHTML page type and content type

jensscherbl:
With [IE8 slowly on it's way out][2], is there any reason not to support and encourage proper [XHTML5][1]? Especially since XSLT forces us to write well-formed XML anyway.

In a [perfect world][3], this would be a pull request for a `2.7.x` release, but I guess we [aren't really that strict anyway][4]?

[1]: http://www.xmlplease.com/xhtml/xhtml5polyglot/
[2]: http://caniuse.com/#feat=xhtml
[3]: https://github.com/symphonycms/symphony-2/issues/2398#issuecomment-91020048
[4]: https://github.com/symphonycms/symphony-2/issues/2398#issuecomment-95013766

### DIFF
--- a/symphony/lib/toolkit/class.frontendpage.php
+++ b/symphony/lib/toolkit/class.frontendpage.php
@@ -219,6 +219,9 @@ class FrontendPage extends XSLTPage
                 else if(General::in_iarray('JSON', $this->_pageData['type'])) {
                     $this->addHeaderToPage('Content-Type', 'application/json; charset=utf-8');
                 }
+                else if(General::in_iarray('XHTML', $this->_pageData['type'])) {
+                    $this->addHeaderToPage('Content-Type', 'application/xhtml+xml; charset=utf-8');
+                }
                 else{
                     $this->addHeaderToPage('Content-Type', 'text/html; charset=utf-8');
                 }
--- a/symphony/lib/toolkit/class.pagemanager.php
+++ b/symphony/lib/toolkit/class.pagemanager.php
@@ -647,7 +647,7 @@ class PageManager
      */
     public static function fetchAvailablePageTypes()
     {
-        $system_types = array('index', 'XML', 'JSON', 'admin', '404', '403');
+        $system_types = array('index', 'XML', 'JSON', 'XHTML', 'admin', '404', '403');
 
         $types = PageManager::fetchPageTypes();
 


nitriques:
> With IE8 slowly on it's way out, is there any reason not to support and encourage proper XHTML5? 

Yes there is. XHTML will crash if there is malformed xml in the document. With CDATA's and third party  data, I often get invalid things.

+ I also use the html render in my xslt templates and it does not produce valid xml. I like the idea of having it as an option (I would pull your pr in for 2.6.2) but I would never vote for it to be the default.

jensscherbl:
> Yes there is. XHTML will crash if there is malformed xml in the document. With CDATA's and third party data, I often get invalid things.

- Isn't *CDATA* supposed to prevent exactly this?
- Isn't well-formed XML required for the XSLT template to be rendered in the first place?

jensscherbl:
To clarify, this is an addition, not a replacement for the default `text/html` content type.

nitriques:
> To clarify, this is an addition, not a replacement for the default text/html content type.

Yes totally and I agree with it. My thoughts were about "I can't be the default".

> Isn't CDATA supposed to prevent exactly this?

The CDATA part is in the generated xml. It never gets outputted to the browser wrapped in CDATA.

> Isn't well-formed XML required for the XSLT template to be rendered in the first place?

Yes the generated xml is valid xml. But the result of the xslt transformation does not output valid xml.

If the browser sees invalid xml in your xhtml document, it will crash. In html, if there is malformed element, the browser will degrade gracefully. But in XHTML mode, it does not.

michael-e:
> The CDATA part is in the generated xml. (…) Yes the generated xml is valid xml. But the result of the xslt transformation does not output valid xml.

I would like to know how you achieve this. I was pretty sure that CDATA may even contain KFC chicken wings without breaking anything.

nitriques:
>  was pretty sure that CDATA may even contain KFC chicken wings without breaking anything.

Yes totally. But that KFC may break the xml validity of the response sent to the client...

In the Data-source:

```xml
<entry>
   <kfc>... cdata'd data </kfc>
</entry>
```

In my page

```xslt
<xsl:value-of select="entry/kfc" disable-output-escaping="yes" />
```

Then in the html response you get the un-escaped, un-cdata KFC chicken that would ruin any XHTML page but would render fine in any HTML doc.

michael-e:
You mean like this?

```xml
<entry>
   <kfc><![CDATA[kcf chicken wings]]></kfc>
</entry>
```

I thought that even using `disable-output-escaping="yes"` (which one should avoid if possible, but that's a different story) wouldn't attempt to make real food out of the chicken wings.

Maybe you are right. I have no time to test this at the moment, unfortunately.

nitriques:
> You mean like this?

Yup.

> I thought that even using disable-output-escaping="yes" (which one should avoid if possible, but that's a different story) wouldn't attempt to make real food out of the chicken wings.

Ohhh yes it does. I have a recaptcha DS source that represent a consensual XSS. In the DS, I have:

```xml
<entry>
   <kfc><![CDATA[
     <script>...</script>
     <link href="malformed>
   ]]></kfc>
</entry>
```

with `disable-output-escaping="yes"` I get valid html, but not valid xml (xhtml).

I sadly need this. A lot.

michael-e:
Wow. how bad. :-(

Nevertheless, I just asked out of curiosity. I also wouldn't suggest to make the new content-type the default.  But I think it's a nice addition.

nitriques:
> But I think it's a nice addition.

Me too! ;)

brendo:
I'm inclined not to pull this as the Content Type Mappings can fulfil this requirement.

jensscherbl:
> I'm inclined not to pull this as the Content Type Mappings can fulfil this requirement.

What content type mappings?

michael-e:
The [Content Type Mappings](https://github.com/symphonists/content_type_mappings) extension.

jensscherbl:
> The Content Type Mappings extension.

Oh, and unmaintained extension from the Symphony abandoned extensions graveyard.

Will try it out, thanks.

Still, any good reasons why XML, JSON or HTML should be hardcoded / provided by the core, but XHTML or even plaintext needs an extension?

Would it be a good idea to get rid of the hardcoded content types and integrate content type mappings into the core instead? I mean, this is a pretty fundamental thing for pages, isn't it?

brendo:
Be careful. Just because there's no commits doesn't mean it doesn't
function.

jensscherbl:
> Just because there's no commits doesn't mean it doesn't
function.

jensscherbl:
> Just because there's no commits doesn't mean it doesn't function.

Yeah. I also know that you are basically maintaining most of these extensions yourself.

But since there's no one who has taken ownership of the extension and is directly responsible, and it hasn't been marked to be compatible with the current version, it's a bit like russian roulette... ;)

The [The League of Extraordinary Packages][1] has a [similar concept for unmaintained packages][2], by the way...

> If the maintainer of a package decides for any reason that they no longer want to maintain it, the Leadership Group will assist in finding a new maintainer, if the original owner would like to leave their package with the League. If a new owner cannot be found within a reasonable timeframe, the project will be moved to the **Cemetery**, until a new Project Lead turns up, or they might decide to maintain it themselves.

[1]: http://thephpleague.com
[2]: https://github.com/thephpleague/thephpleague.github.io/wiki/Leadership-Group#responsibilities

nitriques:
> The League of Extraordinary Packages has a similar concept for unmaintained packages, by the way...

I really like that idea. Where's should this infos be ? I think the best spot would be the extension site, but we do not have access to it to change it easily.

jensscherbl:
Closing in favor of #2417.

PR #2417, opened 2015-05-01
Added configurable content types

jensscherbl:
- Removed hardcoded content types from Symphony's system page types.
- Content types are now configurable in the `config.php`.
- Existing core content types are pre-configured.
- Added `text/plain` and `application/xhtml+xml` as well.
- Fallback is `text/html` if a content type is missing from page types or `config.php`.
- Added new system page type `attachment` that triggers download of page/resource.
- Makes [Content Type Mappings][1] obsolete from a functional point of view.
- Doesn't provide a UI on the preferences page like Content Type Mappings does, so the extension could be updated to provide its UI on top of the core functionality for people who need it.

See discussion #2414 for reference.

Makes sense?

[1]: https://github.com/symphonists/content_type_mappings

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,27 +1,46 @@
 {
-    "name": "symphonycms/symphony-2",
-    "version": "2.6.1",
-    "description": "Symphony is a PHP & MySQL based CMS that utilises XML and XSLT as its core technologies.",
-    "homepage": "http://www.getsymphony.com",
-    "license": "MIT",
-    "authors": [
+    "name"        : "symphonycms/symphony-2",
+    "version"     : "2.6.1",
+    "description" : "Symphony is a PHP & MySQL based CMS that utilises XML and XSLT as its core technologies.",
+    "homepage"    : "http://www.getsymphony.com",
+    "license"     : "MIT",
+    "authors"     : [
+
         {
-            "name": "Symphony Team",
-            "email": "team@getsymphony.com"
+            "name"  : "Symphony Team",
+            "email" : "team@getsymphony.com"
         }
     ],
-    "support": {
-      "forum": "http://www.getsymphony.com/discuss/",
-      "issues": "https://github.com/symphonycms/symphony-2/issues",
-      "wiki": "https://github.com/symphonycms/symphony-2/wiki"
+
+    "support" : {
+
+        "forum"  : "http://www.getsymphony.com/discuss/",
+        "issues" : "https://github.com/symphonycms/symphony-2/issues",
+        "wiki"   : "https://github.com/symphonycms/symphony-2/wiki"
     },
-    "minimum-stability": "stable",
-    "autoload": {
-      "classmap": ["symphony/content", "symphony/lib", "symphony/template", "install"],
-      "files": [
-        "symphony/lib/boot/func.utilities.php",
-        "symphony/lib/boot/defines.php",
-        "symphony/lib/toolkit/util.validators.php"
-      ]
+
+    "minimum-stability" : "stable",
+
+    "autoload" : {
+
+        "classmap" : [
+
+            "symphony/content",
+            "symphony/lib",
+            "symphony/template",
+            "install"
+        ],
+
+        "files" : [
+
+            "symphony/lib/boot/func.utilities.php",
+            "symphony/lib/boot/defines.php",
+            "symphony/lib/toolkit/util.validators.php"
+        ]
+    },
+
+    "require" : {
+
+        "aura/accept" : "2.2.*"
     }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,79 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
+    ],
+    "hash": "878e6a5827844ab0ef81b652b3f94a4e",
+    "packages": [
+        {
+            "name": "aura/accept",
+            "version": "2.2.2",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/auraphp/Aura.Accept.git",
+                "reference": "1c40bc38333a7899b95823a4d85ab758f7c618b2"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/auraphp/Aura.Accept/zipball/1c40bc38333a7899b95823a4d85ab758f7c618b2",
+                "reference": "1c40bc38333a7899b95823a4d85ab758f7c618b2",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=5.3.0"
+            },
+            "require-dev": {
+                "aura/di": "~2.0"
+            },
+            "type": "library",
+            "extra": {
+                "aura": {
+                    "type": "library",
+                    "config": {
+                        "common": "Aura\\Accept\\_Config\\Common"
+                    }
+                },
+                "branch-alias": {
+                    "dev-develop-2": "2.0.x-dev"
+                }
+            },
+            "autoload": {
+                "psr-4": {
+                    "Aura\\Accept\\": "src/",
+                    "Aura\\Accept\\_Config\\": "config/"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "BSD-2-Clause"
+            ],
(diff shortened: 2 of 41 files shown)


nitriques:
> Makes sense?

Rapidly, I think I does, yes.

jonmifsud:
Seems to make sense, will make content type mapping extension less
important/for more extreme cases example using the same page to output
different formats eg html and csv versions of the same page. I guess that's
not doable with this core change right?

jensscherbl:
> using the same page to output different formats eg html and csv versions of the same page. I guess that's not doable with this core change right?

You mean content negotiation? No, but that's not possible with the Content Type Mappings extension either. Just found [an old extension][1] that did that. Maybe that can be baked in too, either for `2.7.0` or `3.0.0`. Will have a look soon.

[1]: https://github.com/klaftertief/content_negotiation/

jensscherbl:
> Maybe that can be baked in too, either for 2.7.0 or 3.0.0. Will have a look soon.

Thinking more about it, while content types are very fundamental to pages, content negotiation is more or less nice to have and should be left to an (new or updated) extension.

michael-e:
It's s difficult matter, I mean: where to start and where to stop. For example, what about response headers? My [HTTP Response Header Mappings](https://github.com/michael-e/http_response_header_mappings) extension allows to set headers like `HTTP/1.1 503 Service Unavailable` based on the page type.

jensscherbl:
> where to start and where to stop

Start where the core makes assumptions, hardcodes some options and needs an extension to overwrite these or add your own. Stop where an extension adds features that are not needed in every project.

> what about response headers

Looking at this, it's a more general approach to what content type mappings did and might be an even better fit for core integration.

jensscherbl:
What about this?

- Make content types configurable.
- Take accept headers into account to provide a foundation for content negotiation.
- Don't make status codes configurable. Instead, just let the core handle all [status codes][1], so no need for an extension to provide proper status headers.
- Provide an extension called "HTTP Response Headers" that handles everything else.
- Provide an extension called "Attachments" that provides the attachment page type (currently part of this pull request).
- Provide an extension called "Content Negotiation" that provides multiple templates for pages with multiple content types.

[1]: http://en.wikipedia.org/wiki/List_of_HTTP_status_codes

michael-e:
So in the end I need a lot of extensions, just like now? At the moment, HTTP Response Headers provide attachments functionality as well!

jensscherbl:
> So in the end I need a lot of extensions, just like now?

You specifically for your particular use case, yes. Most other people for most other projects, no.

You'll still need extensions if you need more advanced settings and features, but not for the most fundamental stuff that the core already does (in a hardcoded and not configurable way).

> HTTP Response Headers provide attachments functionality as well!

And the Content Type Mappings extension provides this as well. That's why we should aim for a more modular ecosystem. A single extension for a single task.

michael-e:
I myself wouldn't like to have one extension per Micro-Task. It doesn't make things simpler in the end.

jensscherbl:
> I myself wouldn't like to have one extension per Micro-Task. It doesn't make things simpler in the end.

You could still use your one extension that does everything.

michael-e:
But I think we should try and get rid of many extensions instead of creating additional ones. My extension can do everything except content negotiation, and it's pretty simple. Isn't that a good starting point to simplify things?

nitriques:
> But I think we should try and get rid of many extensions instead of creating additional ones. 

+1

jensscherbl:
> But I think we should try and get rid of many extensions instead of creating additional ones.

Where it makes sense, yes. Not just for the sake of it.

> My extension can do everything except content negotiation, and it's pretty simple.

Doing everything isn't simple. Don't mix up convenience with good software architecture.

> Isn't that a good starting point to simplify things?

It is. And it's indeed simple when treating everything just as different HTTP headers. But to be honest, I don't like how it mixes content types (fundamental), status codes (also fundamental) and other uses cases (additional functionality) into a single setting.

Let the core handle the fundamental things (no extension) and simplify your extension even further to only provide the additional functionality.

The attachment type isn't just a HTTP header because it needs extra logic to determine an appropriate filename. That can either be part of the core or a very simple and easy to maintain extension. I'd vote for the latter.

Not sure about what to do regarding different templates for different content types, but let's take one step after another.

jonmifsud:
> Not sure about what to do regarding different templates for different
content types, but let's take one step after another.

I am not sure everyone is currently using this but when I required it I
opted for a URL parameter (probably could use url-routing + params) and
then I would use the url parameter to switch the template within the XSLT,
nothing too complicated. Obviously you'd have to switch headers according
to the parameter but nothing too complicated.

jensscherbl:
You can always serve different content types with different URLs, but that's not what [content negotiation][1] is about... ;)

[1]: http://en.wikipedia.org/wiki/Content_negotiation

jensscherbl:
- [x] Configurable content types.
- [x] Content negotiation based on HTTP accept header.
- [x] Attachments.
- [ ] HTTP status codes.
- [ ] Extension "HTTP Response Headers".
- [ ] Different templates for different content types.

jensscherbl:
> then I would use the url parameter to switch the template within the XSLT

Problem I see with this approach is that you can't use parameters in `xsl:import`, so you always import a bunch of unused templates.

Is this an issue? Or [premature optimization][1]?

[1]: http://en.wikipedia.org/wiki/Program_optimization#When_to_optimize

michael-e:
@jensscherbl: Removing the charset from the configuration is no good idea in my eyes, because XSLT can use different character sets for output. I once had to use UTF-16 for CSV data created via XSLT, for example.

brendo:
I'm not overly fond of this change to be completely honest, I'm satisfied with what the core offers and what extensions are available to execute the other use cases discussed in this thread.

jensscherbl:
> I'm satisfied with what the core offers

I'm not. Can you please explain why JSON has to be a hardcoded content type in the core, but for using XHTML in an XML and XSLT centric CMS I need an extension to overwrite the hardcoded defaults?

jensscherbl:
> I once had to use UTF-16 for CSV data created via XSLT, for example.

You mean you actually had to create the data/encoding in XSLT? oO How? (Just curious...)

brendo:
> I'm not. Can you please explain why JSON has to be a hardcoded content type in the core, but for using XHTML in an XML and XSLT centric CMS I need an extension to overwrite the hardcoded defaults?

JSON and XML are present to allow developers to generate pages that can be consumed by Dynamic/Remote Datasource. 

The proper XHTML content type is not used historically because of poor browser support. These days, I see it as an edge case considering best practice is to use the HTML 5 doctype. Using the HTML 5 doctype is not compatible when sending the Content Type as `application/xhtml+xml` so including it as a default option feels odd.

Nevertheless, this PR is now more than just an XHTML content type being added. It's grown to encompass multiple other extension's functionality and I don't believe it fits with the philosophy of Symphony.

michael-e:
@jensscherbl:

> You mean you actually had to create the data/encoding in XSLT? oO How? (Just curious…)

Oh, pretty simple:

```xml
<xsl:output method="text"
	encoding="utf-16"/>
```

Regarding this PR, I must say that I agree with @brendo. I think that you are going in the wrong direction here.

jensscherbl:
> JSON and XML are present to allow developers to generate pages that can be consumed by Dynamic/Remote Datasource.

- Which are no longer part of the core, right?
- JSON and XML are still present by default in this pull request.

> Using the HTML 5 doctype is not compatible when sending the Content Type as application/xhtml+xml

That's not correct.

> Nevertheless, this PR is now more than just an XHTML content type being added.

Because you were against adding another hardcoded content type and I agree. That's why I made content types configurable by design instead.

> It's grown to encompass multiple other extension's functionality

Other extensions that were previously necessary to do basic HTTP stuff and overwrite a few hardcoded options in the core.

> and I don't believe it fits with the philosophy of Symphony

*Don't make assumptions*?* Or *flexibility* and *complete control*? That's exactly what this pull request is about.

jensscherbl:
> Oh, pretty simple

Basically converting UTF-8 content to UTF-16 during output creation, correct?

Seems to be a valid use case then.

michael-e:
Yes, correct, it's simply setting the _output encoding_.

jensscherbl:
> Yes, correct, it's simply setting the output encoding.

I tested with actual UTF-16 data internally and ran into a lot of issues, so I assumed UTF-8 was the only choice anyway.

jensscherbl:
Closing due to disagreement.

PR #2420, opened 2015-05-04
Escape all commas in filtering suggestions.

### LINKED ISSUE #2349: New filter interface: Filtering on values containing commas returns no results
Using the new entry filters (integration branch), filtering on values containing commas always returns no results. It works fine for values without commas.

michael-e:
The previous fix replaced the first comma only.

Fixes #2349.

### DIFF
--- a/symphony/assets/js/src/symphony.suggestions.js
+++ b/symphony/assets/js/src/symphony.suggestions.js
@@ -325,7 +325,7 @@
 				insert(value, input);
 			}
 			else {
-				input.val(value.replace(',', '\\,'));
+				input.val(value.split(',').join('\\,'));
 				input.addClass('updated');
 				input.change();
 			}


nitriques:
I would use a regular expression in order to replace ALL of them. ?

```js
input.val(value.replace(/,/g, '\\,'));
```

Which one *is* better ?? :smile:

michael-e:
I don't know which one is _better_, but isn't regex slower, most of the time? (And the regex `g` modifier didn't work in my quick test, don't ask me why.)

nitriques:
> And the regex g modifier didn't work in my quick test

Is the value multiline ?

michael-e:
No, I don't htink so.

nitriques:
I get

```js
var value = 'test1,test2,test3';
value.replace(/,/g, '\\,'));
"test1\,test2\,test3"
```

You're in Safari ?

nitriques:
But it looks likes it's only a tiny-tiny bit faster. http://jsperf.com/regexp-vs-split-join Leave it with split...

michael-e:
Your test code works fine in the Safari console. Let's see what @brendo thinks.

nitriques:
Yup.

brendo:
Happy with the perf option, thanks @michael-e

PR #2421, opened 2015-05-07
Fix author email already exists

jonmifsud:
This is similar to an issue which was fixed with the Username. As fetchVar returns a string it always returns false, stating that the email is duplicate.


nitriques:
LGTM.

brendo:
Thanks!

Issue #2422, opened 2015-05-10
[integration] Fatal database errors


designermonkey:
Anyone else getting fatal database errors since the PDO upgrade?

```
Symphony Fatal Database Error: SQLSTATE[42000]: Syntax error or access violation: 1103 Incorrect table name ''
An error occurred while attempting to execute the following query
DELETE FROM ``sym_fields_input`` WHERE `field_id` = ? LIMIT 1
Backtrace
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/class.database.php:491]
    Database->error();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/class.database.php:345]
    Database->q();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/class.mysql.php:413]
    Database->delete();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/class.fieldmanager.php:150]
    MySQL->delete();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/fields/field.input.php:126]
    FieldManager::saveSettings();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/content/content.blueprintssections.php:805]
    FieldInput->commit();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/class.administrationpage.php:703]
    contentBlueprintsSections->__actionNew();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/class.administrationpage.php:666]
    AdministrationPage->__switchboard();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/toolkit/class.administrationpage.php:448]
    AdministrationPage->action();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/content/content.blueprintssections.php:26]
    AdministrationPage->build();
[/Users/designermonkey/Projects/Sites/symphony.local.embarknow.net/htdocs/symphony/lib/core/class.administration.php:212]
    contentBlueprintsSections->build();
[...]

PR #2426, opened 2015-05-12
Remove backticks from field table name

designermonkey:
As per #2422

### DIFF
--- a/symphony/lib/toolkit/class.fieldmanager.php
+++ b/symphony/lib/toolkit/class.fieldmanager.php
@@ -147,7 +147,7 @@ class FieldManager implements FileResource
         $type = self::fetchFieldTypeFromID($field_id);
 
         // Delete the original settings:
-        Symphony::Database()->delete("`tbl_fields_".$type."`", "`field_id` = ? LIMIT 1", array($field_id));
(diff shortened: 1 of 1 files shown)


brendo:
@nitriques I know this was originally added by you for consistency with `insert()` and `update()` methods, but with PDO it looks like it's not required and it's merely used as a best practice.

nitriques:
> it's merely used as a best practice.

Even if the table names contains weird characters ? Wasn't this the goal ? 

Anyhow, less if more so I am all up for this ! But we should do it everywhere, not only on a single statement....

michael-e:
I don't get this. I always thought that backticks – even if they may not be required – won't cause any issues. Why have they been removed here?

designermonkey:
PDO will automatically wrap table and column names with backticks, so providing them only causes an error, as a table would be double wrapped.

michael-e:
Aha, thanks! So the backticks should probably be removed everywhere, correct?

designermonkey:
only in the table names passed through. in the rest of the queries, they can either be there or not. I was only error-preventing here.

nitriques:
Ok Thanks John.

PR #2427, opened 2015-05-12
Add delegate for composer to allow use of autoloaded extensions

designermonkey:
Using composer means that classes can be added to the autoloader. This delegate allows extensions to use autoloaded classes in their code.

The extension driver need only listen to the delegate and run something like the following

```php
$context['autoloader']->setPsr4('SymphonyCMS\\Extensions\\MyExtensionClass\\', EXTENSIONS . '/extension_name/libs');
```

Based on work by @psychoticmeow

### DIFF
--- a/index.php
+++ b/index.php
@@ -9,10 +9,24 @@
     ));
 
     // Include autoloader:
-    require_once DOCROOT . '/vendor/autoload.php';
+    $autoloader = require_once DOCROOT . '/vendor/autoload.php';
 
     // Include the boot script:
     require_once DOCROOT . '/symphony/lib/boot/bundle.php';
+
+    /**
+     * When the composer autoloader has been loaded.
+     *
+     * @delegate ComposerReady
+     * @param string $context
+     *  '/all/'
+     * @param Composer\Autoload\ClassLoader $autoloader
+     *  The Composer autoloader.
+     * @since 3.0.0
+     */
+    Symphony::ExtensionManager()->notifyMembers('ComposerReady', '/all/', [
+        'autoloader' => $autoloader
+    ]);
 
     // Begin Symphony proper:
     symphony(APP_MODE);


psychoticmeow:
@designermonkey Sorry man, didn't notice you'd made this so I've made my own pull request that's slightly different. See #2428.

designermonkey:
Damn you, stealing work I stole from you.

jensscherbl:
Do I get this right, the composer autoloader can be modified during runtime? What about an extension's dependencies?

psychoticmeow:
It all happens during runtime, so yeah. What's your concern with extension dependencies?

jensscherbl:
> What's your concern with extension dependencies?

How do I add third party dependencies to the global autoloader (without replicating each dependencies composer file)?

psychoticmeow:
Oh, that's not really what I'm trying to solve here, I want to allow autoloading of the extensions own libraries because when you have a lot of extensions installed that explicitly include their classes it tends to slow down page load time somewhat.

designermonkey:
Yeah, @jensscherbl if you want to load dependencies properly, use the composer file, but for extensions with many class files, use the delegate. It's easier and more performant.

jensscherbl:
> I want to allow autoloading of the extensions own libraries because when you have a lot of extensions installed that explicitly include their classes it tends to slow down page load time somewhat.

Yeah sure, but what if the extension uses its own composer file, vendor folder and autoloader to manage its own third party dependencies?

For example...

- https://github.com/jensscherbl/commonmark
- https://github.com/jensscherbl/htmlpurifier

How would this be handled?

psychoticmeow:
In that situation it makes sense to include the extensions composer autoloader at the top of the extension driver.

jensscherbl:
> In that situation it makes sense to include the extensions composer autoloader at the top of the extension driver.

Ah ok. So everything stays as is for these extensions.

PR #2429, opened 2015-05-13
set seed to that at start of ds processing

jonmifsud:
The seed value gets overwritten by any DatasourcePreExecute and DatasourcePostExecute delegates, thus the timings would be incorrect. This change ensures that the correct profile start time is used when generating profile timing for datasources.


nitriques:
Thanks Jon!

PR #2430, opened 2015-05-13
Added a `Database::prepare` method and a PDOStatement wrapper

psychoticmeow:
Added a `Database::prepare` method and a PDOStatement wrapper so that we can integrate prepared statements.

This patch is PHP 5.6 compatible, more work will be needed to support earlier versions, but since PHP 5.5 will be officially unsupported as of next month, and I'm not sure what Symphony is expecting to support I've left it for later.

### DIFF
--- a/symphony/lib/toolkit/class.database.php
+++ b/symphony/lib/toolkit/class.database.php
@@ -64,20 +64,56 @@ Class DatabaseException extends Exception{
 
 class DatabaseStatement
 {
+    /**
+     * The database connection.
+     *
+     * @var Database
+     */
     public $database;
+
+    /**
+     * The internal PDOStatement.
+     *
+     * @var PDOStatement
+     */
     public $statement;
 
+    /**
+     * Wrap a PDOStatement object so that we can log its queries
+     * and handle its errors.
+     *
+     * @param Database      $database
+     * @param PDOStatement  $statement
+     */
     public function __construct(Database $database, PDOStatement $statement)
     {
         $this->database = $database;
         $this->statement = $statement;
     }
 
+    /**
+     * Call a method on the internal PDOStatement object.
+     *
+     * @link    http://php.net/manual/en/class.pdostatement.php
+     *  For a list of available methods.
+     * @param   string $name
+     * @param   array  $arguments
+     * @return  mixed
+     */
     public function __call($name, $arguments)
     {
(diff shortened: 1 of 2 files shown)


brendo:
Doc blocks my boy ;) Then it's sweet to be merged!

psychoticmeow:
There we go.

brendo:
Boom, thanks! Thanks for stripping all the trailing space too, I must of messed that up on an earlier rebase.

symphony/lib/toolkit/class.database.php
```diff
@@ -62,6 +62,110 @@ public function getDatabaseErrorMessage()
...
+     *
+     * @link    http://php.net/manual/en/class.pdostatement.php
+     *  For a list of available methods.
+     * @param   string $name
+     * @param   array  $arguments
+     * @return  mixed
+     */
+    public function __call($name, $arguments)
+    {
+        return $this->statement->{$name}(...$arguments);
```
brendo: Actually, I missed this. I think we'll have to use `func_get_args()` as splats at PHP 5.6+

symphony/lib/toolkit/class.database.php
```diff
@@ -62,6 +62,110 @@ public function getDatabaseErrorMessage()
...
+
+    /**
+     * Executes the internal PDOStatement.
+     *
+     * @link    http://php.net/manual/en/pdostatement.execute.php
+     *  For complete documentation.
+     * @throws  DatabaseException
+     * @return  boolean
+     */
+    public function execute(...$arguments)
```
brendo: And here.

symphony/lib/toolkit/class.database.php
```diff
@@ -62,6 +62,110 @@ public function getDatabaseErrorMessage()
...
+    {
+        $start = precision_timer();
+
+        $this->database->flush();
+
+        $query = $this->statement->queryString;
+        $hash = md5($query . $start);
+
+        try {
+            $result = $this->statement->execute(...$arguments);
```
brendo: And here. Doh, should of taken a closer look :)

symphony/lib/toolkit/class.database.php
designermonkey: would it not be `call_user_func_array` then they can be passed through as they are?

Issue #2433, opened 2015-05-14
Switch to InnoDB


psychoticmeow:
We should switch to MySQLs InnoDB and MariaDBs XtraDB.

We would be able to make use of:

- Row-level locking instead of table level locking allows for faster writing of data.
- Transactions and save points would allow us to use nested transactions and correctly roll back failed actions to avoid data corruption.
- Foreign key support means that if the tables are defined correctly all you have to do to delete an entry and all of its data is a simple `delete from tbl_entries where id = #`.
- Better protection against table corruption at the expense of speed *or* similar protection against table corruption that MyISAM offers to enable better read performance.

nitriques:
> Foreign key support means that if the tables are defined correctly all you have to do to delete an entry and all of its data is a simple 

IMHO, delete cascade is not a good idea...

But I like everything else.

Also, to the cache table, I would recommend using memory storage in order to create a hash.

see https://github.com/symphonycms/symphony-2/issues/2163#issuecomment-52266694

psychoticmeow:
Here's a discussion about moving to InnoDB from 2012: #1238

michael-e:
From that discussion: Switching to InnoDB  lifts the requirement to MySQL 5.6.4. Is that right?

It's not a problem for me (since I am using MariaDB 10 anyway), but what about shared hosting? Will this requirement be an issue?

psychoticmeow:
It was released in late 2011, if it's actually a problem I'd be very surprised, especially since there's been 20 point releases since then. We're also using MariaDB 10, which is roughly equivalent to later 5.6 releases anyway.

nitriques:
> Here's a discussion about moving to InnoDB from 2012:

Is fulltext indexing still an issue with InnoDB ? Seems like 5.6 only ... https://dev.mysql.com/doc/refman/5.6/en/innodb-fulltext-index.html

And still, I do not have any answer for: "delete cascade is not a good idea..".. is this really planned ?

I am still not sure about wheter to force InnoBD for extensions or leave them decide...

And I would love `tbl_cache` to be of type HEAP (https://dev.mysql.com/doc/refman/5.6/en/memory-storage-engine.html)

psychoticmeow:
> Is fulltext indexing still an issue with InnoDB ? Seems like 5.6 only ... https://dev.mysql.com/doc/refman/5.6/en/innodb-fulltext-index.html

[...]

PR #2440, opened 2015-05-15
Replace MyISAM with InnoDB

psychoticmeow:
Removed any reference to MyISAM and explicitly set InnoDB as the engine when connecting. As part of #2433.


nitriques:
Merged as 54a9e7d

Issue #2361, opened 2015-02-27
Date Field: "date" column should be removed


michael-e:
As @nilshoerrmann [pointed out in a different thread](https://github.com/symphonycms/symphony-2/issues/2360#issuecomment-76220694), the Date Fields "date" column is not used anymore. So it should be removed.

nitriques:
I would propose this to be 2.7.0 ? Can it be 2.6.0 ?

michael-e:
2.6.0 has RC status, so I suggest to do it in 2.6.1 (and 2.7, of course).

nitriques:
Great.

nitriques:
Fixed in 1377797

Issue #2373, opened 2015-03-12
Migration to 3.0.0


nitriques:
Since we are now working on 3.0.0, I would like to address the issues of migration.

We are planning to make breaking changes (I am thinking about handles specially).

Will we create all the migration code ? 

I think we should and I would like to have everybody's opinion on that matter.

andrewminton:
Are we talking migration of core or data?

nitriques:
Code + Data.

I mean, we need to first figure out if we want an upgrade path from 2.x to 3.x... And what should be included or not :smile:

brendo:
Migration code (should we choose to help migrate things) will live in [the migration file](https://github.com/symphonycms/symphony-2/blob/integration/install/migrations/3.0.0.php).

I think it will need to be approached on a case by case basis.

nitriques:
> I think it will need to be approached on a case by case basis.

Yeah that's for sure. But can we all agree that we should try to make migration from 2.x to 3.x easy as possible ?

michael-e:
+1

jensscherbl:
Should be added to the [3.0.0][1] milestone.

[1]: https://github.com/symphonycms/symphony-2/milestones/3.0.0

jensscherbl:
I think we have to restructure the installers `index.php` and/or the `bundle.php`, making the updater more independent and run it as early as possible.

There's a new `order` column in the `extensions_delegates` table in `3.0.0`. Problem is, the `ExtensionManager` class, with an SQL statement in its constructor that uses this new column, is called before the updater runs, resulting in a Symphony Fatal Database Error.

Not sure how to solve this, or if there are any other possible breaking points.

nitriques:
Finally, the migration will be possible only from 2.7.x LTS.
Lower version will need to first update to the LTS version, and then update to 3.0.0

PR #2441, opened 2015-05-16
Removed the 'date' column from the date field table

psychoticmeow:
Fixes issue #2361.


brendo:
I just can't help but feel we're doing it wrong by using the text value. Anyone else? I don't really have anything to justify it other than a gut feeling :confused:

michael-e:
Looking at the PR, I assume that Symphony uses the text value exclusively? But why? There might be reasons for this. Maybe @nilshoerrmann knows more, because he has worked a lot on date and time stuff.

Indeed it feels a bit strange. Unfortunately I don't know anything about the benfits and shortcomings of `DATETIME` values in MySQL. Anyway this change should be discussed. There should only be one of these two columns left, I think everybody agrees on that.

Last but not least: AFAIK, there has not been a final decision on a possible upgrade path from Symphony 2 to Symphony 3. (Correct me if I am wrong.) As long as we have the chance to provide an upgrade, shouldn't structural changes to the database be replicated in the updater, too?

nitriques:
@michael-e

Using string was the only way to "remove" timezones...

>  AFAIK, there has not been a final decision on a possible upgrade path from Symphony 2 to Symphony 3.

See https://github.com/symphonycms/symphony-2/issues/2373

nitriques:
Merged as 1377797

Issue #2432, opened 2015-05-13
Late static binding... self:: vs static::


nitriques:
I've run into this bug 2 times today.... I've wanted to overwrite a static method, but could not since the code uses `self::` instead of `static::` which prevents late static binding...

Is it ok if I look around the codebase and change every `self::` into `static::` ??

cc @jensscherbl @brendo @michael-e

brendo:
It should be, but it is hard to know if we're inadvertently relying on this functionality (wrongly) anywhere though, so I worry about adding it to 2.6.3.

nitriques:
> so I worry about adding it to 2.6.3.

Ok I agree, we should wait to 3.0.0 to do it everywhere. But for 2.6.3, I would do it in the field class since they really are meant to be inherited from... And I should be able to test it properly...

It's preventing me from releasing an update to image upload extension...

nitriques:
Would you consider things like `self::__ERROR__` to fit in the 'inadvertently relying on this functionality (wrongly)' category ?

Cause I would not allow users to change constants... but would let them overwrite protected or public static function...

I would also never use in in non-static context as this can lead to bugs (see Example 3 http://php.net/manual/en/language.oop5.late-static-bindings.php)

nitriques:
@brendo Ping!

PR #2451, opened 2015-05-22
Upload Field: Change self:: to static:: to allow overrides

nitriques:
This changes will allow field that inherits from fieldUpload to provide a custom version of getMetaInfo()

This is RE: #2432 

Since we do not want to make great chances that might break, this is a really small one with impact only on classes that extends this one. 

I my case (for image_upload_field), I need this functionality.


nitriques:
Thanks Brendan!

PR #2459, opened 2015-06-20
Update README.markdown 

diomed:
adds gitter chatroom


nitriques:
Thanks, but it's already there in the [2.6.x branch](https://github.com/symphonycms/symphony-2/tree/2.6.x), which will get merge with master when 2.6.4 (or any other version) comes out.

PR #2476, opened 2015-08-08
Fix ds profile

jonmifsud:
I don't think that I had issues a PR with this before been pending a while. This corrects the amount of time each datasource takes to process, as it would be otherwise restarted by delegates running their own timing functions within a datasource; eg Association Output.


jonmifsud:
moving to integration

Issue #2466, opened 2015-07-08
Weird colors in success notice after animation


nitriques:
Is this a bug ? Or as designed ?

![image](https://cloud.githubusercontent.com/assets/771169/8579992/0044f056-2586-11e5-870a-73fa4be971e8.png)

I would rather see this:

![image](https://cloud.githubusercontent.com/assets/771169/8580002/0d9052f0-2586-11e5-9757-a46a27aa835e.png)

cc @nilshoerrmann

michael-e:
Buggg.

nitriques:
;) Let's wait for the designer's comment hahaha

jurajkapsz:
I also noticed this in webkit browsers.

nilshoerrmann:
That's by design of course! You should never be forced to create another entry ;)
No, seems like an old Webkit bug is popping up again when transitioning from dark to light colours …

nitriques:
:) Ok thanks Nils I'll try to fix this up without breaking anything else ;)

nitriques:
I am so a lame designer (even, not a graphical designer at all), I really wanted to check with you first :P

jurajkapsz:
Maybe a way to do it is through js.

nitriques:
> Maybe a way to do it is through js.

? Everything is done in css right now... Should be pretty easy to fix.

nitriques:
Finally @jurajkapsz was right: we must do it in js... this is the bug: http://stackoverflow.com/questions/8490828/css3-color-transition-not-working-in-chrome

:visited links can't be animated...

jurajkapsz:
yeah, js will be a sure thing.

nitriques:
Adding a cache buster (?cache=<timestamp>) would solve the problem. But it ain't pretty.

brendo:
> Is this a bug ? Or as designed ?

Bug.

> Adding a cache buster (?cache=) would solve the problem. But it ain't pretty.

Agreed, I don't like this. 

Is another solution to use JS to add classes? I notice the current implementation is quite elegant in that it uses an infinite CSS transition, but if it's not working then it's not totally useful to us :) Perhaps we could listen to the `transitionend` event and add the appropriate class for better compatibility?

nitriques:
> transitionend 

Good idea. But still, the css animation won't work.

johnpuddephatt:
[...]

PR #2479, opened 2015-08-11
Fixing issue 2466

johnpuddephatt:
See #2466.

### DIFF
--- a/symphony/assets/js/src/symphony.notify.js
+++ b/symphony/assets/js/src/symphony.notify.js
@@ -3,6 +3,7 @@
  */
 
 (function($, Symphony) {
+	'use strict';
 
 	/**
 	 * Notify combines multiple system messages to an interface that focusses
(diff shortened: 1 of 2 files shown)


brendo:
Hey @johnpuddephatt, can you open this against `2.6.x` instead of `integration` please?

symphony/assets/js/src/symphony.notify.js
```diff
@@ -69,6 +70,13 @@
 				storage = (window.localStorage[settings.storage]) ? $.parseJSON(window.localStorage[settings.storage]) : [];
 			}
 
+			// Dimmed success transition
+			if (!!type && !!~type.indexOf('success')) {
```
brendo: I haven't see this syntax before, what does `!!~` do? Out of curiosity :)

brendo:
Patch manually applied due to merge conflicts, thanks @johnpuddephatt!

symphony/assets/js/src/symphony.notify.js
michael-e: Seems that it's equivalent to `ToBoolean`, see: http://stackoverflow.com/questions/10467475/double-negation-in-javascript-what-is-the-purpose

Issue #2485, opened 2015-08-22
Redirect issue for datasources and events


jensscherbl:
Getting a redirect error (browser error) for `../blueprints/datasources/` and `../blueprints/events/` when using the `2.6.x` branch. Works with `master`. Can someone confirm this?

michael-e:
Yep, confirmed.

jensscherbl:
It's caused by [this][1]. @nitriques @brendo 

[1]: https://github.com/symphonycms/symphony-2/pull/2470

michael-e:
Hmmm, I still have the redirect error if I remove the placeholder. Are you really sure?

michael-e:
(I mean the `$1` replacement, sorry for being unprecise.)

jensscherbl:
Check out the commit before this commit and see if it works.

michael-e:
It still works for me some commits later than the one you mentioned: e8e92cdaaafa788f1e223e15b5c1656b6c72b7cc

michael-e:
If you use Safari, be careful with the caching. It may cache those redirects.

jensscherbl:
Yeah, using Safari. So, is e8e92cd the breaking commit then?

michael-e:
I haven't found it yet. It's hard for me to find the right one, because I always have to re-apply some commits on top — I hacked Symphony, a bit.

jensscherbl:
Hold on, I'll do it (again) so you only have to confirm what I come up with.

michael-e:
Found it! This is the commit which introduced the issue: 2ee98b19867be043b9275048c2e94ac728ac0bb0

jensscherbl:
Yep, same here.

michael-e:
Now all we have to do is find out why. You might be better than me with that.

michael-e:
Comment out [this line](https://github.com/symphonycms/symphony-2/blob/2ee98b19867be043b9275048c2e94ac728ac0bb0/symphony/lib/toolkit/class.resourcespage.php#L42), and things start working again. Don't ask me why.

jensscherbl:
You're always two steps ahead of me...

michael-e:
Sometimes I am fast in finding things. Solving issues may be a different thing.

jensscherbl:
I'll look into it.

michael-e:
Thanks!

michael-e:
Funny, this works:

```php
[...]

PR #2486, opened 2015-08-22
Fixed a regression

jensscherbl:
Fixed an issue where `null` is sometimes passed by reference and can't be assigned a new value. See #2485.


michael-e:
Thanks a lot!

PR #2489, opened 2015-08-24
Dates not formatted correctly in Select Box Link

jensscherbl:
Depending on the field's setting, dates are formatted with or without time on publish page, publish index and association drawer. For Select Box Link values, this setting has been ignored and dates were always formatted with time.

By the way, dates are also formatted even more differently when used as dynamic values for a normal Select Box field (not covered by this change).


nitriques:
LGTM

brendo:
ty!

PR #2495, opened 2015-09-04
Cleanup session cookies when redirecting

### LINKED ISSUE #2494: Symphony sends (useless) cookies when redirecting
If you use Symphony's `redirect` function (from `/symphony/lib/boot/func.utilities.php`) while a page is being built (e.g. in an extension that hooks into a frontend delegate), Symphony will issue a potentially useless `Set-Cookie` header.

This is due to the fact that `cleanup_session_cookies()` is called late in `index.php` (Sym 2.3) resp. the `symphony_launcher` function (Sym 2.6), immediately before sending the rendered output. When redirecting, this funtion won't be called at all, so the `Set-Cookie` header won't be cleaned up.

I can't think of any reason to send such a header (or any other custom page header) with a redirect, so my proposal is to add `header_remove()` to Symphony's redirect function, like so:

```php
function redirect ($url){
	// Just make sure.
	$url = str_replace('Location:', null, $url);

	if(headers_sent($filename, $line)){
		echo "<h1>Error: Cannot redirect to <a href=\"$url\">$url</a></h1><p>Output has already started in $filename on line $line</p>";
		exit;
	}

    header_remove();
	header('Expires: Mon, 12 Dec 1982 06:00:00 GMT');
	header('Last-Modified: ' . gmdate('D, d M Y H:i:s') . ' GMT');
	header('Cache-Control: no-cache, must-revalidate, max-age=0');
	header('Pragma: no-cache');
	header("Location: $url");
	exit;
}
```

Can someone think of any counter-argument? If not, I can send a pull request, of course.

michael-e:
Fixes #2494

### DIFF
--- a/symphony/lib/boot/func.utilities.php
+++ b/symphony/lib/boot/func.utilities.php
@@ -28,6 +28,7 @@ function redirect ($url)
         $url  = str_replace($host, idn_to_ascii($host), $url);
     }
 
+    cleanup_session_cookies();
     header('Status: 302 Found');
     header('Expires: Mon, 12 Dec 1982 06:00:00 GMT');
     header('Last-Modified: ' . gmdate('D, d M Y H:i:s') . ' GMT');


nitriques:
Thanks Michael.

michael-e:
No prob. Again something that I "simply noticed", and wondered why — why it happens, and why I noticed it. :-)

nitriques:
Never stop to do this, please :)

PR #2496, opened 2015-09-04
Added root-dir parameter

nitriques:
Now that Symphony can be installed in subfolder, I need to be able to get that subfolder without the domain attached to it.

So I've splitted up the DOMAIN constant to create the DIRROOT constant that contains only the basename of PHP_SELF.

DOMAIN does not change values, so this is not a breaking change.

### DIFF
--- a/symphony/lib/boot/defines.php
+++ b/symphony/lib/boot/defines.php
@@ -241,11 +241,16 @@ define_safe('__SECURE__',
 );
 
 /**
+ * The root url directory.
+ * @var string
+ */
+define_safe('DIRROOT', rtrim(dirname($_SERVER['PHP_SELF']), '\/'));
+
+/**
  * The current domain name.
  * @var string
  */
-define_safe('DOMAIN', HTTP_HOST . rtrim(dirname($_SERVER['PHP_SELF']), '\/'));
-
+define_safe('DOMAIN', HTTP_HOST . DIRROOT);
 
 /**
  * The base URL of this Symphony install, minus the symphony path.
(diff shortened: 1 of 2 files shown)


nitriques:
I never understood why $root did contain the domain name + folder. I would really like to have it in the core.

I know I can always add it via an extension, but I feel like to would encourage developers to drop $root and use $root-dir.

Using the full url (with protocol + domain name) is not a good practice IMHO. That's why I think it should be in the core.

munki-boy:
+1 won't need to substring sometimes.

nitriques:
> won't need to substring

Yes. Substring really is not a good practice.

munki-boy:
especially substring-before(substring-after) just to get the folder name on its own.

nitriques:
> especially substring-before(substring-after) just to get the folder name on its own.

especially considering the value IS ALREADY separated before constructing root.

@jensscherbl Are we planning 2.6.4 ? I do not really mind releasing more often. Would you still want 2.7 or we go strait to 3.0.0 ?

jensscherbl:
> Are we planning 2.6.4?

Yes.

> I do not really mind releasing more often. Would you still want 2.7 or we go strait to 3.0.0 ?

But the addition you proposed is not a bugfix, so that would be a `2.7.0` release.

nitriques:
> But the addition you proposed is not a bugfix, so that would be a 2.7.0 or 3.0.0 release.

Ok let's do it for 2.7.0 then.

nitriques:
Manually merged in `2.7.x`

PR #2502, opened 2015-10-08
Move date conversion to the PHP level (away from MySQL)

michael-e:
This fix moves date conversion to the PHP level (not MySQL).

Creating timestamps (from datetime strings) in MySQL would only work as intended if MySQL was aware of a named timezone including daylight saving time (DST) logic. Unfortunately, Symphony can only set a fixed time offset - and not a named timezone - for database connections, because it must be friendly to shared hosting environments. So date conversion in MySQL can produce wrong results for system dates, because these have been saved without an explicit (DST/non-DST) offset. The solution is to perform date conversion in PHP exclusively, because Symphony’s `DateTimeObj` will assume the correct time offset.

### DIFF
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -420,8 +420,8 @@ class EntryManager
 
         $sql = sprintf("
             SELECT %s`e`.id, `e`.section_id, e.`author_id`,
-                UNIX_TIMESTAMP(e.`creation_date`) AS `creation_date`,
-                UNIX_TIMESTAMP(e.`modification_date`) AS `modification_date`
+                e.`creation_date` AS `creation_date`,
+                e.`modification_date` AS `modification_date`
             FROM `tbl_entries` AS `e`
             %s
             WHERE 1
@@ -441,6 +441,13 @@ class EntryManager
         );
 
         $rows = Symphony::Database()->fetch($sql);
+
+        // Create UNIX timestamps, as it has always been (Re: #2501)
+        foreach ($rows as &$entry) {
+            $entry['creation_date'] = DateTimeObj::get('U', $entry['creation_date']);
+            $entry['modification_date'] = DateTimeObj::get('U', $entry['modification_date']);
+        }
+        unset($entry);
 
         return ($buildentries && (is_array($rows) && !empty($rows)) ? self::__buildEntries($rows, $section_id, $element_names) : $rows);
     }
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -419,9 +419,9 @@ class EntryManager
         }
 
         $sql = sprintf("
-            SELECT %s`e`.id, `e`.section_id, e.`author_id`,
-                e.`creation_date` AS `creation_date`,
-                e.`modification_date` AS `modification_date`
+            SELECT %s`e`.`id`, `e`.section_id, `e`.`author_id`,
+                `e`.`creation_date` AS `creation_date`,
+                `e`.`modification_date` AS `modification_date`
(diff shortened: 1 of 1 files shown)


nitriques:
Thanks Michael. I'll take care of this after 2.6.4 is out.

brendo:
:+1: for a bump into `2.6.4`. The functionality is identical, it's just the value will now be more accurate.

michael-e:
Should we change the milestone to `2.6.4` then?

nitriques:
There you go. And labels.

michael-e:
I already tested this. Are there more volunteers?

brendo:
Curious what performance impact you noticed from this change?

michael-e:
Oh, sorry, I didn't do any performance tests. Do you think it will haven a significant impact?

nitriques:
I've tested and it seems ok. But maybe this now needs fixing ? https://github.com/symphonists/system_date_fields/blob/master/fields/field.systemcreateddate.php#L62

nitriques:
And I did not saw any perf issue.

michael-e:
> But maybe this now needs fixing ?

I don't think so, but I could test that, on Sunday. But this extension isn't even declared as compatible with 2.6.x (and, at first glance, it should maybe use a better way to get the system date).

nitriques:
> But this extension isn't even declared as compatible with 2.6.x

It does work tho.

>  at first glance, it should maybe use a better way to get the system date

What's the current problem ? I was more worried about the '+00:00' thingny.

michael-e:
I am a bit stressed at the moment, I will take care (and explain what I mean) on Sunday.

nitriques:
> I am a bit stressed at the moment, 

Take you time Michael, no worries.

michael-e:
Re: System Date Fields extension:

1. Out of the box, the fields display the date only (without time) in Symphony 2.6.x. I don't know why this is the case. I had to fool it (by using `'date_format' => 'Y-m-d H:i',`  in the config file in order to do some testing at all.
2. The fields display a correct datetime in the entry index table using the `prepareTableValue` function (which you mentioned), but on an etry's edit page the dates are always the current date, because the function footprint of `displayPublishPanel` is missing the `$entry_id` (and is outdated with other variables, too).
3. Apart from that, the logic you linked to (which occurs twice in each field) will output the correct datetime, but it shouldn't do it like this.
    * `strtotime` shouldn't be used if we have out own methods and functions to deal with datetimes; and it is even unneccessary, at least with the `DateTime` object in Symphony 2.6
    * the explicit offset `+00:00` must be used because GMT datetimes are used. However, Symphony always uses the other column in the database (the localized datetime). This should be unified. But I think that Symphony does it wrong. Ideally, Symphony should only save GMT values, including the offset string (and drop the column with the localized value completely).

I don't know where to go from here. But I am sure that we should open new issues, because problems with the System Date Fields extension or the way Symphony saves system dates are new/different issues. (The current issue is fixed by my PR.)

michael-e:
I [posted some extension issues](https://github.com/symphonists/system_date_fields/issues), so we can discuss these in the extension's issue tracker.

nitriques:
THANKS Michael, I'll review this ASAP!

michael-e:
This PR was supposed to be included in 2.6.4 (see Milestone). Has it been forgotten, or do you want to push it to a later release (2.6.5)?

nitriques:
> Has it been forgotten,

Yes, my bad....

michael-e:
We should create a 2.6.5 milestone, I guess?

nitriques:
> We should create a 2.6.5 milestone, I guess?

Why not! This would give me some practice for creating releases.

nitriques:
Done!

nitriques:
There you go @michael-e I won't forget it this time!

PR #2508, opened 2015-10-12
Removed dead __SYM_DB_INSERT_FAILED__ const

nitriques:
I've made a git blame to try to find the origin of this but I could not
find anything since it was part of the original commit (7f90fa0) !

Nothing in the current base code checks its value.


brendo:
Nice (random) find!

nitriques:
Indeed ;)

PR #2509, opened 2015-10-14
Better error handling and disabled display of exception data by default

nitriques:
Ok this one is a biggie.

I've been wrapping my head around this for quite some time now and with #2504 I though that it was to time to do it.

I won't repeat what I've already described in each commit, it should be easy to follow.

The biggest problem was that the `GenericExceptionHandler` was only disabled after a failed auth. Now, it's activated after a successful auth.

Basically, when the `GenericExceptionHandler` is disabled (i.e `$enabled === false`), we still let it through but display a generic message that does not leak anything. This is not a breaking change since we shortcut'ed the thing with a 404 (which was better then nothing)

I did not changed the default value in the class *per se*, but made it switch to false before we register the functions. I am still not sure about that.

### DIFF
--- a/symphony/lib/core/class.errorhandler.php
+++ b/symphony/lib/core/class.errorhandler.php
@@ -277,6 +277,7 @@ class GenericExceptionHandler
      * Exception in a user friendly way.
      *
      * @since Symphony 2.4
+     * @since Symphony 2.6.4 the method is protected
      * @param string $template
      *  The template name, which should correspond to something in the TEMPLATE
      *  directory, eg `fatalerror.fatal`.
@@ -290,7 +291,7 @@ class GenericExceptionHandler
      * @return string
      *  The HTML of the formatted error message.
      */
-    public static function renderHtml($template, $heading, $message, $file = null, $line = null, $lines = null, $trace = null, $queries = null)
+    protected static function renderHtml($template, $heading, $message, $file = null, $line = null, $lines = null, $trace = null, $queries = null)
     {
         $html = sprintf(
             file_get_contents(self::getTemplate($template)),
--- a/symphony/lib/core/class.errorhandler.php
+++ b/symphony/lib/core/class.errorhandler.php
@@ -13,7 +13,7 @@
 class GenericExceptionHandler
 {
     /**
-     * Whether the `GenericExceptionHandler` should handle exceptions. Defaults to true.
+     * Whether the `GenericExceptionHandler` should display exceptions. Defaults to true.
      * @since Symphony 2.6.4
      *  When disabled, exception are now rendered usign the fatalerror.disabled template,
      *  to prevent leaking debug data.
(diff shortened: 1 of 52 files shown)


brendo:
What happens if there is a fatal error before someone can log in?

nitriques:
> What happens if there is a fatal error before someone can log in.. and you are the developer?

You ssh the box and check the logs. Everything is still there. *Security has a cost.*
We could also do auth sooner, but that would break things I guess (?)

It's worth to mention that the installer still leaks errors tho.

Finally, for the core dev, there's only one line to comment out. I'd rather to that then having full paths disclosed.

I am sure @brendo wants feedback as bad as I do guys (cc @michael-e @jensscherbl in particular)

symphony/lib/core/class.errorhandler.php
```diff
@@ -68,21 +85,19 @@ protected static function __nearbyLines($line, $file, $window = 5)
...
+     * @since Symphony 2.6.4
+     *  The method is final
+     *
      * @param Exception $e
      *  The Exception object
      * @return string
      *  The result of the Exception's render function
      */
-    public static function handler(Exception $e)
+    public final static function handler(Exception $e)
```
brendo: `final` should be first for PSR-2

symphony/lib/core/class.errorhandler.php
```diff
@@ -417,7 +456,7 @@ public static function isEnabled()
      * @return string
      *  Usually a string of HTML that will displayed to a user
      */
-    public static function handler($code, $message, $file = null, $line = null)
+    public final static function handler($code, $message, $file = null, $line = null)
```
brendo: `final` should be first for PSR-2

nitriques: Ok, I'll change that.

Is the same rule valid for abstract ?

```php
abstract public ...
```

nitriques: and static too ?

brendo: [According to the PSR-2 spec](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#45-abstract-final-and-static)

> When present, the abstract and final declarations MUST precede the visibility declaration.
> When present, the static declaration MUST come after the visibility declaration.

nitriques:
```php
final|abstract public static function
```

:) I'll change it. I have to rebase anyway.

nitriques:
@brendo Rebased and PSR-2 things fixed.

brendo:
Not super comfortable about putting this in `2.6.x`. It's a complete flip over existing error handling and I think will need more testing

nitriques:
Let's do 2.7.0 then.

nitriques:
Replaced by #2597

PR #2512, opened 2015-10-20
Default Value jQuery plugin 

### LINKED ISSUE #2511: DS/Event Editor: fill name with the target section name if empty
I feel like this would save me a lot of time. What about you ?

nitriques:
This commit introduce a new jQuery plugin called Default Value, which
gives a input a source element to value its value from, when a specified
event occurs on this source element.

This behaviour is switched to off when the users focuses on the input
field and then blurs out, leaving a value in the input.

This commit also hooks this behaviours on the Data Source name input and
on the Event name input.

Closes #2511

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,15 @@
+# Contributing to Symphony
+
+This is a TL;DR version of the [complete contribution guide](https://github.com/symphonycms/symphony-2/wiki/Contributing-to-Symphony).
+
+The two most important things are [comments](https://github.com/symphonycms/symphony-2/wiki/Contributing-to-Symphony#commenting) and [https://github.com/symphonycms/symphony-2/wiki/Contributing-to-Symphony#code-style](code style).
+
+Please insure that any new method is properly documented and that the code style is respected.
+
+Thanks!
+
+- The Symphony Team
+
+### Legal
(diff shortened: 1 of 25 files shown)


nitriques:
rebased.

nitriques:
@brendo Would this be ok for 2.6.4 (even if it's a new feature). I'd like to use it!

nitriques:
Anybody had the chance to test it ?

jensscherbl:
> Would this be ok for 2.6.4 (even if it's a new feature)

You be the judge... ;)

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.

nitriques:
:) I'll wait then. But 2.7.0 will be released fast, count on me for that.

jensscherbl:
> But 2.7.0 will be released fast, count on me for that.

No problem with that.

nitriques:
> No problem with that.

Great. I do not mind release 2.7.1 early either, if necessary. I,ll package 2.6.4 tomorrow.

nitriques:
Manually merged in 0e27ff5

PR #2520, opened 2015-11-27
Add support for tag negation and filterable operators

jonmifsud:
I've taken a big part of the logic from what is currently in place for the Association Field, but I also made sure to keep the regexp: variants which where there before.

On this implimentation I'm not seeing a significant difference between `not:` and `sql-null-or-not:` for the simple reason that the `NOT EXISTS` filters out both. It might mean to get a 'not' which requires some data (as I assume that's what it's supposed to mean) we'd need to add the JOIN and ensure that the `value IS NOT NULL` - feedback appreciated. Same logic would/should apply to association field in regards to this matter.

### DIFF
--- a/symphony/lib/toolkit/fields/field.taglist.php
+++ b/symphony/lib/toolkit/fields/field.taglist.php
@@ -614,50 +614,133 @@ class FieldTagList extends Field implements ExportableField, ImportableField
         }
     }
 
+    public function fetchFilterableOperators()
+    {
+        return array(
+            array(
+                'title' => 'is',
+                'filter' => ' ',
+                'help' => __('Find values that are an exact match for the given string.')
+            ),
+            array(
+                'filter' => 'sql: NOT NULL',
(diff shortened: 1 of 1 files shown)


brendo:
In what situation does a Taglist field store a `NULL` value?

jonmifsud:
@brendo it never stores a null, it just doesn't store a value. However if you do a left join and there is no data, you get a null in the joined table. That is there are no tags for that particular entry.

nitriques:
Thanks @jonmifsud

nitriques:
@jonmifsud this looks really nice, but is there a way to hide the filter's textbox when the choice is 'not empty' ? @brendo ?

nitriques:
And `is empty or not` does not seems to be working at the moment..

nitriques:
I merged it against the wrong branch, sorry. Merged in `2.7.x` as 786d1cc

PR #2525, opened 2015-12-07
Included two further date formats

animaux:
Appended two further common german date formats for compatibility with german localisation.


nitriques:
Manually merged in 8cbec80

PR #2526, opened 2015-12-08
Open site name backend header link into a new window.

jurajkapsz:
A proposal: User should not lose focus from the backend.

Maybe 100% of the time so far I targeted the header link to a new tab / window. How about you?

If the user wants to leave the backend, (s)he should do it on purpose, so either by logging out or closing the window.

### DIFF
--- a/symphony/lib/toolkit/class.administrationpage.php
+++ b/symphony/lib/toolkit/class.administrationpage.php
@@ -451,7 +451,7 @@ class AdministrationPage extends HTMLPage
         }
 
         $h1 = new XMLElement('h1');
-        $h1->appendChild(Widget::Anchor(Symphony::Configuration()->get('sitename', 'general'), rtrim(URL, '/') . '/'));
+        $h1->appendChild(Widget::Anchor(Symphony::Configuration()->get('sitename', 'general'), rtrim(URL, '/') . '/', null, null, null, array('target' => '_blank')));
         $this->Header->appendChild($h1);
 
         $this->appendUserLinks();


michael-e:
Personally I believe that a user should decide himself if he wants to open a link in a new tab/window, which is perfectly possible as long as you __don't add a link attribute__. But yes, I know, many users don't even understand how to do this in a browser. (Sometimes I hate working on the web...)

jurajkapsz:
I completely agree with all you've written Michael, and yes, a "wheel click" is a troubling knowledge also in my geographic area (how interesting!).

But I think there is another level here - I don't think you can properly await, that a link suddenly takes you away from your scope or UI, which the header link does right now.
It's not clear that it will happen and makes behaviour of links in backend inconsistent. I understand the frontend and the backend to be two separate or standalone environments / roles.
With other words, I also think the first impression people had when they clicked for the first time on the header title was a surprise of leaving the backend. And this is not a good design of user actions.

munki-boy:
Yep, always CTRL+Click for me when I do click that link :+1:

nitriques:
I always add a link in the frontend to go back to the back end. There is also CTRL/CMD+click and right click open in a new tab.

I would prefer to not do this in the core.

@jurajkapsz You could use backend_add_script extension to always have it or code an extension that does this for you. But I am against it in the core.

nitriques:
> don't think you can properly await, that a link suddenly takes you away from your scope or UI

This is the default behavior of all links on the web, so I do not worry about it.

munki-boy:
@nitriques fair enough.

CTRL+Click then lots of CTRL+Tab back and forth is my favourite method.

nitriques:
> CTRL+Click then lots of CTRL+Tab back and forth is my favourite method.

Same here. And you should try shift+click :) And if the open in new window feature is not useful, [please consider replacing it with something else...](https://shft.cl) :) (we made that tool btw)

munki-boy:
The only time I switch from FF to Chrome is when I want to.... Shift-Click :)

nitriques:
:D haha

jurajkapsz:
Thanks for your reviews and insights guys, and the tip on the extension, which is doable.

Also I just want to respond and maybe clarify my why-core idea.

> This is the default behavior of all links on the web ...

I am not sure if it is. There are links that should change current location, and others that should open up in a new window. There is no default behavior in it. As with there are links which have a`src` attribute and links that don't. `src` is not a mandatory or a default attribute. It's just an attribute like `target`.

I don't see that the `target` attribute will leave the specs any time soon, under contrary it was reintroduced in html5. But I am for the better solution anyway.

Anyway, my point was that we might be missing out here on the concept of predictable user interfaces with this header link changing the ui: http://www.w3.org/WAI/intro/people-use-web/principles#predictable

I don't think it's user friendly as it is now in this one manner. But I will digest the idea a bit further.

Thanks.

jurajkapsz:
As for opening links into a new window, I like to wheel-click on them, it's pretty handy and faster than a keyb combination.

nitriques:
> There is no default behavior in it.

I meant, a basic anchor tag will never open in a new window.

>  don't see that the target attribute will leave the specs any time soon

Me neither! But the developer has to opt-in into that, changing the default behavior of the anchor tab.

> Anyway, my point was that we might be missing out here on the concept of predictable user interfaces 

Boy do I like to talk about UI! I could use the same document to state that predictability dictates that a link won't open a new window :) It's predictable because the user wants to leave the backend.

moretaste:
For the convenience of the user, I always have an edit link that goes back to the backend in the front-end templates.

```
<!-- LOGGED IN -->
<xsl:variable name="is-logged-in" select="/data/authors/author"/>

<!-- EDIT LINK -->
<xsl:if test="$is-logged-in">
  <span class="edit">
    <a href="{$root}/symphony/publish/{../section/@handle}/edit/{@id}" target="edit">edit</a>
  </span>
</xsl:if><!-- /EDIT LINK -->
```

nitriques:
@moretaste I have pretty much the [same](https://github.com/DeuxHuitHuit/symphony-2-template/blob/master/workspace/utilities/lib/menu-admin.xsl) [thing](https://github.com/DeuxHuitHuit/symphony-2-template/blob/master/workspace/utilities/lib/edit-buttons.xsl)

Issue #2523, opened 2015-12-03
Potential crash when rotating a log file due to initialisation order


brendo:
I suddenly had a site today go down with a PHP warning about setting the default timezone. It was quite odd because the site was working fine until a random refresh.

I discovered that as the `bundle.php` uses static Symphony methods to set the configuration, start the error handler, database and extension manager, we actually aren't invoking Symphony's constructor until later on. This means the date, language and sessions all are not started until the `symphony()` method is called.

This caused an issue because when the error handler initialised, it wanted to write a log file and give it a date but the configuration hadn't been yet set, and as my local PHP doesn't set the timezone either, the warning was raised.

There is also a potential for edge cases where an error occurs before the error handler is invoked or before the session is actually initialised.

This error was introduced in Symphony 2.5 with the index reorganisation and launcher abilities. Fortunately it's fairly easy to resolve as everything goes through `bundle.php` these days, regardless of the renderer.

Proposal:

- [] add a `Symphony::boot` method which replaces the existing `Symphony::initialise*` calls in the bundle
- [] move the date configuration settings out of the constructor into their own method
- [] remove the duplicate calls to start the error handler
- [] remove the `Frontend::_construct()` method (unused variable)

brendo:
@nitriques Let me know what version you'd like me to target for this and I can make the necessary changes.

nitriques:
@brendo This would need to wait until #2509 is merged. I was hitting similar problems with things being 'not initialised' for some time.

The `Symphony::boot` proposal is great, but would also mean the we need to target this to 2.7.0

brendo:
There is a quick fix that can be implemented for a 2.6.5, simply move the date constant settings from the constructor to the `initialiseConfiguration` method call. I can submit a PR for this if you like?

nitriques:
Yes please !

nitriques:
> Let me know what version you'd like me to target for this and I can make the necessary changes.

@brendo If you'd like to tackle this problem in the `2.7.x` branch, it would be nice!

nitriques:
@brendo Ping? Thanks mate.

brendo:
Mmm, gonna leave this alone. It'll no doubt happen as I pull things apart to get a CI build going.

nitriques:
Great!

PR #2527, opened 2015-12-09
Ensure that the date constants are set as soon as possible.

brendo:
This pull request moves the date constant code to the initialization of the configuration. This ensures that as soon as the configuration is set, the date constants will be available.

It's raises some questions about how Symphony should be booted, but for the meantime, it solves #2523 in a backwards compatible manner.

### DIFF
--- a/symphony/lib/core/class.symphony.php
+++ b/symphony/lib/core/class.symphony.php
@@ -75,12 +75,12 @@ abstract class Symphony implements Singleton
     private static $exception = null;
 
     /**
-     * The Symphony constructor initialises the class variables of Symphony.
-     * It will set the DateTime settings, define new date constants and initialise
-     * the correct Language for the currently logged in Author. If magic quotes
-     * are enabled, Symphony will sanitize the `$_SERVER`, `$_COOKIE`,
-     * `$_GET` and `$_POST` arrays. The constructor loads in
-     * the initial Configuration values from the `CONFIG` file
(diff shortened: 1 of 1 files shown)


symphony/lib/core/class.symphony.php
```diff
@@ -93,14 +93,6 @@ protected function __construct()
             General::cleanArray($_POST);
         }
 
-        // Set date format throughout the system
-        define_safe('__SYM_DATE_FORMAT__', self::Configuration()->get('date_format', 'region'));
-        define_safe('__SYM_TIME_FORMAT__', self::Configuration()->get('time_format', 'region'));
-        define_safe('__SYM_DATETIME_FORMAT__', __SYM_DATE_FORMAT__ . self::Configuration()->get('datetime_separator', 'region') . __SYM_TIME_FORMAT__);
-        DateTimeObj::setSettings(self::Configuration()->get('region'));
-
-        self::initialiseErrorHandler();
```
nitriques: Why remove this line ? is it an accident ? I do not see it added back...

brendo: The error handler is already initialised in the [bundle.php](https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/boot/bundle.php#L27), therefore this is initialising it again :)

nitriques: Perfect, I just wanted to be sure :) Thanks!

Issue #2529, opened 2015-12-12
Installation: The default area for the default author shouldn't be "null"


twiro:
While doing a fresh install of Symphony the first (default) author is created without the option to choose a "default area" that he's forwarded to after login.

So in "sym_authors" the row "default_area" simply says "null" after installation.

In my opinion this should be replaced with "/blueprints/sections/" because apart from the installation process the option to not have a default area isn't considered in the UI.

If the default author opens his author details the field **Default Area** in the fieldset **Login Details** will tell him that "**Sections Index**" is set as default area - which is not true until he actually resaves his author settings.

If he doesn't resave his author settings after installation his default area - as soon as at least one section exists - will become the alphabetically first section, while the selectbox in his author settings will still keep showing "**Sections Index**" because it's not prepared for not having a value and therefore simply shows the first available value.

This behavior is pretty irritating and has bugged me quite a few times in recent projects.

By the way: this problem only exists for the fist/default author because in the symphony backend you can't create other authors without setting a default area.

brendo:
This change potentially means code can be removed from `Administration::__buildPage` which contains the logic for resolving the default section.

It was a best effort and wasn't meant to be annoying :)

twiro:
> _This change potentially means code can be removed from Administration::__buildPage which contains the logic for resolving the default section._

You're speaking about [these lines](https://github.com/symphonycms/symphony-2/blob/2.6.x/symphony/lib/core/class.administration.php#L155-L161), right? 

Should I update my PR by also fixing this part or would you recommend to not change the behavior at all, but rather it's misleading representation in the UI? Which would mean allowing the **Default Area**-Selectbox in the **Login Details** of the Author Section to be empty and not display **Sections Index** as selected when it's actually not.

Though that would be the more complicated solution - and I don't see much speaking against the simpler one. What do you think?

> _It was a best effort and wasn't meant to be annoying :)_

I bet it was ;) And "annoying" sounds way too harsh - I was just a little bit irritated... multiple times ;)

nitriques:
> Should I update my PR by also fixing this part 

Hum... I think we should still retain the default behaviour since ensembles may rely on it.

But yeah, redirecting to the section index would be ok too I guess.

> Which would mean allowing the Default Area-Selectbox in the Login Details of the Author Section to be empty and not display Sections Index as selected when it's actually not.

Yes why not.

twiro:
[...]

PR #2530, opened 2015-12-12
Set "Sections Index" as default area of default (first) author

twiro:
As described here: https://github.com/symphonycms/symphony-2/issues/2529

@nitriques - am I right in open this PR against the 2.6.X-Branch? Or should I use another branch?

### DIFF
--- a/symphony/lib/core/class.administration.php
+++ b/symphony/lib/core/class.administration.php
@@ -149,14 +149,14 @@ class Administration extends Symphony
                 } elseif (!is_null(Symphony::Author()->get('default_area'))) {
                     $default_area = preg_replace('/^' . preg_quote(SYMPHONY_URL, '/') . '/i', '', Symphony::Author()->get('default_area'));
                 }
-                
+
                 // Fallback: No default area found
                 if (is_null($default_area)) {
                     if (Symphony::Author()->isDeveloper()) {
                         // Redirect to the section index if author is a developer
                         redirect(SYMPHONY_URL . '/blueprints/sections/');
                     } else {
-	                    // Redirect to the author page if author is not a developer
+                        // Redirect to the author page if author is not a developer
                         redirect(SYMPHONY_URL . "/system/authors/edit/".Symphony::Author()->get('id')."/");
                     }
                 } else {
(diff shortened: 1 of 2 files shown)


nitriques:
> am I right in open this PR against the 2.6.X-Branch? 

Yes you are! Just check my comments in #2529 :)

symphony/lib/core/class.administration.php
```diff
@@ -149,20 +149,14 @@ private function __buildPage($page)
...
-                            // If there are sections created, redirect to the first one (sortorder)
-                            redirect(SYMPHONY_URL . "/publish/{$section_handle}/");
-                        } else {
-                            // If there are no sections created, default to the Section page
-                            redirect(SYMPHONY_URL . '/blueprints/sections/');
-                        }
+                        // Redirect to the section index if author is a developer
+                        redirect(SYMPHONY_URL . '/blueprints/sections/');
                     } else {
+	                    // Redirect to the author page if author is not a developer
```
nitriques: Could you fix the whitespace issue here please :)

twiro: Fixed :)

nitriques:
Great, this will be merged for 2.7.0 !

nitriques:
Before I can pull this we need this change this line: https://github.com/symphonycms/symphony-2/blob/master/install/includes/install.sql#L13

And also do the change in the upgrade script ?

twiro:
> _Before I can pull this we need this change this line: https://github.com/symphonycms/symphony-2/blob/master/install/includes/install.sql#L13_

> _And also do the change in the upgrade script ?_

Mmh... what would you change it to? We don't have a new default value for **all** types of authors and I don't see a problem keeping `null`.

Apart from setting `/blueprints/sections/`as default area for the **first developer at installation** my PR only changes the redirection after login for authors that actually do have `null` as default area - dependent on whether they are developers or not.

So having `null` in the database still is ok - which also means we wouldn't have to run any database-updates.

nitriques:
Well, ok then!

nitriques:
Sorry I had to delete the merge commit since it's in the wrong branch. It's now 17d87a5

Issue #2532, opened 2015-12-15
Unify the styling of non-editable fields (simple label+value-pairs) in the publish area


twiro:
While [updating](https://github.com/symphonists/system_date_fields/pull/14) the "**System Date Fields**" extension the question came up how a field that only exists out of a label and an uneditable (and short) value should be properly styled.

![screen shot 2015-12-15 at 13 56 58](https://cloud.githubusercontent.com/assets/870227/11811148/f94e4f8c-a333-11e5-9b18-0ea61b95da3d.png)

There seems to be no proper styling solution yet, and while I first proposed to at least align the styling to what extensions like **members** and **member last visit** already do I must admit that I actually don't think they're doing it right in "hijacking" the right-floated `<i>`-element that's used for marking a field as optional. Not only does this tear apart label and value more than reasonable (especially in the main column), but also creates a big gap beneath due to the default minimum height of a field.

![screen shot 2015-12-15 at 11 13 44](https://cloud.githubusercontent.com/assets/870227/11810813/56ada91e-a331-11e5-8702-728d67624a84.png)

I mocked up some possible solutions and do post them here, cause this concerns multiple fields/extensions and therefore the styling should better be provided by the core and then be adopted by all affected extensions.

---

### A) Adopt the styling of reflection field

![screen shot 2015-12-15 at 10 37 24](https://cloud.githubusercontent.com/assets/870227/11810539/1f725da2-a32f-11e5-992a-91e0b4e1595a.png)

This would solve the alignment- and gap-issues described above.
But despite the grey color it looks too editable & input-like in my eyes.

---

### B) Intensify the readonly/uneditable-look

![screen shot 2015-12-15 at 10 40 15](https://cloud.githubusercontent.com/assets/870227/11810558/3f10bf82-a32f-11e5-916c-83ad4b3fceb9.png)

Combining this with a soft grey background works better, but still the input-like look feels a little exagaggerated for this type of data...

---

### C) Return to a less input-like look

![screen shot 2015-12-15 at 10 37 15](https://cloud.githubusercontent.com/assets/870227/11810562/460d3f54-a32f-11e5-963f-64e6cf042ef5.png)

Removing the border while keeping the bigger and greyed out font as well as the default height/alignement-behavior feels like the right solution to me - everything is properly aligned, it's easy so separate between the label and the value of the field and it doesn't imply to be something editable.

---

[...]

PR #2534, opened 2015-12-17
New class "field-value-readonly"

twiro:
Adds a new class that allows for consistent styling of field values that are displayed as plain text and are not editable _(As seen in extensions like „System Date Fields“, „Members“, „Member Last Visit“, and maybe more)_.

See https://github.com/symphonycms/symphony-2/issues/2532 for details.

I’m planning to send PRs for the affected extensions after this has been published in the core.


nitriques:
Thanks!

PR #2538, opened 2016-01-01
Fix SMTP issues with comma-separated recipients

michael-e:
This fix has been proposed by @nitriques here: https://github.com/creativedutchmen/email_template_manager/issues/85


nitriques:
Thanks. I already had the exact same thing on my machine, forgot to push it..

PR #2544, opened 2016-01-22
Gateway: Curl error

nitriques:
This caught my attention: curl_error actually contained the curl_errno

I know this is a breaking fix.... so I'll leave the PR open until it's ready to merge into 2.7.0


nitriques:
Duplicated. See #2551 for  details.

PR #2551, opened 2016-01-29
Add curl_error message

nitriques:
This caught my attention: curl_error actually contained the curl_errno

Since it confusing, this aligns things with how curl names things


brendo:
Just flagging a dependency on [Remote Datasource](https://github.com/symphonycms/remote_datasource/blob/master/data-sources/datasource.remote.php#L139) for this change :)

brendo:
And you also have #2544 as well? Duplicate?

nitriques:
> Just flagging a dependency on Remote Datasource for this change :)

Oh ok thanks!

> And you also have #2544 as well? Duplicate?

Ah damn yes.

nitriques:
This is postpone to 3.0.0 has it is a breaking change, sadly.

nitriques:
Remote Datasource has been taken care of

PR #2382, opened 2015-03-20
Association filters

jonmifsud:
Opening PR to continue the discussion on #2322 as is the PR is missing an update to add the `filters` field to the associations table.

In conjunction with this there are PR's on the respective association field components, which make use of the filtering. [Association UI Selector](https://github.com/hananils/association_ui_selector/pull/44) and [Association UI Editor](https://github.com/hananils/association_ui_editor/pull/26)

I do think there might be a bit more work required and adjustments but would be good to have someone try it out for feedback at this stage.

PR #2565, opened 2016-02-24
Association filters: Rebased version of #2382

nitriques:
As the title says!

@jonmifsud This would need testing if you do not mind!


nitriques:
This still needs a proper update script.

PR #2568, opened 2016-03-01
Only request the needed schema in table view

nitriques:
This changes uses the array returned by $section->fetchVisibleColumns() in order to create the element names array to pass to the entry manager. This yields a nice performance update when a section contains many (hundreds) of fields but only a few of them are needed for the publish table view.

I do not see how this can't be backward compatible, but I created a PR just to document the change and ask for testers! :)

Thanks!

### DIFF
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -472,10 +472,17 @@ class contentPublish extends AdministrationPage
          */
         Symphony::ExtensionManager()->notifyMembers('AdjustPublishFiltering', '/publish/', array('section-id' => $section_id, 'where' => &$where, 'joins' => &$joins));
 
+        // get visible columns
+        $visible_columns = $section->fetchVisibleColumns();
+        // extract the needed schema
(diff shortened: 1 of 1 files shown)


michael-e:
Honestly, I don't see any performance difference. Should I increase the number of entries per page, maybe?

nitriques:
> Should I increase the number of entries per page, maybe?

Yes. But you get a real gain when you have a field with hundreds of fields (which, I agree, can be a pain to create).

But most importantly: there's no bug?

michael-e:
No, it didn't break, which is good. :-)  

> hundreds of fields

My test section has 75 fields — no visible difference, even with 100 entries per page. Maybe my server is too fast. :-))

nitriques:
> Maybe my server is too fast. :-))

Well, maybe :P The problem was not the webserver tho, it was the DB... I saved 500ms of a request that was taking 2500ms (which is 20% decrease)

nitriques:
After testing by myself and @michael-e it's safe to merge in.

PR #2570, opened 2016-03-09
Use Composer

brendo:
This pull request removes the `/vendor` directory from the codebase in favour of introducing composer into the development workflow.

There are additional fixes for some broken bits and pieces on the `integration` branch.


nitriques:
Those will be fun to rebase :+1: Thanks!

Issue #2223, opened 2014-10-12
Allow Log integration with Monolog


brendo:
Our Symphony Log is limited to file rotation logging and doesn't follow any particular standard. Integration with the ever popular and extensible [Monolog](https://github.com/Seldaek/monolog) will allow greater freedom and flexibility to log information and errors.

By default we could use the [StreamHandler](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/StreamHandler.php) or [RotatingFileHandler](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/RotatingFileHandler.php) to mirror what we get at the moment.

brendo:
So this is something I've been wanting to do for a long time. The core Logger has been ripped out and replaced by [Monolog](https://github.com/Seldaek/monolog). Monolog is a PSR-3 compatible logging system and out of the box offers integration with many third party logging systems.

One of the more interesting things about this change, is that the Handler/Formatter used is totally derived from the [config file](https://github.com/symphonycms/symphony-2/blob/a3c495780bd96a8da96bf90ab23422f74477ad1c/install/includes/config_default.php#L28-L50). This Log configuration also supports placeholders within the context of the entire configuration file, or within `Symphony:: initialiseLog`. It's a little rough around the edges, but I wanted to prove the concept can work and thought a first effort would allow more detailed discussion to take place.

This change is likely to not be backwards compatible as the API of the Log component has changed.

@nitriques @pointybeard

nitriques:
@brendo this is nice. But:

1. Did we lose features (I.e log rotation?)
2. Can't we create a compatibility layer to make it backward compat ? Or it is impossible ?
3. Could it be possible to split the commit ? github has a hard time showing the diff...

Thanks.

brendo:
> Did we lose features (I.e log rotation?)

In it's exact replica, yes, we no longer have log rotation after 'x' size. We do have access to Monolog's `RotatingFileHandler` however:

> `RotatingFileHandler: Logs records to a file and creates one logfile per day. It will also delete files older than $maxFiles. You should use logrotate for high profile setups though, this is just meant as a quick and dirty solution.`

> Can't we create a compatibility layer to make it backward compat ? Or it is impossible ?

[...]

PR #2571, opened 2016-03-09
Replace custom logging with Monolog

brendo:
This implements #2223 for the `integration` branch. It's a great step forward, although there's still more to do to allow easier configuration.. stay tuned.


brendo:
Probably best to merge #2570 first, as I rebased off that work ;)

nitriques:
Awesome!

brendo:
Perfecto, next stop config files

nitriques:
This also opens the path to installing extensions with composer (I mean downloading it)

brendo:
> This also opens the path to installing extensions with composer (I mean downloading it)

Yep, the path I'm currently heading down is making composer be able to install Symphony, so we run Symphony in a CI environment...

nitriques:
That would be awesome.

I often dream about splitting the core across multiple module :+1:

Issue #2499, opened 2015-10-01
Publish Filter : "between mode" (x to y) not working


twiro:
I was just checking if number field is fully compatible with the current 2.6.3 release and noted that publish filtering doesn't throw any results when using the "between mode".

Looks like the filtering-url is constructed the wrong way.
If i try to filter inbetween a range of "1" and "3" the resulting URL looks like this: 

```
.../publish/entries/?filter[number]=x to y 1 to 3
```

If I manually edit the URL to...

```
.../publish/entries/?filter[number]=1 to 3
```

...everything works as expected.

nitriques:
This looks more like a number field issue since declaring which filter is supported by each field can be customized (see https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/toolkit/class.field.php#L1178-L1208)

Or I do no understand what's the problem...

twiro:
I tried to fix this, but stumbled upon a bigger conflict along the way:

---

### A) Trying to fix the field

Number-field defines the "between"-filter [in these lines](https://github.com/symphonycms/numberfield/blob/master/fields/field.number.php#L230-L237). As described above this doesn't work because the `filter`-value `x to y` gets included in the URL when using the between-comparison.

As filtering works when removing "**x to y**" from the URL, I simply tried to change [line 232](https://github.com/symphonycms/numberfield/blob/master/fields/field.number.php#L230-L237) and set the `filter`-value to `' '`.

**Result:** Filtering works as expected, but as the filter-value is used as actual `value`-attribute of the comparison mode selectbox-options I now had two options with the same value `' '`. That obiously doesn't work - when choosing the **between**-mode publish filtering gets confused and displays the help-text of the **is**-mode.

So as each filtering option has to come with a unique `filter`-value by design I don't see a way to solve this with the filter-settings of the field.

---

### B) Trying to fix the core

My other attempt to solve this was to prepare the core to properly get along with the filtering settings of number-field.

I replaced [line 138 of **symphony.filtering.js**](https://github.com/symphonycms/symphony-2/blob/master/symphony/assets/js/src/symphony.filtering.js#L138) with the following code:

	// Comparison mode "between" (x to y)
	if (comparison.replace(/\s+/g, '') == 'xtoy') {
		value = 'filter[' + query.attr('name') + ']=' + $.trim(query.val());
	
	// Other comparison modes
	} else {
		value = 'filter[' + query.attr('name') + ']=' + comparison + $.trim(query.val());
	}

[...]

Issue #2564, opened 2016-02-19
Datasource & Publish Filtering / Tests and Issues


twiro:
While fixing a bug regarding the **between**-filter-mode of **number field** in symphony's publish filtering interface I stumbled upon some other issues and started to do a lot of testing and comparing to assure none of them were caused by my changes to the publish filtering interface.

While figuring out the best way to document my test results I stumbled upon [this page in the wiki](https://github.com/symphonycms/symphony-2/wiki/DateField) and decided to set up something similar for each field/extension I tested.

To add additional value I not only tried to make the test cases cover most common use cases (I could think of) but also started to run each test not only for **Publish Filtering** but also for **Datasource Filtering**:

**[The results](https://github.com/twiro/symphony-filtering-tests)** verified that my changes didn't cause any of the issues, but revealed that the filtering system (especially publish filtering) has quite some shortcomings.

So my questions to @nitriques and the community are the following:

1. Would it help to post a new issue for each single bug I found or should I better let it be? I don't expect most of the bugs to get fixed in the near future and I don't want to flood the issue tracker with publish filtering bugs noone really cares about (not even me).
2. Would my [Symphony Filtering Tests](https://github.com/twiro/symphony-filtering-tests)-Repo be a useful addition to symphony's documentation ecosystem? If so I would gladly pass it over to the "symphonycms"-account so that it might get extended, maintained and updated by the community and whoever might stumble upon problems regarding filtering might use this as a reference, starting point and helpful resource for his own tests and extend/update the repo with his results.

nitriques:
> Would it help to post a new issue for each single bug I found or should I better let it be? 

Yes, I do think so!

> Would my Symphony Filtering Tests-Repo be a useful addition to symphony's documentation ecosystem?

Yes it would if there was a way to run the tests automatically... (I'll be working on that...)

PR #2577, opened 2016-03-18
Add data-attributes to publish filtering interface

twiro:
This PR introduces data-attributes to the publish filtering interface. This allows for better and more precise and flexible handling of comparison modes and their help texts.

It also removes the (annoying) leading whitespace in the input fields of publish filtering interface.

See https://github.com/symphonycms/symphony-2/issues/2499 for details.

### DIFF
--- a/symphony/assets/js/src/symphony.filtering.js
+++ b/symphony/assets/js/src/symphony.filtering.js
@@ -88,12 +88,13 @@
 			var item = $(this),
 				comparison;
 
-			// Show help contextually
+			// Case: Adding a new filter instance
 			if(item.is('.instance')) {
-				comparison = item.find('.comparison').val();
+				comparison = item.find('.comparison option:selected').attr('data-comparison');
 			}
+			// Case: selecting a new comparison mode of an existing filter instance
 			else {
-				comparison = item.val();
+				comparison = item.find('option:selected').attr('data-comparison');
 				item = item.parents('.instance');
 			}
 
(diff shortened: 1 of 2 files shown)


nitriques:
At first glance, it seems ligit. I'll test ASAP. Thanks!

nitriques:
Did you tested it with some of the most used extensions ?

twiro:
> _Did you tested it with some of the most used extensions ?_

You bet I did :)

As I noticed lots of little quirks and bugs in Publish Filtering (mostly related to special charactes and umlauts) while testing my changes I had to reassure that they weren't caused by anything I did - so I started testing really intensely and documented my results for both cases - before my changes and after - they both were exactly the same except for the one scenario I wanted to fix (the "between"-filter in the Publish Filtering Interface).

I posted about this [here](https://github.com/symphonycms/symphony-2/issues/2564) and documented my filtering results in an [own repo](https://github.com/twiro/symphony-filtering-tests).

My changes won't do much to fix most of the issues Publish Filtering has, but nonetheless they're fixing a basic conceptual flaw and at least will make the [last four failed test cases in number field](https://github.com/twiro/symphony-filtering-tests/blob/master/fields/number-field.md) disappear! 

(Though the number field extensions needs to be fixed for that too - will do so after this change has been released!)

nitriques:
> You bet I did :)

:+1: 

> I posted about this here and documented my filtering results in an own repo.

This is really nice thanks.

I'll test ASAP.

PR #2585, opened 2016-04-16
Lock/unlock tables when editing entries

### LINKED ISSUE #2472: When editing entries, Symphony will have time slots without data
When editing existing entries (see `EntryManager::edit`), Symphony will iterate over all fields, delete existing data first, then insert a new row for the field's data. So there will always be a small time slot without data.

This caused massive headaches for me the other day, because I experienced simultaneous processes which edited entry data, but every one of them relied on data being available. My solution was to stop using `EntryManager::edit`. Instead I added an additional unique key to the table and updated data using `Symphony::Database()->insert($data, 'tbl_entries_data_' . $field_id, true);` (where `true` will make Symphony add `ON DUPLICATE KEY UPDATE` to the query).

Still, after having applied this hackish fix, I asked myself: Why doesn't Symphony lock the table?

Like so, basically:
- `LOCK TABLES tbl_entries_data_{$field_id} WRITE`
- delete data
- insert data
- `UNLOCK TABLES`

Is there any reason against this?

michael-e:
When editing an entry, Symphony will iterate over all fields, delete existing data first, then insert a new row for the field's data. So there will always be a time slot without field data. If multiple processes edit the same entry, this means a race condition that may lead to data being corrupted in several ways (also depending on the field type). Locking the table solves the issue.

Fixes #2472

### DIFF
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -174,39 +174,48 @@ class EntryManager
                 continue;
             }
 
+            $did_lock = false;
             try {
+                $did_lock = Symphony::Database()->query('LOCK TABLES tbl_entries_data_' . $field_id . ' WRITE');
                 Symphony::Database()->delete('tbl_entries_data_' . $field_id, sprintf("
                     `entry_id` = %d", $entry->get('id')
                 ));
-            } catch (Exception $e) {
-                // Discard?
-            }
 
-            if (!is_array($field) || empty($field)) {
-                continue;
-            }
-
-            $data = array(
-                'entry_id' => $entry->get('id')
-            );
-
-            $fields = array();
-
-            foreach ($field as $key => $value) {
-                if (is_array($value)) {
-                    foreach ($value as $ii => $v) {
-                        $fields[$ii][$key] = $v;
+                if (!is_array($field) || empty($field)) {
+                    if ($did_lock) {
+                        Symphony::Database()->query('UNLOCK TABLES');
                     }
-                } else {
-                    $fields[max(0, count($fields) - 1)][$key] = $value;
+                    continue;
                 }
+
+                $data = array(
+                    'entry_id' => $entry->get('id')
+                );
+
+                $fields = array();
+
+                foreach ($field as $key => $value) {
+                    if (is_array($value)) {
+                        foreach ($value as $ii => $v) {
+                            $fields[$ii][$key] = $v;
+                        }
+                    } else {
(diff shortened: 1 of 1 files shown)


symphony/lib/toolkit/class.entrymanager.php
```diff
@@ -174,39 +174,48 @@ public static function edit(Entry $entry)
...
+                    $fields[$index] = array_merge($data, $field_data);
+                }
+
+                Symphony::Database()->insert($fields, 'tbl_entries_data_' . $field_id);
+            } catch (Exception $e) {
+                Symphony::Log()->pushExceptionToLog($e, true);
             }
 
-            Symphony::Database()->insert($fields, 'tbl_entries_data_' . $field_id);
+            if ($did_lock) {
```
nitriques: This block *should* be in a finally statement to be 100% kasher (if I am not wrong?)

symphony/lib/toolkit/class.entrymanager.php
```diff
@@ -174,39 +174,48 @@ public static function edit(Entry $entry)
                 continue;
             }
 
+            $did_lock = false;
             try {
+                $did_lock = Symphony::Database()->query('LOCK TABLES tbl_entries_data_' . $field_id . ' WRITE');
                 Symphony::Database()->delete('tbl_entries_data_' . $field_id, sprintf("
```
nitriques: The `WRITE` part means the table is still available for reads ?

nitriques:
Thanks Michael. See my comments!

symphony/lib/toolkit/class.entrymanager.php
michael-e: We discussed that some months ago (https://github.com/symphonycms/symphony-2/issues/2472#issuecomment-134315796); the issue is that `finally` requires PHP 5.5+.

symphony/lib/toolkit/class.entrymanager.php
michael-e: No, it means that no other session can access the table, which is exactly what we need. See http://dev.mysql.com/doc/refman/5.7/en/lock-tables.html

nitriques: From what I am reading: 

> No other session can access it until the lock is released.

This is then an exclusive lock. Other operation are "queued". And yes, this is what we need ;)

symphony/lib/toolkit/class.entrymanager.php
nitriques: I always forget about that. Thanks!

nitriques:
I merge ASAP then!

PR #2588, opened 2016-04-25
Update install.sql for MySQL 5.7

Petertron:
In MySQL 5.7 a datetime field cannot have a default value of "0000-00-00" for the date, the minimum value is "1000-01-01".


nitriques:
Thanks!

PR #2589, opened 2016-04-25
Update class.cryptography.php for PHP 7 strictness.

Petertron:
mt_srand function argument must be an integer.


nitriques:
What was it getting ? a float ?

Petertron:
`microtime(true)` indeed gives a floating point result.

nitriques:
Thanks!

PR #2597, opened 2016-05-31
Better error handling and disabled display of exception data by default for 3.0.0

nitriques:
#2509 was submitted against 3.0.x. Here's a rebased version.

### DIFF
--- a/symphony/lib/core/class.errorhandler.php
+++ b/symphony/lib/core/class.errorhandler.php
@@ -154,6 +154,14 @@ class GenericExceptionHandler
         }
 
         // Pending nothing disasterous, we should have `$e` and `$output` values here.
+        self::sendHeaders($e);
+
+        echo $output;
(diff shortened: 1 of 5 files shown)


symphony/lib/core/class.errorhandler.php
```diff
@@ -87,20 +104,19 @@ public static function isValidThrowable($e)
...
+     * @since Symphony 3.0.0
+     *  The method is final
+     *
      * @param Throwable $e
      *  The Throwable object
      * @return string
      *  The result of the Throwable's render function
      */
-    public static function handler($e)
+    final public static function handler($e)
```
brendo: Curious on why `final`? What problem does this avoid?

nitriques: I am paranoid and do not want to override it by accident.

PR #2601, opened 2016-06-08
Adjust installation wording on deleting install folder

DavidOliver:
The previous wording could be read as inferring that the deletion will automatically take place when proceeding.

### DIFF
--- a/install/lib/class.installerpage.php
+++ b/install/lib/class.installerpage.php
@@ -208,7 +208,7 @@
 
             $this->Form->appendChild(
                 new XMLElement('p',
-                    __('I think you and I will achieve great things together. Just one last thing: how about we remove the %s folder and secure the safety of our relationship?', array('<code>' . basename(INSTALL) . '</code>'))
+                    __('I think you and I will achieve great things together. Just one last thing: please remove the %s folder to secure the safety of our relationship.', array('<code>' . basename(INSTALL) . '</code>'))
                 )
             );
 
@@ -220,7 +220,7 @@
 
         protected function viewConfiguration()
         {
-            /* -----------------------------------------------
+        /* -----------------------------------------------
          * Populating fields array
          * -----------------------------------------------
          */


michael-e:
AFAIK, Symphony actually **attempts** to delete the install folder. It may simply not work because of owner/permission settings.

DavidOliver:
I believe Symphony had write access.

DavidOliver:
Even if that's the case, I suppose it's not a bad thing to directly instruct/request manual deletion in case Symphony can't.

michael-e:
> I believe Symphony had write access.

In my experience, if owner/permissions allow to remove the folder, it will be removed. There should be something in yout log file!

If we ask for manual deletion, we should remove auto-deletion. If we keep auto-deletion, we must find a wording to tell the user that it will only be **attempted**.

DavidOliver:
There's nothing in the install log and Symphony definitely had write access as I've just used Symphony's notification prompt to delete the install directly.

By the way, the install log states version 2.6.7, but I expect that's something that will be updated as part of procedure before the release.

michael-e:
OK, so there is something broken. It either must be fixed or the wording has to be changed.

DavidOliver:
Do you know where the code would be for the deletion of the install directory if Symphony does attempt the deletion? I've had a quick look but haven't found anything yet.

michael-e:
No sorry, I don't know. And unfortunately I don't find the time to look into it.

DavidOliver:
I've had another look and it seems to me that [the installer would delete the directory](https://github.com/symphonycms/symphony-2/blob/2.7.x/install/index.php#L44) if there was a URL query parameter named `action` with a value of `remove`, but that the installer doesn't get requested with that parameter as part of the process. [The `viewSuccess` method/page](https://github.com/symphonycms/symphony-2/blob/2.7.x/install/lib/class.installerpage.php#L179) just seems to take the user straight to the admin area.

[`/install/?action=remove` is how Symphony deletes the install directory](https://github.com/symphonycms/symphony-2/blob/2.7.x/symphony/lib/core/class.administration.php#L235) when the user clicks the prompt at the top a normal admin page, which is not great as anyone could trigger it. Probably not very likely to be a problem often, but I would say it should only be doable when logged in as a Symphony developer.

nitriques:
Hey guys, I've never had a install folder delete itself. NEVER. They always had the right to do it tho.

Deleting the install folder via a GET request is not a good idea, indeed, but it's not a real problem, since the install directory should never go live, and anybody requesting it does the site owner a favor.

That being said, I'll take a closer look at the code base and get back to you.

nitriques:
The only place where we delete the install folder is when requesting the install page with `?action=remove`. The install script NEVER tries to delete its folder.

So the wording should never make the user believe that's it's done.

I like @DavidOliver proposal. @michael-e would you change it a bit ?

nitriques:
And what about adding a link in that p element to let the user do it right away ?

nitriques:
@DavidOliver ?

DavidOliver:
> And what about adding a link in that p element to let the user do it right away ?

Fine with me, though would it require additional logic for checking for the existence of the install directory?

nitriques:
> would it require additional logic for checking for the existence of the install directory?

No. The request you see comes from that folder. As I said, symphony never tries to delete the install directory without user navigating to ?action=remove

nitriques:
@nilshoerrmann if you ever feel like it, I'd like to know your thoughts on this issue. Thanks!

hotdoy:
image
image

nitriques:
Closing since it's it https://github.com/DeuxHuitHuit/symphony-2/commit/52e12b911b0d6807ed372b5f6a685e221377d1e4

PR #2602, opened 2016-06-16
Tag list field: accept an array of tags

nilshoerrmann:
While the tag list field always returned string (parameter pool) and array values (XML) in the output, it did only accept a string of tags on post. This commit extends the field to accept an array of tags as well which is helpful, if you are building a tag list widget on the front-end and would like to post back to the system.

This change is supposed to be backwards compatible (it doesn't change the behaviour for posted string values).


nitriques:
Thanks @nilshoerrmann ! Wanna test/review 0e299c6 too ? (it's in `2.7.x` already!)

nilshoerrmann:
Thanks for merging!

Issue #2553, opened 2016-02-02
EntryManager::fetchByPage has no param to set sorting


michael-e:
Sorting of paginated entries is a "global state" for the EntryManager.

When using `EntryManager::fetchByPage`, it's not possible to define a sorting directly in the function call. The sorting of entries will depend on two protected static variables of the EntryManager, `$_fetchSortField` and `$_fetchSortDirection`.

These variables can be set with public methods, e.g.`EntryManager::setFetchSorting`.

You may get surprising results if, for example, you are using delegates and callbacks while looping over entries (fetched using `fetchByPage`). If any sub-process calls `EntryManager::setFetchSorting`, this will influence the sorting/ordering of `EntryManager::fetchByPage`.

There are only two solutions to this: Either you must set the desired sorting manually (e.g. at the end of every loop in your script), or — much more elegant — we add a parameter to the `fetchByPage` method which allows to set the sorting with each call.

Note: `fetch` has a param called `$enable_sort` — you can switch off sorting here (which is no good idea, because the resulting sorting is undetermined).  `fetchByPage` has no such param.

nitriques:
I am glad to report that this will be fixed in 3.0.0 :)

nitriques:
This is now fixed, because fetchByPage and setFetchSorting are deprecated !

PR #2608, opened 2016-07-24
Sorting entries in associations drawer

jensscherbl:
Would make a lot more sense to me if the associations drawer would show entries of associated sections in the same order as the publish index table.

My quick and dirty hack seems to work fine so far, but not sure if this has any nasty side effects or if there's a more elegant solution to the problem (maybe #2553).

Also not sure if others agree with me on this matter in general, so feel free to dismiss this pull request if this is actually not a desired behavior.

On a side note, I guess it would be possible to provide this functionality as an extension, but the extension would have to duplicate most of the core code for the associations drawer. Might still be an option for bringing this to the `2.6.x` branch as well, though.

### DIFF
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -244,12 +244,12 @@ class contentPublish extends AdministrationPage
 
         // Custom field comparisons
         foreach ($data['operators'] as $operator) {
-            
+
             $filter = trim($operator['filter']);
-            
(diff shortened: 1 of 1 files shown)


michael-e:
I agree.

nitriques:
I agree with it, will test ASAP. Thanks Jens!

PR #2610, opened 2016-07-26
Cleaning up the data source output

### LINKED ISSUE #2607: Remove associated count with the section handle only
Keep only the one with the section handle + field handle.

Same should be done for ds output parameters

ChoboHub:
Fixes #2607


nitriques:
Merged as 637a84559f and 9ec3d384fe

PR #2612, opened 2016-07-27
Fix deprecate in content.blueprintsdatasources

ChoboHub:
Datasource::grab() is deprecated since Symphony 2.3.1
Using `execute()` instead

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -168,7 +168,7 @@ class contentBlueprintsDatasources extends ResourcesPage
 
                             // Handle pre Symphony 2.2.1 Static DS's
                         } else {
-                            $fields['static_xml'] = trim($existing->grab());
+                            $fields['static_xml'] = trim($existing->execute());
                         }
                         break;
                     default:


symphony/content/content.blueprintsdatasources.php
```diff
@@ -168,7 +168,7 @@ public function __form()
 
                             // Handle pre Symphony 2.2.1 Static DS's
                         } else {
-                            $fields['static_xml'] = trim($existing->grab());
+                            $fields['static_xml'] = trim($existing->execute());
```
brendo: This code block runs on DS's generated in Symphony 2.2.1, where `execute` was not deprecated. I don't think we should make this change.

nitriques: Oh you're right. BTW @brendo I am working with @ChoboHub . Those issues where raised by scrutinizer, which does not always get it right. Sorry for have wasted your time on this :(

brendo: Don't be sorry! I had the same thing when I started fixing Scrutinizer and PHPStorm problems, sometimes it doesn't get it quite right :)

nitriques: ;) Thanks mate. Using Scrutinizer bugs as a start for @ChoboHub to learn system has been great so far tho.

ChoboHub:
yes ^^ sorry @brendo for the time. Thanks a lot for all your review and comments!

PR #2613, opened 2016-07-27
Unused code

ChoboHub:
In content.blueprintspages :  '$autogenerated_handle' was always set to false when he was call.
In class.frontendpage : '$key' wasn't used, so a simple refactoring.


symphony/content/content.blueprintspages.php
```diff
@@ -649,7 +648,6 @@ public function __actionEdit()
 
             if (trim($fields['handle']) == '') {
                 $fields['handle'] = $fields['title'];
-                $autogenerated_handle = true;
```
brendo: But this scenario is when `$autogenerated_handle` is set to `true`, not `false` so although rare, it _is_ possible?

nitriques: Look at the complete code path. Even if true, it will revert to false on line 680. 2 pages with the same name should not cause any problems, as long as their respective handles are different. I had to do it once and had to hack the DB.

nitriques:
Merged as 2c6a37d

PR #2614, opened 2016-07-27
Follow up on #2601

nitriques:
See #2601 for details

cc @DavidOliver


brendo:
Try not to put the HTML into the translation string as it makes it harder for the translators, /cc @nilshoerrmann

nitriques:
Good catch, sorry that's my fault. I'll leave the PR as reminder, but DO NOT MERGE.

nitriques:
Merged as 4c7e9e3...52e12b9, with the recommendations of @brendo

Issue #796, opened 2011-10-04
Implement New Password Hashing


michael-e:
_There has been a discussion about Symphony's security at the Symposium 2011. This is an abstract of one of the topics._

Currently passwords for Members are stored using a SHA1 hash with a fixed salt. Author/developer passwords are stored using SHA1 without a salt.

In the future all password storage should use the same hashing function. It should use a random salt, similar to the 'SSHA' password scheme (e.g. in OpenLDAP). (We will use an even longer salt, however.)

(An explanation of password schemes can be found here: http://wiki.dovecot.org/Authentication/PasswordSchemes)

Hashes might basically look like this (50 character string):

```
{SYMSSHA1}P4jeTuq8sLBuXyv0BH/jRcFMquoyRlltZVd6emJa
```
### Updating password hashes

When Symphony switched from MD5 to SHA1, it was necessary to reset all the author passwords. This should not be necessary now (or in the future) if we **use a prefix** like in the example above. It's quite easy then to see how a specific password has been hashed and execute the corresponding action (e.g. choose the validation function that fits or initiate the hash update method.)

The logic to update the hash would be quite simple: If the old hash matches the hash which is generated using the password, the new hash will be generated to overwrite the old hash.

It must still be discussed if updating the hashes should be done automatically upon user login. Especially @phoque (a.k.a. @nils-werner) felt bad about this.
### Suggested hash

It has been discussed to generate the salt from the password itself (so there is no need to store it), but as @phoque pointed out, the salt would not be random then. In that case a rainbow table which fits one Symphony installation (i.e. password hashes) would fit any installation in the world. So we decided to go for a random salt.

We don't use the SSHA mechanism "as is", but choose a longer random salt. Although that may not increase the security as much as one might expect, it's definitely not bad to do this. As @creativedutchmen pointed out, this will be useful if SSHA rainbow tables occur one day since these wouldn't fit to Symphony's method.
- suggested prefix: {SYMSSHA1} (which means 'Symphony's Salted SHA1')
- suggested hashing algorithm: SHA1
- suggested salt: random, 10 characters
- suggested encoding of the hash: base64

(In OpenLDAP the default encoding of the hash and the salt is base64 for SHA-type password schemes. Authentication software like e.g. Dovecot will automatically detect base64 vs. HEX encoding.)
### Hashing functions

Currently in `class-general.php` we have:

```
/**
 * Uses SHA1 or MD5 to create a hash based on some input
 * This function is currently very basic, but would allow
[...]

PR #2615, opened 2016-07-27
Unused code #796

ChoboHub:
In class.pagemanager : `$success` will always be = PageManager::editPageChildren(...);
but we want to return false. Small fix to `return false;`

in field.taglist : the `$condition` wasn't use at all in the `else`.

in field.upload : `$renamed_file = $new_file;` was simply not use.
Did couple tests without it and upload working as espected.

### DIFF
--- a/symphony/lib/toolkit/class.pagemanager.php
+++ b/symphony/lib/toolkit/class.pagemanager.php
@@ -300,14 +300,6 @@ class PageManager
                 'path' => $page_path
             );
 
-            if (!PageManager::createPageFiles($page_path, $child['handle'], $child['path'], $child['handle'])) {
-                $success = false;
-            }
-
-            if (!PageManager::edit($child_id, $fields)) {
-                $success = false;
-            }
-
             $success = PageManager::editPageChildren($child_id, $page_path . '/' . $child['handle']);
         }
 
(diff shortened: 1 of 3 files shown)


symphony/lib/toolkit/class.pagemanager.php
```diff
@@ -301,11 +301,11 @@ public static function editPageChildren($page_id = null, $page_path = null)
             );
 
             if (!PageManager::createPageFiles($page_path, $child['handle'], $child['path'], $child['handle'])) {
-                $success = false;
+                return false;
```
brendo: Note this changes the current behaviour. As this is in a loop, previously some children may of been updated, whereas now, as soon as one fails, the execution will stop (even if some previously worked).

nitriques: It does change the behaviour. But I could not see how continuing to save things, while being technically in a error state, is a good thing. I thought about only issuing the calls and removing the if completely but that was not better.

Thinking about it, should we have 3 new boolean and return

```
return $successCreatePageFiles && $successEdit && $sucess
```

and perhaps rename $success ? That would not change the behaviour *per se*: only the return value... ?

nitriques:
Merged as 4c7e9e3, without the changes to PageManager, as @brendo said it changes the behaviour.

PR #2617, opened 2016-08-03
Add the Content-ID header on attachments

nitriques:
This allows developer to reference attachments in the html body to allow
images to be viewed offline.

You can use it as follow:

``` html
<img src="cid:name-of-the-file.ext" />
```

Where the name-of-the-file.ext part matches the actual file name of the
file. Notice the `cid:` prefix, which makes it work!

@michael-e I can wait for you to review this. If @brendo or even @creativedutchmen could do it too I would appreciate, since your _mail-fu_ is greater than mine.

I've tested with and without references and everything works.

### DIFF
--- a/symphony/lib/toolkit/class.emailgateway.php
+++ b/symphony/lib/toolkit/class.emailgateway.php
@@ -305,6 +305,10 @@ abstract class EmailGateway
     /**
      * Appends one file attachment to the attachments array.
      *
+     * @since Symphony 3.0.0
+     *   The file array can contain a 'cid' key.
+     *   When set to true, the Content-ID header field is added.
+     *
      * @since Symphony 2.3.5
      *
      * @param string|array $file
@@ -332,6 +336,7 @@ abstract class EmailGateway
                 'file' => $file,
                 'filename' => null,
                 'charset' => null,
+                'cid' => false,
             );
 
             // is array, but not the right key
@@ -342,6 +347,7 @@ abstract class EmailGateway
                 'file' => $file[$keys[0]],
                 'filename' => is_numeric($keys[0]) ? null : $keys[0],
                 'charset' => null,
+                'cid' => false,
             );
         }
 
@@ -630,7 +636,7 @@ abstract class EmailGateway
 
             if ($file_content !== false && !empty($file_content)) {
                 $output .= $this->boundaryDelimiterLine('multipart/mixed')
-                     . $this->contentInfoString($file['mime-type'], $file['file'], $file['filename'], $file['charset'])
+                     . $this->contentInfoString($file['mime-type'], $file['file'], $file['filename'], $file['charset'], $file['cid'])
                      . EmailHelper::base64ContentTransferEncode($file_content);
             } else {
                 if ($this->_validate_attachment_errors) {
@@ -703,10 +709,11 @@ abstract class EmailGateway
      * @param string $file optional the path of the attachment
      * @param string $filename optional the name of the attached file
      * @param string $charset optional the charset of the attached file
+     * @param boolean $cid  optional add a generated Content-ID header field
      *
      * @return array
      */
-    public function contentInfoArray($type = null, $file = null, $filename = null, $charset = null)
+    public function contentInfoArray($type = null, $file = null, $filename = null, $charset = null, $cid = false)
     {
         // Common descriptions
         $description = array(
@@ -749,12 +756,15 @@ abstract class EmailGateway
             $type = General::getMimeType($file);
         }
         // Return binary description
-        return array(
+        $bin = [
             'Content-Type'              => $type.';'.$charset.' name="'.$filename.'"',
             'Content-Transfer-Encoding' => 'base64',
-            'Content-ID'                => "<$filename>",
             'Content-Disposition'       => 'attachment; filename="' .$filename .'"',
-        );
+        ];
+        if ($cid) {
+            $bin['Content-ID'] = "<$filename@" .  DOMAIN . '>';
+        }
+        return $bin;
     }
 
     /**
@@ -764,9 +774,9 @@ abstract class EmailGateway
      *
      * @return string|null
      */
-    protected function contentInfoString($type = null, $file = null, $filename = null, $charset = null)
+    protected function contentInfoString($type = null, $file = null, $filename = null, $charset = null, $cid = false)
     {
-        $data = $this->contentInfoArray($type, $file, $filename, $charset);
+        $data = $this->contentInfoArray($type, $file, $filename, $charset, $cid);
         $fields = array();
 
         foreach ($data as $key => $value) {
--- a/symphony/lib/toolkit/class.emailgateway.php
+++ b/symphony/lib/toolkit/class.emailgateway.php
@@ -251,7 +251,7 @@ abstract class EmailGateway
 
     /**
      * This function sets one or multiple attachment files
-     * to the email.
+     * to the email. It deletes all previously attached files.
      *
      * Passing `null` to this function will
      * erase the current values with an empty array.
@@ -310,6 +310,7 @@ abstract class EmailGateway
      *   When set to true, the Content-ID header field is added with the filename as id.
      *   The file array can contain a 'disposition' key.
      *   When set, it is used in the Content-Disposition header
+     * @throws EmailGatewayException if the parameter is not valid.
      *
      * @since Symphony 2.3.5
      *
@@ -319,6 +320,7 @@ abstract class EmailGateway
      *
      *   Can also be a keyed array. This will enable more options, like setting the
      *   charset used by mail agent to open the file or a different filename.
+     *   Only the "file" key is required.
      *
      *   Example:
      *   ````
@@ -327,30 +329,22 @@ abstract class EmailGateway
      *      'filename' => 'bar.txt',
      *      'charset' => 'UTF-8',
      *      'mime-type' => 'text/csv',
+     *      'cid' => false,
+     *      'disposition' => 'inline',
      *   ));
      *   ````
      */
     public function appendAttachment($file)
     {
         if (!is_array($file)) {
-            // treat the param as string (old format)
+            // treat the param as string
             $file = array(
                 'file' => $file,
-                'filename' => null,
-                'charset' => null,
-                'cid' => false,
             );
 
             // is array, but not the right key
         } elseif (!isset($file['file'])) {
-            // another (un-documented) old format: key is filename
-            $keys = array_keys($file);
-            $file = array(
-                'file' => $file[$keys[0]],
-                'filename' => is_numeric($keys[0]) ? null : $keys[0],
-                'charset' => null,
-                'cid' => false,
-            );
+            throw new EmailGatewayException('The file key is missing from the attachment array.');
         }
 
         // push properly formatted file entry
@@ -639,12 +633,12 @@ abstract class EmailGateway
             if ($file_content !== false && !empty($file_content)) {
                 $output .= $this->boundaryDelimiterLine('multipart/mixed')
                      . $this->contentInfoString(
-                            $file['mime-type'],
+                            isset($file['mime-type']) ? $file['mime-type'] : null,
                             $file['file'],
-                            $file['filename'],
-                            $file['charset'],
-                            $file['cid'],
-                            $file['disposition']
+                            isset($file['filename']) ? $file['filename'] : null
+                            isset($file['charset']) ? $file['charset'] : null
+                            isset($file['cid']) ? $file['cid'] : null
+                            isset($file['disposition']) ? $file['disposition'] : 'attachment'
                         )
                      . EmailHelper::base64ContentTransferEncode($file_content);
             } else {
--- a/symphony/lib/toolkit/class.emailgateway.php
+++ b/symphony/lib/toolkit/class.emailgateway.php
@@ -307,7 +307,7 @@ abstract class EmailGateway
      *
      * @since Symphony 3.0.0
      *   The file array can contain a 'cid' key.
-     *   When set to true, the Content-ID header field is added.
+     *   When set to true, the Content-ID header field is added with the filename as id.
      *   The file array can contain a 'disposition' key.
      *   When set, it is used in the Content-Disposition header
      *
@@ -718,7 +718,7 @@ abstract class EmailGateway
      * @param string $file optional the path of the attachment
      * @param string $filename optional the name of the attached file
      * @param string $charset optional the charset of the attached file
-     * @param boolean $cid  optional add a generated Content-ID header field
+     * @param string|boolean $cid optional add a Content-ID header field. If true, uses the filename as the cid
      * @param string $disposition optional the value of the Content-Disposition header field
      *
      * @return array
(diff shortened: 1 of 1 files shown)


michael-e:
Can you add some references? I never heard about this, so some specs/articles might be helpful.

michael-e:
`<opinionated>` Sending all email images as attachments means a crude user experience `</opinionated>`, and I have not seen that anymore in the last two years or so.

That being said, the feature is defined in the RFC 2392, so it is a standards thingie. Here's some information I found on the web:

* https://www.ietf.org/rfc/rfc2392.txt
* https://sendgrid.com/blog/embedding-images-emails-facts/
* http://mailformat.dan.info/headers/mime.html

**But if we implement this, we should do it correctly.**

1. The format should be `<foo@bar.net>`, where the domain name is the FQDN of the sending server.

    The email process will not know the FQDN of the server (nor will the developer, sometimes). Could we use the domain name instead? (The Email Newsletter Manager, for example, uses Symphony's `URL` constant as `pseudo_root` parameter for the background process.)

2. URL Encoding is needed as well. The RFC describes the process precisely (but provides a wrong example IMHO):

        A "cid" URL is converted to the corresponding Content-ID message
        header [MIME] by removing the "cid:" prefix, converting the % encoded
        character to their equivalent US-ASCII characters, and enclosing the
        remaining parts with an angle bracket pair, "<" and ">".  For
        example, "cid:foo4%25foo1@bar.net" corresponds to

          Content-ID: <foo4%25foo1@bar.net>

        Reversing the process and converting URL special characters to their
        % encodings produces the original cid.

(Given the description, I would expect `Content-ID: <foo4%foo1@bar.net>`.)

michael-e:
Regarding the necessity to URL encode, I don't really understand the RFC. In chapter 2 I read:

> Since addr-spec in a Message-ID or Content-ID might contain
     characters not allowed within a URL; any such character (including
     "/", which is reserved within the "mid" scheme) must be hex-encoded
     using the %hh escape mechanism in [URL].

which means that the `Content-ID` value must also be encoded.

Does anybody understand this?

nitriques:
>  Sending all email images as attachments means a crude user experience

It makes them works offline. It makes is possible to save the email as a .eml file and have all assets to print it at a later date. It also makes it possible to send image that should not be put online (discoverable) but still display them 'inline'.

1. So it would be filename.ext@domain ? Using the `DOMAIN` const would be just fine. 
2. I do not see how the example is bad. Yes the `Content-ID` can be encoded as well, that's what the example shows :)

But the first text reads that the process starts from the url found in the content (that part is not 100% clear), then removes the cid: part, then % decode the chars, then adds <>, then does the lookup in the headers. So the cid:url in the html may or may not be % encoded, but the header value must be because the lookup will be done that way.

May or may not: This is the tricky part (second excerpt). No relationship between the too values is assumed. Some characters are invalid in URLs. Some characters are invalid in the headers (which may be a different set then in a URL). Having a < in your content-id would screw it up, but can be found in a url, hence you need to % encode it in the header but would not be needed in the template.

Thanks for sharing the ❤️  for RFCs. It's written to be able to make a change at one place and the document is still valid.

michael-e:
This will be a bit more complicated. :-)

Regarding the [example in the RFC 2392](https://github.com/symphonycms/symphony-2/pull/2617#issuecomment-237504736): If you look at [the older RFC 2111](https://www.ietf.org/rfc/rfc2111.txt) you will find exactly what I expected. So what is right? I am not sure, maybe it's indeed a mistake in the newer RFC.

> So it would be filename.ext@domain ? Using the DOMAIN const would be just fine.

Yep, probably.

Regarding your pull request: Are you sure that such an "item" still has a `Content-Disposition` header as well? In the RFC examples, I don't see it. But your PR simply adds a `Content-ID` header in addition to the `Content-Disposition` header.

If I had more time, I would surely dive deeper. But at the moment, unfortunately, I can only express my concerns.

nitriques:
> So what is right? 

Or more like: What's the implementation in the wild ? :P

> still has a Content-Disposition header as well?

If the spec does not clearly states that they must not be both present, I do not see a problem.

If you feel like this is risky, we could add it behind a flag. It could be a site wide or local call setting.

I've done much testing and did not broke any client and it worked in my 2007 outlook.

Mail on OS X showed the image inline and as an attachment, while outlook and gmail was only displaying the inline image.

michael-e:
> Mail on OS X showed the image inline and as an attachment, while outlook and gmail was only displaying the inline image.

So could you test what happens if you remove the Content-Disposition header? Maybe the file is displayed inline in all clients without showing up as an attachment? (One might double-check this with Spam E-Mails, which also use inline images a lot.)

nitriques:
Will do ASAP.

nitriques:
I'll add a commit to enable it being a flag, to make sure I do not break stuff.

nitriques:
I've added a commit which requires the $file array to have a 'cid' key == true

michael-e:
LGTM. I guess you have tested everything?

nitriques:
@michael-e Yup yup I get `Content-ID: <exemple.html@exemple.com>` :D

michael-e:
Ah, looking at the code… you have not encoded the filename? Shouldn't this be done? (Forgive me, this disucussion is rather old, so I don't remember all the details.)

nitriques:
Ah damn you are right, I'll do it.

michael-e:
I am not sure. I read all the stuff again, and I still don't get it.

michael-e:
Wait. If at all, it should probably be `rawurlencode`, because this function does hex encodeing according to RFC 3986.

The problem lies in "if at all". I am definitely not sure about it. I may take another look tomorrow.

michael-e:
BTW, here's a small test to see the main difference between `urlencode` and `rawurlencode`:

```php
<?php

$filename = 'foo bar@example.com';
echo urlencode($filename) . PHP_EOL;
echo rawurlencode($filename) . PHP_EOL;
```

The output is:

```
foo+bar%40example.com
foo%20bar%40example.com
```

michael-e:
I spent some time on this.

As a first note, I am no longer sure about the purpose of your latest commit ("Add a flag to enable Content-ID header field"). Doesn't this mean that only programmatic (custom) calls of the mail routines can enable the content-id on attachments? Why?

Apart from that I see several problems here:


### Filenames are not unique

You use the filename to create the identifier. But:

>     A implementation conforming to this specification is required to
>     support the "mid" URL long form (message-id/content-id).  Conforming
>     implementations can choose to, but are not required to, take
>     advantage of the content-id's uniqueness and interpret a "cid" URL to
>     refer to any body part within the message store.

So content-ids are considered to be unique inside a "message store" (which probably means "all emails you have in your email reader"?). So IMHO using the filename as an identifier is dangerous (and it violates the RFC if you can not guarantee its uniqueness). An email program might refer to a different resource with the same filename.

Still not convinced? You also find this in the metioned RFC:

>     Both message-id and content-id are required to be globally unique.
>     That is, no two different messages will ever have the same Message-ID
>     addr-spec; no different body parts will ever have the same Content-ID
>     addr-spec.  A common technique used by many message systems is to use
>     a time and date stamp along with the local host's domain name, e.g.,
>     950124.162336@XIson.com.


### RFC is "broken" regarding URL encoding

>     A "cid" URL is converted to the corresponding Content-ID message
>     header [MIME] by removing the "cid:" prefix, converting the % encoded
>     character to their equivalent US-ASCII characters, and enclosing the
>     remaining parts with an angle bracket pair, "<" and ">".  For
>     example, "cid:foo4%25foo1@bar.net" corresponds to
>
>       Content-ID: <foo4%25foo1@bar.net>
>
>     Reversing the process and converting URL special characters to their
>     % encodings produces the original cid.es that:

IMHO this is a contradiction.

Even if we say that the "Content-ID" line in the above description is a typo, it still contradicts the description of the MID aund CID URL schemes in the same RFC document:

>     The URLs take the form
>
>       content-id    = url-addr-spec
>
>       message-id    = url-addr-spec
>
>       url-addr-spec = addr-spec  ; URL encoding of RFC 822 addr-spec
>
>       cid-url       = "cid" ":" content-id
>
>       mid-url       = "mid" ":" message-id [ "/" content-id ]
>
>       Notes: In Internet mail messages, the addr-spec in a Content-ID
>       [MIME] or Message-ID [822] header is enclosed in angle brackets
>       (<>).  Since addr-spec in a Message-ID or Content-ID might contain
>       characters not allowed within a URL; any such character (including
>       "/", which is reserved within the "mid" scheme) must be hex-encoded
>       using the %hh escape mechanism in [URL].

This tells me that addr-spec is already hex-encoded, so why the hell should this be encoded once again to build the CID URL?


### What can we do?

Building unique identifiers using (encoded) timestamps and random numbers, as proposed in the RFC and elsewhere on the web (e.g. https://www.jwz.org/doc/mid.html) would solve both problems. (Those IDs wouldn't need any hex-encoding.)

However, I have no idea how this could be implemented. (How would you make the templating process know these identifiers?)

(We might, of course, say that it is the developer's responsibility to only use unique and unproblematic filenames like "950124.162336". But, automating this would require a special upload field or similar. Which would introduce a "secret" dependency.)

michael-e:
And BTW, regarding my [comment about the Content-Disposition header](https://github.com/symphonycms/symphony-2/pull/2617#issuecomment-257463241): I have only found one example usage of Content-IDs in my emails, an email generated by a big email marketing company. They did not use this header on the image that was referenced:

```
<img src=3D"cid:=
image.1518182640619.1803@146716b4c084892efdd355e0c3b" alt=3D"" width=3D"500=
" height=3D"387" border=3D"0" />

...

------=_Part_15970192_1253472893.1518182683895
Content-Type: image/png
Content-Transfer-Encoding: base64
Content-ID: <image.1518182640619.1803@146716b4c084892efdd355e0c3b>
```

This prevented the image from showing up as an attachment below the email body in Mail (MacOS). (You could still use the dropdown menu on top of the email message to save the image.)

So one might consider removing the Content-Disposition header when a Content-ID is added. This needs more example emails and/or testing!

nitriques:
I went ahead and re-read the whole RFC. Like you've said, the RFC does contains an error (the content id should be "oo4%foo1" not "oo4%25foo1". The header does not needs to be % encoded, the html can be % encoded. So the last commit is not usefull at all.

I went ahead and also check other implementations. Microsoft lets you set anything you want https://msdn.microsoft.com/en-us/library/microsoft.exchange.webservices.data.attachment.contentid(v=exchg.80).aspx. From my testing, this was working with Outlook, Apple Mail and gmail, hence my first commit.

Apache generates a random string. https://commons.apache.org/proper/commons-email/userguide.html but pretty hard to use in templates.

> How would you make the templating process know these identifiers?

I think this is why nobody ever bothered about it: It does not really needs to be unique in the whole world, just unique in the current email. I think the RFC went too far and nobody ever cared about that fact.

> And BTW, regarding my comment about the Content-Disposition header: 

I think the developer should have control over this! Let me rebase and get rid of 
41f4019 and add a new commit.

nitriques:
After other tests, no mail client seems to bother with duplicate id in different emails.

nitriques:
So in the end, I've decided to:

1. Keep it simple­. The RFC does not seems to describe what's implemented in the wild.
2. Require dev opt-in. If not, the behavior does not change from 2.7.x.
3. Let the developer controls both content-id and content-disposition values.
4. Provide sane and good defaults.

I think that the current state of the PR reflects that.

symphony/lib/toolkit/class.emailgateway.php
```diff
@@ -332,6 +338,7 @@ public function appendAttachment($file)
                 'file' => $file,
                 'filename' => null,
                 'charset' => null,
+                'cid' => false,
```
michael-e: I think we don't need to add `cid` here, because it has a default value of `null` anyway in `EmailGateway::contentInfoString` and `EmailGateway::contentInfoArray` alike.

symphony/lib/toolkit/class.emailgateway.php
```diff
@@ -342,6 +349,7 @@ public function appendAttachment($file)
                 'file' => $file[$keys[0]],
                 'filename' => is_numeric($keys[0]) ? null : $keys[0],
                 'charset' => null,
+                'cid' => false,
```
michael-e: Same here: I think we don't need to add `cid`.

nitriques: > that doesn't need to be "modernized" anyway.

Why ? Why should we keep 2 syntaxes ?

symphony/lib/toolkit/class.emailgateway.php
nitriques: For PHP to be happy, you are right. I think we might add some isset() checks to make PHP super happy. And it felt like this was like documentation also. WDYT ?

symphony/lib/toolkit/class.emailgateway.php
michael-e: I wouldn't mind removing that legacy stuff. It says "another (un-documented) old format". :-)

https://github.com/symphonycms/symphony-2/blob/3da6eeb9a87096b8ece5f8247830759929f3b107/symphony/lib/toolkit/class.emailgateway.php#L338-L345

symphony/lib/toolkit/class.emailgateway.php
michael-e: If you want to explicitely set params to their default values you should add `'disposition' => 'attachment'` as well. :-)

Honestly, I don't get your idea here. To me that's not documentation. The comment sounds like "this is old rubbish": https://github.com/symphonycms/symphony-2/blob/3da6eeb9a87096b8ece5f8247830759929f3b107/symphony/lib/toolkit/class.emailgateway.php#L330, so for Symphony 3.0 you might even decide to no longer support this legacy format for setting attachments. I wouldn't oppose removing this block altogether.

nitriques: > you might even decide to no longer support this legacy format for setting attachments.

Sorry, I've misunderstood what you meant. Yeah let's get rid of the old format !!

I'll remove the "doc" values ;)

symphony/lib/toolkit/class.emailgateway.php
nitriques: Good!

nitriques:
@michael-e I've fixed the issues in 24786e1. Tell me what you think!

nitriques:
@michael-e are you good here ?

symphony/lib/toolkit/class.emailgateway.php
```diff
@@ -630,7 +632,14 @@ protected function getSectionAttachments()
 
             if ($file_content !== false && !empty($file_content)) {
                 $output .= $this->boundaryDelimiterLine('multipart/mixed')
-                     . $this->contentInfoString($file['mime-type'], $file['file'], $file['filename'], $file['charset'])
+                     . $this->contentInfoString(
+                            isset($file['mime-type']) ? $file['mime-type'] : null,
+                            $file['file'],
+                            isset($file['filename']) ? $file['filename'] : null
+                            isset($file['charset']) ? $file['charset'] : null
+                            isset($file['cid']) ? $file['cid'] : null
```
michael-e: The three lines above are missing a trailing comma. PHP doesn't like this, really.

nitriques: ;) I really wonder why my "lint dirty php" did not caught this... Thanks!

nitriques:
>  I have only reviewed the code and tested some "standard email functionalities" (simple emails and emails with attachments) which I use in my projects. I have not noticed any change in behaviour.

Great, I've taken care of testing the new behaviour

symphony/lib/toolkit/class.emailgateway.php
nitriques: See 5c759921ae

Issue #2560, opened 2016-02-10
DataSource::determineFilterType()'s doc is wrong


nitriques:
```
By default, all Symphony filters are considering to be AND filters, that is
they are all used and Entries must match each filter to be included. It is
possible to use OR filtering in a field by using an + to separate the values.
eg. If the filter is test1 + test2, this will match any entries where this field
is test1 OR test2. This function is run on each filter (ie. each field) in a
datasource
```

This is plain wrong, since in the code, it's the opposite (by default, it's OR and + and used for AND)

nitriques:
cc @ChoboHub

PR #2619, opened 2016-08-15
Refactor data source filter type doc #2560

ChoboHub:
Responding to DataSource::determineFilterType()'s doc is wrong #2560.


nitriques:
Merged as 2a7360ae

PR #2622, opened 2016-08-19
Fix problems with not: and ands

beaubbe:
The logic for filtering by system:id was broken when there were "and". The filter string was only splitted by "+" instead of splitting all vales by ",". But without splitting by ",", all the IDs were lost when converted to int.

Also, the "not:" were only handled on the first part of an AND statement. If there was a second "not:", it was not handled


nitriques:
Merged as 63809edb

PR #2623, opened 2016-09-09
Correct check for undefined variable

mathiasrw:
The code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it should compare to `'undefined'`

---

This change is [Reviewable](https://reviewable.io/reviews/symphonycms/symphony-2/2623)


nitriques:
Thanks !

PR #2648, opened 2017-02-11
Fix for MySQL ANSI compliance issue with order by clause and MySQL 5.7

pwnell:
After upgrading to MySQL 5.7 on Ubuntu 16.04.1 LTS, the datasource for navigation stopped working with this error:

`Expression #1 of ORDER BY clause is not in SELECT list, references column 'symphony.s.sortorder' which is not in SELECT list; this is incompatible with DISTINCT`

The attached code makes the SQL compliant and fixes the issue.   This is my first pull request so I hope I have done it correctly.  It is against 2.7.x as I am using PHP 7.


nitriques:
Merged as c913fd12 Thanks!

PR #2654, opened 2017-04-04
Add unattended installation process

nitriques:
This commit adds the ability to install Symphony without the need to
fill the install form manually. The only thing needed to implement it is
to fake POST data with a php file, namely manifest/unattend.php.

The unattended process can launch the installation process or simply
prefill the form. This is done with the 'action' array.

If something goes wrong, the form will be shown, with all the error
messages. Next form POSTs will override values coming from the
unattend.php file, giving you a quick way to fix the problems.

This feature was motivated with the need to be able to automatize the
install process without much change to the current code base.

Finally, a scaffolding version of a unattend.php file is provided in the
install/includes folder.

---

I'd like to have review on this as much as possible! Feedback is welcome!

### DIFF
--- a/install/lib/class.installer.php
+++ b/install/lib/class.installer.php
@@ -4,6 +4,8 @@
 
     class Installer extends Administration
     {
+        private static $POST = array();
+
         /**
          * Override the default Symphony constructor to initialise the Log, Config
          * and Database objects for installation/update. This allows us to use the
@@ -39,6 +41,9 @@
             // Initialize error handlers
             GenericExceptionHandler::initialise(Symphony::Log());
             GenericErrorHandler::initialise(Symphony::Log());
+
+            // Copy POST
+            self::$POST = $_POST;
         }
 
         /**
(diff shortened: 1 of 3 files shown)


brendo:
Ah this is a lighter touch than #2604!

This LGTM, probably the best we'll get without heavy refactoring 👍

install/lib/class.installer.php
```diff
@@ -380,6 +392,40 @@ private static function __checkConfiguration()
...
+         * This function checks if there is a unattend.php file in the MANIFEST folder.
+         * If it finds one, it will load it and check for the $settings variable.
+         * It will also merge the default config values into the 'fields' array.
+         *
+         * You can find an empty version at install/include/unattend.php
+         *
+         * @return array
+         *   An associative array of values, as if it was submitted by a POST
+         */
+        private static function __checkUnattended()
```
brendo: This private static notion is a bit of an anti pattern.

nitriques: I know, but I tried to kept things as they currently are. There will be a major refactor of this soon ;)

nitriques:
> What's interesting though is config_default.php already exists as a concept.

The problem with this file is that's is committed in git. You could theoretically use it as a pre-filler. But it won't automatically install the whole thing.

Would you merge it in @brendo ?

nitriques:
I had the same problem. I think this a an advanced feature. I do not think that it will be in the README, a wiki entry would suffice. I do create all my manifest folder beforehand already and the manifest folder **is** the safest place.

The question is, which is thing makes it less secure ? I do not know that.

The installer won't delete the file, but users should. Again, advanced feature.

Since both @brendo and @michael-e seems to agree, I would consider this as "mergeable". I'll still keep it open for reviews and comments.

nitriques:
Review period is closing, will be merged as is soon. Thanks y'all.

PR #2659, opened 2017-04-21
Prevent data loss on save (entries and sections)

nitriques:
We add a timestamp check before saving records.

When saving a section, if the data changes between the render of the
page and the form post, it can lead to data loss if a field was added.
The SQL will crash because it assumes the fields did not change.

When saving and entry, the SQL was ok, but still, it can lead to data
loss.

At both places, the fix provided here makes sure we are editing the
lastest version of the record.

cc @brendo @michael-e

### DIFF
--- a/symphony/content/content.blueprintssections.php
+++ b/symphony/content/content.blueprintssections.php
@@ -600,7 +600,10 @@ class contentBlueprintsSections extends AdministrationPage
                 $navigation_group = preg_replace('/^set-navigation-group-/', null, $_POST['with-selected']);
 
                 foreach ($checked as $section_id) {
-                    SectionManager::edit($section_id, array('navigation_group' => urldecode($navigation_group)));
+                    SectionManager::edit($section_id, array(
+                        'navigation_group' => urldecode($navigation_group),
+                        'modification_author_id' => Symphony::Author()->get('id'),
+                    ));
                 }
 
                 redirect(SYMPHONY_URL . '/blueprints/sections/');
@@ -734,6 +737,9 @@ class contentBlueprintsSections extends AdministrationPage
                      */
                     Symphony::ExtensionManager()->notifyMembers('SectionPreCreate', '/blueprints/sections/', array('meta' => &$meta, 'fields' => &$fields));
 
+                    $meta['author_id'] = Symphony::Author()->get('id');
+                    $meta['modification_author_id'] = $meta['author_id'];
+
                     if (!$section_id = SectionManager::add($meta)) {
                         $this->pageAlert(__('An unknown database occurred while attempting to create the section.'), Alert::ERROR);
                     }
@@ -760,6 +766,8 @@ class contentBlueprintsSections extends AdministrationPage
                      *  and the value being a Field object, passed by reference
                      */
                     Symphony::ExtensionManager()->notifyMembers('SectionPreEdit', '/blueprints/sections/', array('section_id' => $section_id, 'meta' => &$meta, 'fields' => &$fields));
+
+                    $meta['modification_author_id'] = Symphony::Author()->get('id');
 
                     if (!SectionManager::edit($section_id, $meta)) {
                         $canProceed = false;
--- a/symphony/content/content.blueprintssections.php
+++ b/symphony/content/content.blueprintssections.php
@@ -359,7 +359,6 @@ class contentBlueprintsSections extends AdministrationPage
 
         if (isset($_POST['fields'])) {
             $fields = array();
-
             if (is_array($_POST['fields']) && !empty($_POST['fields'])) {
                 foreach ($_POST['fields'] as $position => $data) {
                     if ($fields[$position] = FieldManager::create($data['type'])) {
@@ -368,9 +367,13 @@ class contentBlueprintsSections extends AdministrationPage
                     }
                 }
             }
+            $timestamp = isset($_POST['action']['timestamp'])
+                ? $_POST['action']['timestamp']
+                : $section->get('modification_date');
         } else {
             $fields = FieldManager::fetch(null, $section_id);
             $fields = array_values($fields);
+            $timestamp = $section->get('modification_date');
         }
 
         if (isset($_POST['meta'])) {
@@ -386,9 +389,8 @@ class contentBlueprintsSections extends AdministrationPage
             'rel' => 'canonical',
             'href' => SYMPHONY_URL . $canonical_link,
         )));
-
         $this->appendSubheading(General::sanitize($meta['name']),
-            Widget::Anchor(__('View Entries'), SYMPHONY_URL . '/publish/' . $section->get('handle'), __('View Section Entries'), 'button')
+            Widget::Anchor(__('View Entries'), SYMPHONY_URL . '/publish/' . $section->get('handle') . '/', __('View Section Entries'), 'button')
         );
         $this->insertBreadcrumbs(array(
             Widget::Anchor(__('Sections'), SYMPHONY_URL . '/blueprints/sections/'),
@@ -524,6 +526,9 @@ class contentBlueprintsSections extends AdministrationPage
         $button->setAttributeArray(array('name' => 'action[delete]', 'class' => 'button confirm delete', 'title' => __('Delete this section'), 'type' => 'submit', 'accesskey' => 'd', 'data-message' => __('Are you sure you want to delete this section?')));
         $div->appendChild($button);
 
+        $div->appendChild(Widget::Input('action[timestamp]', $timestamp, 'hidden'));
+        $div->appendChild(Widget::Input('action[ignore-timestamp]', 'yes', 'checkbox', array('class' => 'irrelevant')));
+
         $this->Form->appendChild($div);
     }
 
@@ -616,6 +621,10 @@ class contentBlueprintsSections extends AdministrationPage
             if ($edit) {
                 $section_id = $this->_context[1];
                 $existing_section = SectionManager::fetch($section_id);
+                $canProceed = $this->validateTimestamp($section_id, true);
+                if (!$canProceed) {
+                    $this->addTimestampValidationPageAlert($this->_errors['timestamp'], $existing_section, 'save');
+                }
             }
 
             // Check handle to ensure it is unique
@@ -788,9 +797,19 @@ class contentBlueprintsSections extends AdministrationPage
                             $field->set('sortorder', (string)$position);
                             $field->set('parent_section', $section_id);
 
-                            $newField = !(boolean)$field->get('id');
+                            // It is possible that the "existing" field has been deleted
+                            // so we are dealing with an invalid id.
+                            // First make sure the field table still exists and that
+                            // there is a field with the corresponding id in it
+                            $newField = !((boolean)$field->get('id') && $field->tableExists() && $field->exists());
+                            // If the field has not been found, erase the id from $_POST
+                            if ($newField) {
+                                $field->set('id', false);
+                            }
 
+                            // Save data
                             $field->commit();
+                            // Get the new id
                             $field_id = $field->get('id');
 
                             if ($field_id) {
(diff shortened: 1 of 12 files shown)


symphony/lib/toolkit/class.timestampvalidator.php
```diff
@@ -0,0 +1,50 @@
...
+    private $table = '';
+
+    /**
+     * Creates a new TimestampValidator object, for a particular table.
+     * The `tbl_` prefix is automatically added.
+     * @param [string] $table the table name
+     */
+    public function __construct($table)
+    {
+        assert($table, 'Table cannot be falsy');
```
brendo: Shouldn't really rely on this for production...

symphony/lib/toolkit/class.sectionmanager.php
```diff
@@ -34,6 +34,11 @@ class SectionManager
      */
     public static function add(array $settings)
     {
+        $defaults = array();
+        $defaults['creation_date'] = $defaults['modification_date'] = DateTimeObj::get('Y-m-d H:i:s');
+        $defaults['creation_date_gmt'] = $defaults['modification_date_gmt'] = DateTimeObj::getGMT('Y-m-d H:i:s');
+        $defaults['author_id'] = '1';
```
brendo: Noticed below it's `1`, any reason for a string here?

jonmifsud:
@nitriques nice touch was looking at this from an extension perspective to provide a `locking` functionality showing that another user is editing.

From what I can see however (code inspection) there is no way someone can confirm an `override`. So if I had two tabs open and/or there was another user who saved. But I actually want to save my version what does actually happen?

Secondly if I save and get the error, I'm presuming the post data would be pre-populated as per existing functionality. If so wouldn't saving again have the right time-stamp or is that also in a way maintained through post? Might have missed something as I didn't try it out.

symphony/lib/toolkit/class.sectionmanager.php
nitriques: Because it's a copy of this line https://github.com/symphonycms/symphony-2/blob/2.7.x/symphony/lib/toolkit/class.entry.php#L105

Should we change both ?

symphony/lib/toolkit/class.timestampvalidator.php
nitriques: Why ? You would prefer to throw ?

nitriques:
> So I'm assuming the position is that this protection will not be enabled for entries saved from Event's? This is purely a backend only feature to help in scenarios where there are multiple authors?

@brendo Indeed. Right now, the validation is in the publish pages. I did not wanted to break things with this feature. But any entry updated via the frontend will get a new mod_date, so the validation will fail for authors. 

> But I actually want to save my version what does actually happen?

Would have to start from scratch. I would hope to have time to do some diffs and try to resolve conflicts, but hey. that's hard.

> Secondly if I save and get the error, I'm presuming the post data would be pre-populated as per existing functionality.

Exactly. Right now, the save will NEVER work. You have to refresh this page and start from scratch.

This is why it's a PR ;) Let's debate.

symphony/lib/toolkit/class.timestampvalidator.php
brendo: The behaviour of assert depends on [php configuration](http://php.net/assert) values. The idea is usually that development code will throw errors, but production instances optimise away the calls.

I'd prefer regular throwing in this scenario.

symphony/lib/toolkit/class.sectionmanager.php
brendo: Yeah I think both should become integers.

michael-e:
I understand the reasoning, but I still try to understand the solution.

Does this mean that still two people can open the same entry, but only the first save will be successful? This wouldn't be any better than "the last save wins" (which is what we have now).

Also, I think that any "lock" needs the possibilty to "override" the lock.

Shouldn't a proper solution prevent access to an entry that is edited currently (by another author)? (Also this would need the possibility to override the lock.)

symphony/lib/toolkit/class.timestampvalidator.php
nitriques: Fixed in d9227959. Thanks!

symphony/lib/toolkit/class.sectionmanager.php
nitriques: Fixed in d1d7ddfe

nitriques:
> Does this mean that still two people can open the same entry, but only the first save will be successful?

Yes.

> This wouldn't be any better than "the last save wins"

I do not agree. We had calls from client because of this, and it took the activity tracker to prove to the client that this was happening. From a UX point of view, it's better IMHO.

I think that the current error message could be improve, to let the user know more about what happened, like who did the last save and on what time.

michael-e:
As I said before, I think that a proper solution should in advance prevent access to an entry that is currently edited (by another author). With your solution, authors will ask "Why have you allowed me to edit an entry if I am not allowed to save my work?". There is no good answer to that.

One way or another, there should be the possibility to override the "lock", i.e. save nevertheless.

jonmifsud:
Agreed re override. That's a must have.

Might also debate if this is something for the core to handle.. at least I
would be happy with this working as you suggested + another flag override
entry which would skip/ignore the check. Could be a checkbox in the UI -
but equally used by other API's to update entries regardless. So if I get
the error modal I don't need to re-start but I can say Ignore the other
dude's changes I want mine.

nitriques:
>  I think that a proper solution should in advance prevent access to an entry that is currently edited

That's not always possible. Let's say I open the edit page and go make a coffee. I come back, edit and then press save: only then, do the system know that it's trying to edit based on old records.

> One way or another, there should be the possibility to override the "lock", i.e. save nevertheless.

Ok this seems to make consensus, I'll do it. Thanks guys!

nitriques:
> Why are we moving away from Symphony::Database() in favor of MySQL

Right now, `cleanValue` is static, so it should be called with the class name. This needs to change.

As for the UI, I've come up with this:

image

michael-e:
"The action requested…" sounds complicated. What about "Saving failed. The entry has been changed by Test on…"?

And instead of "Override? Please understand…" I would prefer "Overwrite these changes?".

This should be decided by a native speaker. @brendo, ah, no, he's Australian… Hmmm… :-)

michael-e:
> Right now, `cleanValue` is static, so it should be called with the class name. This needs to change.

I understand. Thanks.

nitriques:
> This should be decided by a native speaker. @brendo, ah, no, he's Australian… Hmmm… :-)

Hahaha. But yeah, I need help with that haha

nitriques:
I've got override working, and I'm back with the non-existing table problem hahaha

nitriques:
1914abc is the fix I've meant to not do, buy creating the timestamp validation. But with overriding in place, it still needed to be done. I hope the fix does not suck as much as I think it sucks, but hey, it the only was to do it.

@brendo if you could get your blessing regarding the wording... @michael-e and @jonmifsud suggested

```
Saving failed. The entry has been changed by Test on…

[ ] Don’t care about Joe’s changes? Yes I would much prefer if Joe’s changes never existed!
```

nitriques:
@michael-e @jonmifsud @brendo would you mind checking the last commit (fc9adc4). @michael-e I think we need to tell the user at least why Saving "failed" (I prefer aborted since that what's happening)

symphony/lib/toolkit/class.administrationpage.php
```diff
@@ -1261,7 +1262,9 @@ public function addTimestampError($errorMessage, $existingObject, $options = arr
         ));
         $errdiv->appendChild(new XMLElement('label',
             Widget::Input('action[ignore-timestamp]', null, 'checkbox')->generate() . ' ' .
-            __('Override? Please understand that data loss can occur.')
+            __("I don't care about %s's changes, pretend it never existed.", array(
```
michael-e: Shouldn't this be "pretend they…" or "pretend these…"?

Anyway it sounds complicated. What about "I don't care about %s's changes. Save my version."

symphony/content/content.blueprintssections.php
```diff
@@ -967,12 +967,12 @@ protected function validateTimestamp($section_id, $checkMissing = false)
...
         if (!isset($_POST['action']['ignore-timestamp'])) {
             if ($checkMissing && !isset($_POST['action']['timestamp'])) {
-                $this->_errors['timestamp'] = __('The timestamp is missing.');
+                $this->_errors['timestamp'] = __('Saving aborted, timestamp is missing.');
                 return false;
             } else if (isset($_POST['action']['timestamp'])) {
                 $tv = new TimestampValidator('sections');
                 if (!$tv->check($section_id, $_POST['action']['timestamp'])) {
-                    $this->_errors['timestamp'] = __('The action requested was made on a earlier version of the record.');
+                    $this->_errors['timestamp'] = __('Saving aborted, data integrity check failed.');
```
michael-e: "data integrity check failed" isn't very user-friendly, is it? Can we make this more clear?

symphony/lib/toolkit/class.administrationpage.php
nilshoerrmann: This copy sound quite rude …

brendo:
Mm, I wonder if the in field approach could be refined. At the moment, any 'errors' that occur at a generic level are introduced with a page alert. We have a convention where links can appear as well, and clicking on the link perform actions (usually navigation).

Would it be suitable to instead display a page alert along the lines of:

`This entry has been changed by %s %s. _Replace anyway?_ or _View %s's entry_`

eg. `This entry has been changed by Brendan a few moments ago. Replace anyway? or View Brendan's entry`.

I'm not sure if the `Brendan's` bit translates well though, how does it read in German/French?

This cleans up the interface and doesn't introduce a new field, a pattern that doesn't exist. Items in the main area are always the result of field or extension, never by the core.

nilshoerrmann:
I like this approach.

> I'm not sure if the Brendan's bit translates well though, how does it read in German/French?

It's possible translation-wise: the German equivalent would be "Zeige Brendans Eintrag". But you'll run into trouble with names like mine which should be inserted as "View Nils' entry" and not "View Nils's entry". But it's possible to work that out with good copywriting.

brendo:
> But it's possible to work that out with good copywriting.

Yeah I'm wondering if other terms could work:

- `View changed entry`
- `View changes` (might imply diff like view though?)
- `View new entry`
- `View updated entry`
- `Show entry`

nilshoerrmann:
`View other entry`?

nitriques:
Yes, debate, ideas!  I like Brendon's idea. It was my first idea, but it felt wrong with a check box in it...
Let me send some more commits ;)

brendo:
Why does it need a checkbox? Could it be hidden and it's value changed by JS? Or checked by CSS target

nitriques:
> Why does it need a checkbox? Could it be hidden and it's value changed by JS? Or checked by CSS target

It does not!!! Your solution is completely valid! I just did not thought of it!

nitriques:
@brendo There is even a bonus side effect of using the alert: We can display it in index pages as well!!!

nitriques:
How about this :
image

nilshoerrmann:
What about:

> The entry could not be saved due to conflicting changes by X at Y. Dismiss changes? View changes.

brendo:
I like @nilshoerrmann wording, it's succinct. My only worry is that "Dismiss changes" is too _soft_ and doesn't clearly explain that you will essentially delete data. What about "Overwrite changes?" or "Replace changes"?

nilshoerrmann:
Both suggestions sound just fine.

I had the German word “verwerfen” in mind which translates to “dismiss” or “discard“. So maybe you should just decide, Brendan, because as a native speaker you know best what works in the English language. 

My main intention was to shorten Nicolas’ message in order to align the wording of both actions with the rest of the interface.

brendo:
If it's up to me, I'd vote for `Replace changes`

nitriques:
Yes I like this too!

image

Is it good to merge ?

nitriques:
> Is it good to merge ?

I meant can I work on finishing this feature since the UI is ok ? :P

nitriques:
This thing will be an ongoing effort, since it's "complete" (meaning things that are implemented are working), I'll merge soon.

PR #2661, opened 2017-04-24
Enable the email core to ignore attachment errors

### LINKED ISSUE #2656: Should sending emails break on missing attachments?
I noticed that sending emails breaks if attachments are set (e.g. using the Email Template Manager) but missing (unreadable/unexisting path or similar).

[This is intended](https://github.com/symphonycms/symphony-2/blob/2.7.x/symphony/lib/toolkit/class.emailgateway.php#L612-L620).

But this makes sending emails using the Email Newsletter Manager and the Email Template Manager less flexible. The definition of attachments requires the attachments to actually exist — in other words:  you can't provide email logic with *optional* attachments to your authors.

I myself already hacked the `EmailGateway` class in order to not break on missing attachments (and also set email headers and boundaries accordingly).

Is this a change that is interesting for the core?

michael-e:
Resolves #2656 

Please review the code.

When Symphony 2.7 is released, I will add a configuration option to the Email Template Manager (ETM) extension which allows to use the "ignore attachment errors" feature.

### DIFF
--- a/symphony/lib/toolkit/class.emailgateway.php
+++ b/symphony/lib/toolkit/class.emailgateway.php
@@ -57,6 +57,7 @@ abstract class EmailGateway
     protected $_text_plain;
     protected $_text_html;
     protected $_attachments = array();
+    protected $_ignore_attachment_errors = false;
     protected $_reply_to_name;
     protected $_reply_to_email_address;
     protected $_header_fields = array();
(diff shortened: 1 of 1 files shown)


symphony/lib/toolkit/class.emailgateway.php
```diff
@@ -349,6 +350,21 @@ public function appendAttachment($file)
...
+     * Sets the property `$_ignore_attachment_errors`
+     *
+     * This property is false by default, so sending will break if any attachment
+     * can not be loaded; if it is true, attachment errors error will be ignored.
+     *
+     * @since Symphony 2.7
+     * @param boolean $ignore_attachment_errors
+     * @return void
+     */
+    public function setIgnoreAttachmentErrors(bool $ignore_attachment_errors)
```
michael-e: The type declaration `bool` only works on PHP7, so I will have to remove this. I will send a different PR. :-(

michael-e:
Closing. Will send a new PR.

PR #2662, opened 2017-04-24
Enable the email core to ignore attachment errors

### LINKED ISSUE #2656: Should sending emails break on missing attachments?

michael-e:
Resolves #2656 

When Symphony 2.7 is released, I will add a configuration option to the Email Template Manager (ETM) extension which allows to use the "ignore attachment errors" feature.

### DIFF
--- a/symphony/lib/toolkit/class.emailgateway.php
+++ b/symphony/lib/toolkit/class.emailgateway.php
@@ -57,7 +57,7 @@ abstract class EmailGateway
     protected $_text_plain;
     protected $_text_html;
     protected $_attachments = array();
-    protected $_ignore_attachment_errors = false;
+    protected $_validate_attachment_errors = true;
     protected $_reply_to_name;
     protected $_reply_to_email_address;
     protected $_header_fields = array();
@@ -350,21 +350,21 @@ abstract class EmailGateway
     }
 
     /**
-     * Sets the property `$_ignore_attachment_errors`
+     * Sets the property `$_validate_attachment_errors`
      *
-     * This property is false by default, so sending will break if any attachment
-     * can not be loaded; if it is true, attachment errors error will be ignored.
+     * This property is true by default, so sending will break if any attachment
+     * can not be loaded; if it is false, attachment errors error will be ignored.
      *
      * @since Symphony 2.7
-     * @param boolean $ignore_attachment_errors
+     * @param boolean $validate_attachment_errors
      * @return void
      */
-    public function setIgnoreAttachmentErrors($ignore_attachment_errors)
+    public function setIgnoreAttachmentErrors($validate_attachment_errors)
     {
-        if (!is_bool($ignore_attachment_errors)) {
+        if (!is_bool($validate_attachment_errors)) {
             throw new EmailGatewayException(__('%s accepts boolean values only.', array('<code>setIgnoreAttachmentErrors</code>')));
         } else {
-            $this->_ignore_attachment_errors = $ignore_attachment_errors;
+            $this->_validate_attachment_errors = $validate_attachment_errors;
         }
     }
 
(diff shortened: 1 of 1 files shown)


symphony/lib/toolkit/class.emailgateway.php
```diff
@@ -610,13 +631,15 @@ protected function getSectionAttachments()
                      . $this->contentInfoString($file['mime-type'], $file['file'], $file['filename'], $file['charset'])
                      . EmailHelper::base64ContentTransferEncode($file_content);
             } else {
-                if (!$tmp_file === false) {
-                    $filename = $original_filename;
-                } else {
-                    $filename = $file['file'];
+                if ($this->_ignore_attachment_errors === false) {
+                    if (!$tmp_file === false) {
```
nitriques: Why not do `!` here ?

symphony/lib/toolkit/class.emailgateway.php
```diff
@@ -57,6 +57,7 @@ class EmailValidationException extends EmailGatewayException
     protected $_text_plain;
     protected $_text_html;
     protected $_attachments = array();
+    protected $_ignore_attachment_errors = false;
     protected $_reply_to_name;
```
nitriques: I do not like double negatives. I think it would be better to name it `_validate_attachment_errors` and set it to true.

symphony/lib/toolkit/class.emailgateway.php
michael-e: You mean changing `if (!$tmp_file === false)`? It has been like that before. I kept the existing code an just wrapped it into `if ($this->_ignore_attachment_errors === false)`. I think that changing the old code might be part of a cleanup, but I wouldn't like to do this in a PR like this one. The PR should only have one topic.

nitriques: No, I meant `if ($this->_ignore_attachment_errors === false)` .. why not `if (!$this->_ignore_attachment_errors)` ? Because of the line below ?

michael-e: Ah, I see. I just wanted to make it "precise", because `!$this->_ignore_attachment_errors` would also be true for a value of `0` or an empty string. But of course my setter function only allows boolean values—so should I use a "less strict" check here?

It might then simply be `if ($this->_validate_attachment_errors)` with the other change you propose. I would use `if ($this->_validate_attachment_errors === true)`. :-)

I like to learn, and you know better than me what is "good practice". So which version is better in this context?

nitriques: > But of course my setter function only allows boolean values—so should I use a "less strict" check here?

Yup. I do not mind embracing the dynamic aspect of PHP (or javascript for that matter). If it walks like a duck and sings like a duck, it's a duck.

> It might then simply be if ($this->_validate_attachment_errors) 

In fact!

> I like to learn, and you know better than me what is "good practice". So which version is better in this context?

Both have flaws and advantages. Being more readable (less verbose) vs being explicit. Being flexible vs. being performant. I would try to find what PSR tells about this.

michael-e: I don't think that PSR has any guidelines for this, at least I haven't found anything. :-(

I will do what you proposed.

michael-e:
Second commit according to your proposals. Thanks!

michael-e:
Next time I will test before pushing to my repo. The third commit fixes the setter function. Everything works fine now.

nitriques:
> Next time I will test before pushing to my repo.

No it's perfectly fine to push early and open the discussion. We can now rebase or squash pull requests directly on github, so it's cleanable ;)

nitriques:
See, beautiful commit:  ef81239 ;)

michael-e:
Yeah, looks great! Thanks!

PR #2664, opened 2017-04-26
Prevent acting on non existing tables (Fix #2663)

### LINKED ISSUE #2663: Symphony wants to manipulate tables that do not exist
In `2.7.x` @michael-e has introduced »table locking« to prevent data corruption. This apparently leads to problems with fields that do not create an entry table since they do not store any data. 

I’ve encountered an error with the [system_date_fields](https://github.com/symphonists/system_date_fields/issues/20)-extension. See link for the corresponding issue.

To quote Michael:

> In your setup MySQL vomits on the fact that Symphony wants to manipulate tables which do not exist. Symphony has always assumed that these tables exist.

What would be the preferable solution? 

1. symphony checking if the table actually exists, or 
2. `system_date_fields` creating an empty table?

nitriques:
This change makes sure that the changes introduced in 4b5067f38 do not
cause a bug when the table does not exists. The old code (pre 2.7.0)
discarded all errors on delete, then checked if it had data to be
inserted. If not, the insert statement would be skipped.

The locking mechanism did add a lock statement before the delete, and
added logging all exceptions. That resulted in log "vomit".

This changes propose 2 ways of fixing this log problem:
1. It moves the empty data checks BEFORE the lock (so before the delete
as well).
2. It checks in the database if the table exist before doing anything.

4b5067f38 also un-lock the table before doing the insert. I think that
it should still maintain the lock until both delete and insert are done.

Fixes #2663

cc @animaux @michael-e

### DIFF
--- a/symphony/lib/toolkit/class.mysql.php
+++ b/symphony/lib/toolkit/class.mysql.php
@@ -872,6 +872,7 @@ class MySQL
      * if the `$table` contains the `$field`.
      *
      * @since Symphony 2.3
+     * @link  https://dev.mysql.com/doc/refman/en/describe.html
      * @param string $table
      *  The table name
      * @param string $field
@@ -882,8 +883,9 @@ class MySQL
      */
     public function tableContainsField($table, $field)
     {
+        $table = MySQL::cleanValue($table);
+        $field = MySQL::cleanValue($field);
         $results = $this->fetch("DESC `{$table}` `{$field}`");
-
         return (is_array($results) && !empty($results));
     }
 
@@ -892,6 +894,7 @@ class MySQL
      * if it exists or not.
      *
      * @since Symphony 2.3.4
+     * @link  https://dev.mysql.com/doc/refman/en/show-tables.html
      * @param string $table
      *  The table name
      * @throws DatabaseException
@@ -900,8 +903,8 @@ class MySQL
      */
     public function tableExists($table)
     {
+        $table = MySQL::cleanValue($table);
         $results = $this->fetch(sprintf("SHOW TABLES LIKE '%s'", $table));
-
         return (is_array($results) && !empty($results));
     }
 
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -178,18 +178,26 @@ class EntryManager
 
             $did_lock = false;
             try {
-                $did_lock = Symphony::Database()->query('LOCK TABLES tbl_entries_data_' . $field_id . ' WRITE');
-                Symphony::Database()->delete('tbl_entries_data_' . $field_id, sprintf("
-                    `entry_id` = %d", $entry->get('id')
-                ));
-
+                // Check if we have field data
                 if (!is_array($field) || empty($field)) {
-                    if ($did_lock) {
-                        Symphony::Database()->query('UNLOCK TABLES');
-                    }
                     continue;
                 }
 
+                // Check if table exists
+                $table_name = 'tbl_entries_data_' . General::intval($field_id);
+                if (!Symphony::Database()->tableExists($table_name)) {
+                    continue;
+                }
+
+                // Lock the table for write
+                $did_lock = Symphony::Database()->query("LOCK TABLES `$table_name` WRITE");
+
+                // Delete old data
+                Symphony::Database()->delete($table_name, sprintf("
+                    `entry_id` = %d", $entry->get('id')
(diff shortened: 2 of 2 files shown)


animaux:
No errors, entry saves fine. Works at least in conjunction with `system_date_fields`.

jonmifsud:
Would adding a `containsEntryTable` function requirement on Field do the
trick in a more efficient manner?

We can add the default function definition which checks for a table
existence and stores it within the field object (if that is re-used) or
else default to true and only fields which don't have this would override
this parameter.





*Jonathan Mifsud*
Business digitization expert
  T +356 9983 0394 <+35699830394> E jon@maze.digital W maze.digital

symphony/lib/toolkit/class.mysql.php
```diff
@@ -882,8 +883,9 @@ public function fetchVar($column, $offset = 0, $query = null)
      */
     public function tableContainsField($table, $field)
     {
+        $table = MySQL::cleanValue($table);
+        $field = MySQL::cleanValue($field);
```
michael-e: Ideally, this should be a separate PR, because it has nothing to do with the topic.

michael-e:
@jonmifsud: Implementing a `containsEntryTable` method for fields would change a lot. Extensions would need to be updated. However, I see the advantage—it wouldn't slow things down.

So is @nitriques solution slowing things down too much? I wouldn't think so either. Having 3 queries per field and entry instead of 2 (delete/insert) may sound a lot. But I would expect the additional query `SHOW TABLES LIKE…` to be rather fast compared to deleting or especially inserting.

jonmifsud:
@michael-e would it change that much? As in if you implement the default
behaviour to be true (as it was before) most would still work out of the
box right? I think only this field has no `table` and maybe a handful of
others. Obviously we can always go with a `phase1` which is backwards
compatible then in the next major version add the function.





*Jonathan Mifsud*
Business digitization expert
  T +356 9983 0394 <+35699830394> E jon@maze.digital W maze.digital

michael-e:
@jonmifsud: I agree—most field extensions do have data tables, of course. But I have the feeling that @nitriques wants a more "universal" solution like "Never act on tables which do not exist. Never.". Let's see what he thinks.

symphony/lib/toolkit/class.mysql.php
nitriques: Yeah, guilty, I should have `git add . -p`

nitriques:
@brendo XML importer does update stuff ? (I've only used it for inserting)

> @nitriques wants a more "universal" solution like "Never act on tables that do not exist. Never.". Let's see what he thinks.

I sure do. Having those kind of bugs in production environment can cost a lot.

I like @jonmifsud idea tho. I could speed things up. Let's add it (?)

nitriques:
> BTW, my commit 4b5067f did not un-lock the table before doing the insert

@michael-e yup sorry interpreted the code wrong. But right now, no lock occurs for empty field (thus reducing the number of queries!)

michael-e:
> Let's add it (?)

Oh, I have nothing against it, on the contrary! If you can do the work…

nitriques:
I am all for it, I'll do it. Great!

nitriques:
We already have `requiresSQLGrouping`. What about `requiresEntryTable` ?

michael-e:
The naming scheme for the tables is `tbl_entries_data_[:id]`, so what about `requiresEntriesDataTable` or simply `hasEntriesDataTable`?

nitriques:
I'd vote for `requiresEntriesDataTable`, but the current 'table' related method is named `createTable` and we also have a `tearDown` method which, strangely, does not drop the table it created.... few naming really is the second hardest problem there is.

But our best bet would be to stick with what's in there right now and name it `requiresTable`

nitriques:
@michael-e another naming problem here :(

michael-e:
Yes, naming here can be either consistent (at least a bit) or clear… I promise that I won't cry, whatver you decide.

nitriques:
> I promise that I won't cry, whatver you decide.

Thanks hahah

nitriques:
I'll send a separate PR for the new function.

nitriques:
See #2678 for a follow up. cc @jonmifsud

pointybeard:
A bit late to the party, but this PR appears to break the ability to empty an optional field of its value. https://github.com/symphonycms/symphony-2/commit/257aab2deb688c4b6c68805b102dabd4ba880c7e#commitcomment-23450932

PR #2665, opened 2017-04-27
Replace url state with canonical url

nitriques:
I often see urls being shared with the various flags (saved, created) in them.
I also often see them while browsing the history.

Those ideally would not exists, but since we do a redirection on save, we need a way to still display the messages to the user.

The idea behind this commit is to hack the url usign javascript: if the page includes a link rel="canonical" tag in the head section, the javascript will replace the current state with the canonical url. It's also done only on full page load, to let extension read the value before replacing it.

This would not break the existing behaviour and could be removed when the redirection will be removed.

---

Maybe I should create function dedicated to adding the link tag...

### DIFF
--- a/symphony/assets/js/src/backend.js
+++ b/symphony/assets/js/src/backend.js
@@ -68,6 +68,19 @@
 
 		// Render view
 		Symphony.View.render();
+
+		// Update state to canonical url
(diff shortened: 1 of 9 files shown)


nitriques:
Great!

PR #2666, opened 2017-04-27
Add sql: null and sql: not null filters

### LINKED ISSUE #2626: Data Source UI cannot filter for empty core Text Input or Date fields
Examples:
1. I have a Section with many fields. One is an optional core Text Input field.
2. I have another Section with some fields. One is an optional core Date field.

In each of the examples above want a Data Source to filter for entries where those fields are empty.  I have tried the following filter values but none worked:
- `regexp: ^$`
- `not-regexp: ^http` (in the case of a Text Input field that was expecting a URL)
- `not-regexp: .+`
- `not-regexp: ^.+$`
- `sql: NULL` *
- `is null` *

In all filter value cases the output XML is `<error>No records found.</error>`

\* In these filter value cases, where I've selected a parameter in the UI to be output, all `<item`s are output in the `<ds-` parameter, regardless of the filter.

The opposite does work. So `regexp: ^.+` will find the entries with non-empty Date fields, for example.

Symphony version is 2.6.3.

nitriques:
Right now, there was no easy way to get empty or not empty fields via the
data source. This change adds the `sql: null` and `sql: not null` filter
values.

To prevent SQL injection, only those two cases are allowed.

Fixes #2626 

---

Also, this PR #2565, will remove the not-needed input tag when dealing with non-editable filters.

cc @cylkee

### DIFF
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1193,12 +1193,12 @@ class Field
             array(
                 'filter' => 'sql: NOT NULL',
                 'title' => 'is not empty',
-                'help' => __('Find entries where any value is selected.')
+                'help' => __('Find entries with a non-empty value.')
             ),
             array(
                 'filter' => 'sql: NULL',
                 'title' => 'is empty',
-                'help' => __('Find entries where no value is selected.')
+                'help' => __('Find entries with an empty value.')
             ),
             array(
                 'title' => 'contains',
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1191,17 +1191,27 @@ class Field
                 'help' => __('Find values that are an exact match for the given string.')
             ),
             array(
+                'filter' => 'sql: NOT NULL',
+                'title' => 'is not empty',
+                'help' => __('Find entries where any value is selected.')
+            ),
+            array(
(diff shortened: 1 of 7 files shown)


michael-e:
Great addition! Just one small thing: the textual hint (help text) says "Find entries where any/no value is **selected**." for most of the fields. The word "selected" might be a bit misleading.

nitriques:
> The word "selected" might be a bit misleading.

It's a dump copy paste from what was already in the tag field (the tag field was the only field that supported that syntax).

nitriques:
@michael-e 

Find entries where any value exists.

Find entries where no value exists.

?

nitriques:
`Find entries where a file is saved.`
`Find entries where no file is saved.`
?

michael-e:
What about:

* Find entries with non-empty value.
* Find entries with empty value.
* Find entries where a file has been saved.
* Find entries where no file has been saved.

jonmifsud:
Minor amendment 
 
- Find entries with *a* non-empty value.
 - Find entries with *an* empty value.

nitriques:
@michael-e are you ok with @jonmifsud amendment ?

michael-e:
Yes, sure! (He's definitely better in English.)

nitriques:
Committed and pushed!

michael-e:
Great.

PR #2673, opened 2017-05-02
Fixes for strict PHP error reporting

michael-e:
Symphony sets the desired (gracious) error reporting in `/lib/boot/bundle.php`:

```php
error_reporting(E_ALL & ~E_NOTICE & ~E_DEPRECATED & ~E_STRICT);
```

In order to secure code quality, some Content Management Systems (at least Wordpress) provide a way to switch to stricter error reporting during development. Symphony has no config option for this, but you can easily replace the above with the following, for example:

```php
error_reporting(E_ALL);
```

At the moment this will definitely break a lot. But we know that PHP tends to become stricter with every release, so fixing every warning or notice would probably ease the transition to PHP 8 or later.

With this pull request, I include some example fixes (mainly related to the Symphony backend, just clicking around…). They are just a start, and they must be reviewed.

Fixing everything would be a big undertaking, and it should be done by people who know PHP better than I do. I am definitely too bad when it comes to serious programming.

### DIFF
--- a/symphony/lib/toolkit/class.widget.php
+++ b/symphony/lib/toolkit/class.widget.php
@@ -403,7 +403,7 @@ class Widget
                     $th->setAttribute('scope', $col[1]);
                 }
 
-                if (is_array($col[2]) && !empty($col[2])) {
+                if (!empty($col[2]) && is_array($col[2])) {
                     $th->setAttributeArray($col[2]);
                 }
 
@@ -710,7 +710,7 @@ class Widget
      */
     private static function __SelectBuildOption($option)
     {
-        @list($value, $selected, $desc, $class, $id, $attributes) = $option;
+        list($value, $selected, $desc, $class, $id, $attributes) = array_pad($option, 6, null);
 
         if (!$desc) {
             $desc = $value;
@@ -871,7 +871,7 @@ class Widget
             if ($time === true) {
                 $separator = DateTimeObj::getSetting('datetime_separator');
                 $time = DateTimeObj::convertTimeToMoment(DateTimeObj::getSetting('time_format'));
-        
+
                 $calendar->setAttribute('data-calendar', 'datetime');
                 $calendar->setAttribute('data-format', $date . $separator . $time);
             } else {
--- a/symphony/lib/toolkit/class.xsrf.php
+++ b/symphony/lib/toolkit/class.xsrf.php
@@ -23,7 +23,11 @@ class XSRF
      */
     public static function getSessionToken()
     {
-        $token = $_SESSION[__SYM_COOKIE_PREFIX__]['xsrf-token'];
+        $token = null;
+
+        if (isset($_SESSION[__SYM_COOKIE_PREFIX__]['xsrf-token'])) {
+            $token = $_SESSION[__SYM_COOKIE_PREFIX__]['xsrf-token'];
+        }
 
         if (is_array($token)) {
             $token = key($token);
--- a/symphony/content/content.blueprintssections.php
+++ b/symphony/content/content.blueprintssections.php
@@ -14,9 +14,8 @@ class contentBlueprintsSections extends AdministrationPage
 
     public function build(array $context = array())
     {
-        $section_id = $context[1];
-
-        if (isset($section_id)) {
+        if (isset($context[1])) {
+            $section_id = $context[1];
             $context['associations'] = array(
                 'parent' => SectionManager::fetchParentAssociations($section_id),
                 'child' => SectionManager::fetchChildAssociations($section_id)
--- a/symphony/lib/boot/func.utilities.php
+++ b/symphony/lib/boot/func.utilities.php
@@ -211,7 +211,7 @@ function cleanup_session_cookies()
 function is_session_empty()
 {
     $session_is_empty = true;
-    if (is_array($_SESSION)) {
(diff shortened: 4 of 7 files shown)


nitriques:
> In order to secure code quality, some Content Management Systems (at least Wordpress) provide a way to switch to stricter error reporting during development. 

I'd like that for 2.7.x

nitriques:
@michael-e are you able to do `git rebase -i` + `git push --force` on this ? If not, I'll do it. Thanks.

michael-e:
Sure, will do!

michael-e:
Done!

symphony/lib/toolkit/class.widget.php
```diff
@@ -403,7 +403,7 @@ public static function TableHead(array $columns = null)
                     $th->setAttribute('scope', $col[1]);
                 }
 
-                if (is_array($col[2]) && !empty($col[2])) {
+                if (!empty($col[2]) && is_array($col[2])) {
```
nitriques: Should we create method in the General for those uses cases ? `isset + empty` could also be needed.

symphony/lib/toolkit/class.widget.php
```diff
@@ -710,7 +710,7 @@ public static function Select($name, array $options = null, array $attributes =
      */
     private static function __SelectBuildOption($option)
     {
-        @list($value, $selected, $desc, $class, $id, $attributes) = $option;
+        list($value, $selected, $desc, $class, $id, $attributes) = array_pad($option, 6, null);
```
nitriques: It's ok to remove the silencer here ? What if the list call still fails somehow ?

symphony/lib/toolkit/class.xsrf.php
```diff
@@ -23,7 +23,11 @@ class XSRF
      */
     public static function getSessionToken()
     {
-        $token = $_SESSION[__SYM_COOKIE_PREFIX__]['xsrf-token'];
+        $token = null;
+
+        if (isset($_SESSION[__SYM_COOKIE_PREFIX__]['xsrf-token'])) {
+            $token = $_SESSION[__SYM_COOKIE_PREFIX__]['xsrf-token'];
+        }
```
nitriques: Is there really no other way to avoid the duplication (i.e. `$_SESSION[__SYM_COOKIE_PREFIX__]['xsrf-token']`) here ? We must have that pattern everywhere...

nitriques:
I've left a couple of comments, but it LGTM !

symphony/lib/toolkit/class.widget.php
michael-e: If we say that we want to promote good practices, we should not "hide" such patterns behind our own custom functions.

symphony/lib/toolkit/class.widget.php
michael-e: My guess was that the silencer was introduced because the array may be incomplete. But meanwhile this is not enough. (I may be wrong, as I say, it was just a guess.)

You know me. I prefer loud noise when things break — this means that they can be fixed.

symphony/lib/toolkit/class.xsrf.php
michael-e: Actually I hoped that some PHP guru knows a more elegant way to do it, without losing explicity. (As mentioned above, I wouldn't want to hide such patterns in custom functions.)

michael-e:
Feel free to improve my changes, they are meant as a starting point.

symphony/lib/toolkit/class.widget.php
nitriques: > You know me. I prefer loud noise when things break — this means that they can be fixed.

Sold!

symphony/lib/toolkit/class.widget.php
nitriques: `if (General::isNotAnEmptyArray($col[2]))` would hide what ? I feel like it's making the check explicit.

michael-e: `if (General::isNotAnEmptyArray($col[2]))` would hide the `is_array` part. (I would expect this function to return `true` for a `null` value as well, because `null` *is not an empty array*.) So we would have to call it `General::isNotEmptyAndIsAnArray($col[2])`. :-)

Anyway, using a function of the `General` class would only help experienced developers. Others wouldn't use them (and, again, write stuff like `is_array($col[2]) && !empty($col[2])`). (Take me, for example: Working with Symphony taught me everything I know about PHP coding.) If we want people to grow with Symphony, I suggest to find *the best way to deal with certain things* and use these solutions as visible (explicit) as possible. (Is there any "official recommendation" for certain patterns?)

nitriques: You make valid points ;)

> Is there any "official recommendation" for certain patterns?

I've yet to seen some especially for PHP. Maybe I should read more of the PSR-* stuff. I also watched a [Youtube video recently about legacy php apps](https://www.youtube.com/watch?v=jJ5PCbjZIsY) and the book seems interesting, but it is based on 3 books I've already read, which are not php related, but tells me I should create those methods, specially because they contains edge cases so they can be unit tested. But again, here some the naming problem.

.NET has a `String.isNullorEmpty` and `String.isNullOrWhitespace()` which is quite nice. Maybe `Array::isEmpty()` (or `General::isArrayNotEmpty()` would make it so that it's clear that it must be an array ?

PR #2676, opened 2017-05-17
Better fix for MySQL 5.7 Strict mode

nitriques:
58bc57f introduce a possible backward compatiblity change, since PHP 7
requires the child class to declare all optional paremeters, even if not
used. This would not be a breaking change 5.6.

Following was has been proposed in 58bc57f, this commit removes the new
`$select` parameter in favor of a new method, `buildSortingSelectSQL()`.

This new method is called in the `EntryManager`, just after the call to
`buildSortingSQL`. That way, it is possible for fields that requires to
add columns into the project to do it, while not changing a public API.

Note that this is still a temporary fix that would need to be addressed,
but when a major version will come out.

---

Should we call this method only when we detect MySQL 5.7 + ?


symphony/lib/toolkit/class.entrymanager.php
```diff
@@ -408,13 +412,15 @@ public static function fetch($entry_id = null, $section_id = null, $limit = null
         // Handle when the sort field is an actual Field
         } elseif (self::$_fetchSortField && $field = FieldManager::fetch(self::$_fetchSortField)) {
             if ($field->isSortable()) {
-                $field->buildSortingSQL($joins, $where, $sort, self::$_fetchSortDirection, $select);
+                $field->buildSortingSQL($joins, $where, $sort, self::$_fetchSortDirection);
+                $sortSelectClause = $field->buildSortingSelectSQL($sort, self::$_fetchSortDirection);
```
brendo: Minor, but what if was called `buildSortingSelectClause`, as the result is not actually SQL.

nitriques:
@nilshoerrmann I just wanted to let you know that the breaking of extension you reported won't occur ;)

symphony/lib/toolkit/class.entrymanager.php
nitriques: Well I've modeled the name according to the existing analogous method `buildSortingSQL`. We have to stick with the current nomenclature, even if it's not the best one IMHO.

nitriques:
@brendo @michael-e  Should we call this method only when we detect MySQL 5.7 + ?

michael-e:
I wouldn't add any extra logic. It's compatible with "non-strict mode" anyway.

Can you give me steps to reproduce the issue, please? As you know, I have a test setup, and I will manage to switch your fix on and off; but I simply don't know when the query in question occurs. Can I trigger the issue in the default workspace's backend?

nitriques:
> Can I trigger the issue in the default workspace's backend?

Yes! Make sure you have the latest workspace (with the missing unique keys) and change the sort order of the section.

> I wouldn't add any extra logic. It's compatible with "non-strict mode" anyway.

Perfect.

michael-e:
OK, will test tomorrow. I need some time, because I want to make some diffs on the codebase as well.

Will you add a commit with the renamed function as proposed by @brendo?

nitriques:
> Will you add a commit with the renamed function as proposed by @brendo?

Do you agree with his proposal ? I'd like to stick to the current nomenclature. It does produce SQL 'parts'

michael-e:
> There are only two hard things in Computer Science: cache invalidation and naming things.  
> -- Phil Karlton

That said, I am 60:40 for @brendo's naming. :-)

nitriques:
I like that quote so much. I always think of it when naming things in the cachelite code...

Damn what a hard decision...

nitriques:
@nilshoerrmann wanna chime in on the name issue ?

nilshoerrmann:
If the result is not SQL, it's best not to name it so, isn't it? So I agree with Brendan.

nitriques:
> If the result is not SQL,

Well, it it 'part' of the SQL query. It *is* SQL...

Last call: @brendo still voting for `buildSortingSelectClause` ?

brendo:
I realise my comment forgets about the other functions we have that don't generate SQL, but just a "part" of it. Happy with either approach to be honest, the end goal is good and I don't want a naming convention to get in the way of future progress!

nitriques:
Yeah maybe I like to argue about this too much. Let's stick with the current nomenclature. Thanks @brendo (and everybody else!)

PR #2678, opened 2017-05-19
Add requiresTable perf optimization

nitriques:
As suggested by @jonmifsud in #2664, this commit adds a new method in
the Field class, happily named requiresTable. Analogous to createTable,
this boolean flag tells Symphony to act on a table or not, reducing the
need to check for existing table when it is known to not exist.

---

BTW, while doing this, I saw this:

1. The `EntryManager::add()` method should also use LOCKS.
1. The `EntryManager::edit()` implementation and the `EntryManager::add()` implementation are ALMOST IDENTICAL and should be extracted.

I am willing to send a PR that addresses both concerns.

### DIFF
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -116,7 +116,13 @@ class EntryManager
                 continue;
             }
 
-            Symphony::Database()->delete('tbl_entries_data_' . $field_id, sprintf("
+            // Check if table exists
+            $table_name = 'tbl_entries_data_' . General::intval($field_id);
+            if (!$field->requiresTable() || !Symphony::Database()->tableExists($table_name)) {
+                continue;
+            }
+
+            Symphony::Database()->delete($table_name, sprintf("
                 `entry_id` = %d", $entry_id
             ));
 
@@ -141,7 +147,7 @@ class EntryManager
                 $fields[$ii] = array_merge($data, $fields[$ii]);
             }
 
-            Symphony::Database()->insert($fields, 'tbl_entries_data_' . $field_id);
+            Symphony::Database()->insert($fields, $table_name);
         }
 
         $entry->set('id', $entry_id);
@@ -185,7 +191,7 @@ class EntryManager
 
                 // Check if table exists
                 $table_name = 'tbl_entries_data_' . General::intval($field_id);
-                if (!Symphony::Database()->tableExists($table_name)) {
+                if (!$field->requiresTable() || !Symphony::Database()->tableExists($table_name)) {
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


michael-e:
@nitriques regarding the other topic:

> 1. The `EntryManager::add()` method should also use `LOCKS`.
> 2. The `EntryManager::edit()` implementation and the EntryManager::add() implementation are ALMOST IDENTICAL and should be extracted.

1. Why? What do you expect to happen?
2. Yep, why not.

Please open a new issue to discuss number 1. It's hard if several topics are mixed in a single thread.

symphony/lib/toolkit/class.entrymanager.php
```diff
@@ -116,7 +116,13 @@ public static function add(Entry $entry)
                 continue;
             }
 
-            Symphony::Database()->delete('tbl_entries_data_' . $field_id, sprintf("
+            // Check if table exists
+            $table_name = 'tbl_entries_data_' . General::intval($field_id);
+            if (!$field->requiresTable() || !Symphony::Database()->tableExists($table_name)) {
```
jonmifsud: @nitriques since you put or as a condition; wouldn't `requiresTable()` be completely useless if it returns false as it will always or with the table exists? Or is the idea to be 100% sure that there is no table to delete?

I'm fine with the PR just a question.

nitriques: Good catch @jonmifsud it MUST be &&!

nitriques: No, it's ok, it's a double negation... !false will short-circuit

nitriques:
There's a bug with it.. somehow, my dev setup did not picked up all the changes in the commit. https://github.com/symphonycms/symphony-2/pull/2678/commits/95f830a524108479fe965fa31c983aa4fa93469e#diff-57eff69d6ab1319955c5a954c701ca33R121 is an error since `$field` here is an array.

I think that @michael-e will agree with me that it's not worth fetching the field from the DB to call  perf method on it: we might simply check if the table exists, which should be fast enough.

michael-e:
Sorry, I don't get your point here. Do you mean that the commit doesn't work as intended? Or that you messed up something in Git?

> I think that @michael-e will agree with me that it's not worth fetching the field from the DB to call perf method on it: we might simply check if the table exists, which should be fast enough.

I expected this solution to just check the return value of the `requiresTable` function.

nitriques:
> Sorry, I don't get your point here. Do you mean that the commit doesn't work as intended? 

No it does not works as indented. At [this line](https://github.com/symphonycms/symphony-2/pull/2678/commits/95f830a524108479fe965fa31c983aa4fa93469e#diff-57eff69d6ab1319955c5a954c701ca33R121) `$field` is not an object, it is an array. So we can't call the function. Since we do not have any `Field` instance, we need to fetch it, which requires DB call.

michael-e:
Aha. I understand.

nitriques:
Do you also agree ? ;)

nitriques:
See 157b7167

michael-e:
Yes. And this time I have no better idea, and no objections either! :-)

nitriques:
Awesome. Sending a PR right now !

PR #2680, opened 2017-05-26
Refactor EntryManager::edit() and EntryManager::add() to make them identical in terms of field data saving

nitriques:
The title says it all ! + check the commits messages

157b716 also fixes a bug introduced in #2678 .

### DIFF
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -94,38 +94,50 @@ class EntryManager
     }
 
     /**
-     * Given an Entry object, iterate over all of the fields in that object
-     * an insert them into their relevant entry tables.
+     * Executes the SQL queries need to save a field's data for the specified
+     * entry id.
      *
-     * @param Entry $entry
-     *  An Entry object to insert into the database
-     * @throws DatabaseException
-     * @return boolean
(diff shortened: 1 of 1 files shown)


jonmifsud:
@nitriques any specific reason we need the lock table here?

Finding quite a few instances where mysql seems to be getting stuck on a lock for one of the sites we're running.

michael-e:
Locking tables when editing entries has solved https://github.com/symphonycms/symphony-2/issues/2472 — Symphony actually had a race condition without these locks.

jonmifsud:
@michael-e thanks for that. I had to disable it on a couple of sites it seems to have adverse effects in some scenarios. 

Entries where at times taking over 60s to save since a switch to cloudfront today; the client was complaining that they are getting 504s likely this has been the case for a long time and our timeout on varnish was more lenient. 

Since the change it saves in 2s or less. Other than that the lock was obviously also blocking frontend performance at times...

I'm guessing in symphony 3 this would be a transaction rather than a lock. A lock doesn't seem to make logical sense; in that my understanding is we are using it instead of a transaction to ensure it is successfully completed at one go correct?

michael-e:
Sounds bad. But honestly, I never had any issues with table locking.

> I'm guessing in symphony 3 this would be a transaction rather than a lock.

Hmmm, I think so, but I am not sure.

Issue #2618, opened 2016-08-13
General::checkFile() produces "false negatives"


michael-e:
When attemtping to update the JIT Image Manipulation extension to a recent version, I received the error: `.htaccess exists but is not writable.`. Looking into it, I noticed that the extension uses `General::checkFile()` to see if the `.htaccess` file is writable.

There we find:

``` php
if (
    (!is_writable($dir) || !is_readable($dir)) // Folder
    || (file_exists($file) && (!is_readable($file) || !is_writable($file))) // File
) {
    return false;
}
```

Now if a file already exists, it **will** be writable even if the directory is readable but not writable. The above, however, will return false because `(!is_writable($dir) || !is_readable($dir))` will be false.

In short: The logic _seems_ broken.

We might change this to only look into the directory if the file does not exist. Something like:

``` php
if (
    (!file_exists($file) && (!is_writable($dir) || !is_readable($dir))) // Folder
    || (file_exists($file) && (!is_readable($file) || !is_writable($file))) // File
) {
    return false;
}
```

But on the other hand the result of this might be surprising if you attempt to delete a file without write permissions on the directory.

So the main issue I see here is that the function's name is so weak and misleading. `checkFile` does not say what it does.

The current version checks if a file and its directory are readable and writable. The latter will not be needed in order to write to existing files. So you will have **false negatives**.

My proposal checks if the directory is readable and writable if the file does not exist (so you will be able to create the file if the function returns `true`). If the file exists, it will only check the file, not the directory, which is enough to know that you can write to the file — but you may not be able to delete it. So you will have **false positives**.

I don't say which one is better (see above) – I just say it is hard to decide and/or understand it because of the function's name.

Just thinking loud… On the long run, shouldn't we use more than one function checking a file? Maybe:
- General::checkFileRead()
- General::checkFileWrite()
- General::checkFileDelete()

michael-e:
To make this more clear (from the top of my head, I hope that I will not be proven wrong):

* Creating a file **requires** write permissions on the directory.
* Writing to an existing file **does not require** write permissions on the directory.
* Deleting a file **requires** write permissions on the directory.

designermonkey:
Those assertions are correct.

What is needed to counter the file permissions errors, read and write errors and other issues relating to the filesystem that have dogged Symphony over the years is proper abstraction of the filesystem: Proper functions that do what they say.

[...]

PR #2683, opened 2017-05-30
Add General::checkFile*able() API

nitriques:
Re #2618

See commit messages and comments.


nitriques:
@brendo wanna check the last 4 commits please ?

PR #2685, opened 2017-05-30
Move General::buildPaginationElement >> DataSource 

### LINKED ISSUE #2670: General::buildPaginationElement() should be moved elsewhere
It does not belong in this class...

nitriques:
See commit details.

Fixes #2670

(Do not merge for now)


symphony/lib/toolkit/data-sources/class.datasource.section.php
```diff
@@ -581,7 +581,9 @@ public function execute(array &$param_pool = null)
             $result->prependChild($sectioninfo);
 
             if ($include_pagination_element) {
-                $pagination_element = General::buildPaginationElement();
+                $pagination_element = $this->buildPaginationElement(array(
```
brendo: Use `Datasource::` to access the method (because it's static).

symphony/lib/toolkit/class.general.php
```diff
@@ -1494,6 +1494,7 @@ public static function createXMLDateObject($timestamp, $element = 'date', $date_
     /**
      * Construct an XML fragment that describes a pagination structure.
      *
+     * @deprecated @since Symohony 2.7.0. Use DataSource::buildPaginationElement()
```
brendo: Two typos, Symohony (Symphony) and DataSource (Datasource)

symphony/lib/toolkit/class.datasource.php
```diff
@@ -241,6 +241,45 @@ public function emptyXMLSet(XMLElement $xml = null)
...
+     *  to 0.
+     * @param integer $infos.$entries-per-page (optional)
+     *  the number of entries per page. this defaults to 1.
+     * @param integer $infos.$current-page (optional)
+     *  the current page within the total number of pages within this pagination
+     *  structure. this defaults to 1.
+     * @return XMLElement
+     *  the constructed XML fragment.
+     */
+    public static function buildPaginationElement(array $infos = array())
```
brendo: I feel the old signature was cleaner, what prompted the change?

symphony/lib/toolkit/class.datasource.php
```diff
@@ -241,6 +241,45 @@ public function emptyXMLSet(XMLElement $xml = null)
     }
 
     /**
+     * Construct an XML fragment that describes a pagination structure.
+     * Do not call as a static function, since this is only temporary.
```
brendo: ?

symphony/lib/toolkit/data-sources/class.datasource.section.php
nitriques: Indeed.

symphony/lib/toolkit/class.datasource.php
nitriques: It SHOULD be static, but it can't right now (it was crashing on my end...) Maybe I need to revisit this.

symphony/lib/toolkit/class.datasource.php
nitriques: In order to make function signature resilient to changes, its a good practice to try to replace parameter with defaults values with a "settings bags".

You can then

1. Specify the 3rd option only, and keep the default intact.
2. Add a new option without worrying about signatures/API compat
3. Extension could ship there could with new option and it should not break on old install!

It only needs to be documented

Those parameters was the perfect example of this code smell and I did wanted to try it out.

nitriques: Most of the Widget would benefit from this... $value, $attributes.

brendo: That is a really interesting perspective, thanks for sharing!

symphony/lib/toolkit/class.general.php
```diff
@@ -1508,16 +1509,14 @@ public static function createXMLDateObject($timestamp, $element = 'date', $date_
      * @return XMLElement
      *  the constructed XML fragment.
      */
-    public static function buildPaginationElement($total_entries = 0, $total_pages = 0, $entries_per_page = 1, $current_page = 1)
+    public static function buildPaginationElement($total_entries = 0, $total_pages = 0, $entries_per_page = null, $current_page = 1)
```
michael-e: Why `null`? Shouldn't it be `0` (because we assign integer values)?

michael-e (dismissed):

symphony/lib/toolkit/class.datasource.php
nitriques: The same pattern exists in javascript too, where you have an "configuration object" that merges with defaults ;) Like most jQuery plugins

PR #2686, opened 2017-05-30
Authors delegates

nitriques:
This has sitting in branches since November 2015... So this PR is long overdue! It's been rebased countless time, and now it needs to be released!

It adds a bunch of delegates related to the Authors and offer new delegates for extensions to be able to adapt the experience on a user basis.

- AddCustomAuthorColumn, which allows developers to add columns into the index page
- AuthorPreCreate, just before author creation. The creation can be forbidden.
- AuthorPostDelete, just after deletion. This allows extension do to some clean up
- NavigationPostBuild, which allows developers to edit the nav
- CanAccessPage, which controls if the author can see the page. Also used to check for links in the nav.

### DIFF
--- a/symphony/content/content.systemauthors.php
+++ b/symphony/content/content.systemauthors.php
@@ -696,7 +696,28 @@ class contentSystemAuthors extends AdministrationPage
 
             $this->_Author->set('auth_token_active', ($fields['auth_token_active'] ? $fields['auth_token_active'] : 'no'));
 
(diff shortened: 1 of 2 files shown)


nitriques:
Thanks @michael-e no worries. It's been tested a lot, but before the rebase. The operation went quite well, so I hope my testing is still valid.

brendo:
Nice work, Authors was always a neglected part of the system.

PR #2687, opened 2017-05-30
Add the possibility to run with E_ALL

### LINKED ISSUE #2674: Make error_reporting configurable
As proposed in #2673

Should it only be a "script_php" bool ?

nitriques:
This will make it easier to have strict setups

Fixes #2674

### DIFF
--- a/install/includes/config_default.php
+++ b/install/includes/config_default.php
@@ -22,6 +22,7 @@
             'session_gc_divisor' => '10',
             'cell_truncation_length' => '75',
             'enable_xsrf' => 'yes',
+            'error_reporting_all' => 'no',
         ),
         ########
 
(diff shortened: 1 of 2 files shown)


nitriques:
> Now all we need is someone who wants to fix all the quirks by actually enabling this! :-)

I hope some people won't be able to resist in doing so ;)

brendo:
Instead of a bool, what if the config was the actual error reporting value, eg `E_ALL`?

nitriques:
@brendo see https://github.com/symphonycms/symphony-2/issues/2674#issuecomment-303660480

I agreed that nothing more was needed.

nitriques:
The problem with it is that you'll loose the PHP constant when the config.php file is written. But yeah, feel free to send a PR :)

PR #2688, opened 2017-05-30
Add X-Content-Type-Options and X-XSS-Protection

### LINKED ISSUE #2248: X-Frame-Options is deprecated
As per https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options

> Note: The frame-ancestors directive from the CSP Level 2 specification officially replaces this non-standard header. CSP's frame-ancestors has been supported since Gecko 4.0 but is not supported across all browsers. Though X-Frame-Options is not an official standard it is widely supported and can be used in conjunction with CSP.

So I think it should be wise to add the new values too.

https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives

nitriques:
Those headers should help us fight some wrong doing, in browsers that
support them.

Fixes #2248

### DIFF
--- a/symphony/lib/toolkit/class.administrationpage.php
+++ b/symphony/lib/toolkit/class.administrationpage.php
@@ -441,6 +441,14 @@ class AdministrationPage extends HTMLPage
             $this->addHeaderToPage('X-Frame-Options', 'SAMEORIGIN');
         }
 
+        if (!array_key_exists('x-content-type-options', $this->headers())) {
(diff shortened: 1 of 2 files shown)


nitriques:
@brendo Please use the 'rebase' mode when merging, since we will rebase anyways, at least it adds the # of the PR in the commit title ! And linear history is so much easier to read !

PR #2689, opened 2017-05-30
Eliminate code duplication in random order check

nitriques:
This eliminates some code duplication.
Also nice for extensions developers

### DIFF
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1538,7 +1538,7 @@ class Field
     }
 
     /**
-     * Determine if the requested $order is random.
+     * Determine if the requested $order is random or not.
      *
      * @since Symphony 2.7.0
      * @param string $order
@@ -1546,7 +1546,7 @@ class Field
      * @return boolean
      *  true if the $order is either 'rand' or 'random'
      */
-    public static function isRandomOrder($order)
+    protected function isRandomOrder($order)
     {
         return in_array(strtolower($order), array('random', 'rand'));
     }
@@ -1570,7 +1570,7 @@ class Field
      */
     public function buildSortingSQL(&$joins, &$where, &$sort, $order = 'ASC')
     {
-        if (static::isRandomOrder($order)) {
+        if ($this->isRandomOrder($order)) {
             $sort = 'ORDER BY RAND()';
         } else {
             $joins .= "LEFT OUTER JOIN `tbl_entries_data_".$this->get('id')."` AS `ed` ON (`e`.`id` = `ed`.`entry_id`) ";
@@ -1600,7 +1600,7 @@ class Field
      */
     public function buildSortingSelectSQL($sort, $order = 'ASC')
     {
-        if (static::isRandomOrder($order)) {
+        if ($this->isRandomOrder($order)) {
             return null;
         }
         return '`ed`.`value`';
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1538,10 +1538,25 @@ class Field
     }
 
     /**
+     * Determine if the requested $order is random.
+     *
+     * @since Symphony 2.7.0
+     * @param string $order
+     *  the sorting direction.
+     * @return boolean
+     *  true if the $order is either 'rand' or 'random'
+     */
+    public static function isRandomOrder($order)
+    {
+        return in_array(strtolower($order), array('random', 'rand'));
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


symphony/lib/toolkit/class.field.php
```diff
@@ -1538,10 +1538,25 @@ public function buildDSRetrievalSQL($data, &$joins, &$where, $andOperation = fal
...
     /**
+     * Determine if the requested $order is random.
+     *
+     * @since Symphony 2.7.0
+     * @param string $order
+     *  the sorting direction.
+     * @return boolean
+     *  true if the $order is either 'rand' or 'random'
+     */
+    public static function isRandomOrder($order)
```
froschdesign: Why static and public?

* All field types extends the class `Field` and so there is no need for a static or public method.
* Maybe I am wrong, but an extension developer also extends the class `Field`. This is another reason against static and public.

nitriques: static: because it does not uses `$this` and it does not have a state. Why would it need to be an instance method ? I've yet to find a good reasoning on this, any help is appreciated!

public: to be able to unit test it, one day...

froschdesign: > static: because it does not uses $this

You use this method only in non-static methods and inside the same object, therefore it make no sense to add `static`. (Yes, all method calls must be changed.)

> Why would it need to be an instance method ?

You have to ask the question differently: Why static? A static variable or method is always an exception. Static methods are nothing more than namespaced global functions.
The `Field` class is not stateless, why should this method be outside them?

> public: to be able to unit test it, one day...

Also no need for this, because you test (e.g.) `Field::buildSortingSQL()` and there you must test all parameters - included `random` and `rand`.
(You should not test protected or private members directly, but you can do this with PHP's [`ReflectionClass`](http://php.net/manual/en/class.reflectionclass.php).)

nitriques: > You should not test protected or private members directly

I must disagree on that! That's the point of unit tests!

> Static methods are nothing more than namespaced global functions.

Agreed. That's how I would classify that function.

> The Field class is not stateless, why should this method be outside them?

The method is. If `Field` had a `$order` property, it would make sense to make it an instance method. And with late static binding, it can even be overridden by child classes.

The only positive thing of making it protected and instance would be that later on, it could be used with DI. I think that's one of the valid argument. 

Please @froschdesign continue to argue with me!!

@brendo @michael-e What are your thoughts on this ? I know we've had the static or not debate often, but since options can change, I am asking for yours once again!

froschdesign:
@nitriques 
> I must disagree on that! That's the point of unit tests!

Unit testing means: assert the behavior of your public interface. If your internal (private or protected) method is not covered by tests, then you can delete this method, because it is not used. 

([Sebastian Bergmann: "Test Your Privates"](https://web.archive.org/web/20150425132055/https://sebastian-bergmann.de/archives/881-Testing-Your-Privates.html) – last quotation and paragraph)

> That's how I would classify that function.

There is no reason for adding the method as "global helper function", if you use this method only in the _same object_!

> If Field had a $order property, it would make sense to make it an instance method.

…and you open the doors for hidden dependencies and make testing very difficult.

> And with late static binding, it can even be overridden by child classes.

You can override every protected method.

> Please @froschdesign continue to argue with me!!

All is fine! 😃

nitriques:
> There is no reason for adding the method as "global helper function", if you use this method only in the same object!

So in that sense, it could still be protected and static, no ?

> …and you open the doors for hidden dependencies and make testing very difficult.

Agreed! I very much like function purity, which is hard to do in OOP language.

Again, I'd love to ear what @michael-e and @brendo thinks about this design. And thanks @froschdesign I'll sleep on this ;)

michael-e:
I am afraid that I can't say anything useful about this, due to a lack of knowledge.

brendo:
I tend to agree that the method should not be static. 

Many of Symphony's classes suffer from having multiple concerns and being called from a variety of contexts so we don't have any consistency between what methods are static/instance.

nitriques:
@michael-e @brendo Thanks for the input. I'll make it protected and instance then. Thanks @froschdesign !

nitriques:
a745e09 @froschdesign is the requested change. (BTW, feel free to check at other issues as well!)
@brendo We are good to go ?

nitriques:
I just stumple upon why I made it [public static](https://github.com/symphonycms/symphony-2/pull/2689/files#diff-3c741ba68ea8b6d2209ccd796286ebe4R1329)

PR #2695, opened 2017-06-01
Fix not-regexp which leads to wrong results

nitriques:
When filtering with not-regexp, it is required to do AND because we need
all columns to not match: having only one not matching is not enough.

Also, the if checked assumed that if the `regexp:` pattern was not
found, it would default to not-regexp. Make sure that we check both.


nitriques:
Nice spot @michael-e

PR #2696, opened 2017-06-01
Fix broken filter after a sort

### LINKED ISSUE #2465: Sort column while filter is active
Sorting a column while a filter is active is removing the last character of the filter from the URL after the sort is applied.

nitriques:
Sorting a column while a filter is active is removing the last character
of the filter from the URL after the sort is applied, if and only if, it
ends with one of those for chars: &amp;

I do not understand why both 7328c19a5 and 4631a4371 used
`trim($params['filters'], '&amp;');` to get rid of the leading '&amp;'
in the filter string.

This commits merges both mention commits into the same line, and
properly removes the leading &amp; chars.

Fixes #2465

### DIFF
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -17,6 +17,12 @@ class contentPublish extends AdministrationPage
     public function sort(&$sort, &$order, $params)
     {
         $section = $params['current-section'];
+        $filters = '';
+        // Format the filter query string
+        if (isset($params['filters']) && !empty($params['filters'])) {
+            $filters = preg_replace('/^&amp;/i', '', $params['filters'], 1);
+            $filters = '?' . trim($filters);
+        }
 
         // If `?unsort` is appended to the URL, then sorting is reverted
         // to 'none', aka. by 'entry-id'.
@@ -24,11 +30,11 @@ class contentPublish extends AdministrationPage
             $section->setSortingField('id', false);
             $section->setSortingOrder('desc');
 
-            redirect(Administration::instance()->getCurrentPageURL());
+            redirect(Administration::instance()->getCurrentPageURL() . $filters);
         }
 
         // By default, sorting information are retrieved from
-        // the filesystem and stored inside the `Configuration` object
+        // the file system and stored inside the `Configuration` object
         if (is_null($sort) && is_null($order)) {
             $sort = $section->getSortingField();
             $order = $section->getSortingOrder();
@@ -37,18 +43,11 @@ class contentPublish extends AdministrationPage
             EntryManager::setFetchSorting($sort, $order);
         } else {
             $sort = General::sanitize($sort);
-            $filters = '';
 
             // Ensure that this field is infact sortable, otherwise
             // fallback to IDs
             if (($field = FieldManager::fetch($sort)) instanceof Field && !$field->isSortable()) {
                 $sort = $section->getDefaultSortingField();
-            }
-
-            // Format the filter query string
-            if ($params['filters']) {
-                $filters = preg_replace('/^&amp;/i', '', $params['filters'], 1);
-                $filters = '?' . trim($filters);
             }
 
             // If the sort order or direction differs from what is saved,
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -37,6 +37,7 @@ class contentPublish extends AdministrationPage
             EntryManager::setFetchSorting($sort, $order);
         } else {
             $sort = General::sanitize($sort);
+            $filters = '';
 
             // Ensure that this field is infact sortable, otherwise
             // fallback to IDs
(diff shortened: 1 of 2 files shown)


symphony/content/content.publish.php
```diff
@@ -37,33 +37,31 @@ public function sort(&$sort, &$order, $params)
...
 
             // Ensure that this field is infact sortable, otherwise
             // fallback to IDs
             if (($field = FieldManager::fetch($sort)) instanceof Field && !$field->isSortable()) {
                 $sort = $section->getDefaultSortingField();
             }
 
+            // Format the filter query string
+            if ($params['filters']) {
+                $filters = preg_replace('/^&amp;/i', '', $params['filters'], 1);
```
michael-e: Wouldn't it be simpler, faster and more bulletproof to use `ltrim($params['filters'], '&amp;')`?

nitriques: > Wouldn't it be simpler

it's trim vs preg_replace....

> faster

not for a one time call

> bulletproof

Totally not, since, like stated in the php's doc, it trims all '&' OR 'a' OR 'm' OR 'p' OR ';', not the characters in the given order...

So a value of '&amp;aa' would still comes out empty.

michael-e: Oh. I am sorry.

nitriques: No problem ;) Sometimes php treat string like array, which can lead to bad things.

nitriques:
@brendo @michael-e Could you approve the PR please ? No rush tho, take your time. Thanks.

michael-e:
PS: Without filter, the up/down arrows work although the sort order is not in the URL. So it does not need to be in the URL. Still, the arrows do not work with filters applied.

nitriques:
> Should I open a new issue?

Yes please! Filters and sort order have been using redirects since a long time.
I'll `git bisect` to check this out. Thanks.

nitriques:
> The redirect URL contains the filter, but not the sort order

I've just tested it and it works. `?sort=217&order=asc&filter[checkbox]=No` redirects to `?filter[checkbox]=No` with the proper order and arrow direction. Both with and without this PR.

michael-e:
I tested again, and it seems that the bug even exists in Symphony 2.6.11. So it has nothing to do with this PR.

I will try to describe the bug properly in a new issue.

nitriques:
@michael-e Can you test this branch ? both issue should be fixed now.

michael-e:
Will do. I am just moving my test server, please wait a bit. :-)

nilshoerrmann:
The live test server? ;)

michael-e:
No, I did that last weekend. Today it's only the "private sandbox" server. :-)

nitriques:
@michael-e Great. If you test are good, please approve the PR !

michael-e:
Hmm, regarding the arrows: I have to click two times to actually switch the sorting! :-(

michael-e:
Ah, please wait. I think I know what's going on.

michael-e:
Fixed that. Once again, an Opcache issue. (I had to add `opcache.revalidate_freq=0` to the PHP ini settings.)

nitriques:
Thank god (whoever is that god person anyway). Good to merge ?

michael-e:
How can one re-build Symphony's compressed assets? I always forget.

nitriques:
https://github.com/symphonycms/symphony-2/wiki/Building-Symphony's-Assets

nitriques:
I try not to included them in PRs since they will cause merge conflicts. But I should rebuild them when it's merged.

michael-e:
BTW, these Opcache issues will become more and more annoying, because many providers are using opcache now, leaving the revalidate frequency at the default (2 seconds). This means issues with everything that must be available at once and is saved in PHP files (like the sorting, saved in the config file).

nitriques:
Should we included a "flush" in the `General` class ?

michael-e:
You can't flush the Opcache, AFAIK. Or can you?

michael-e:
LGTM, merge it.

nitriques:
> You can't flush the Opcache, AFAIK. Or can you?

I do not know. If you ever feel like that is something Symphony should do, please open a ticket regarding Opcache ;) Thanks!

michael-e:
I simply think that we should keep this in mind for the next big Symphony thingie: Don't use PHP files for configuration, don't read immediately after writing.

(Email Template Manager, for example, has this issue as well: Saving a template may not be reflected in the GUI immediately because of Opcache because ETM write a PHP file with the settings.)

nitriques:
Indeed. [This post](https://ma.ttias.be/how-to-clear-php-opcache/) is worth reading I think.

michael-e:
Great post. So my solution — setting `opcache.revalidate_freq=0` — sounds like the best way to deal with our "volatile PHP files". Unfortunately you can't set this value in `.htaccess` files, so people in hosting environments can only ask the provider.

PR #2697, opened 2017-06-01
Fix the wrong count of associated entries

### LINKED ISSUE #2455: Association Count and Select Fields
Select field always display the assocation count in the section index as `0` (none) or `1` (some). They don't show the actual entry count.

nitriques:
The DISTINCT keywords was always returning 0 or 1.
Removing it makes the count working ok.

b10c4f67284da6eb3cb3a22a8598b44068fbb336 introduced the DISTINCT
keyword, but loads of other things, so it is hard to tell if something
made it not work.

Fixes #2455

### DIFF
--- a/symphony/lib/toolkit/fields/field.taglist.php
+++ b/symphony/lib/toolkit/fields/field.taglist.php
@@ -76,20 +76,14 @@ class FieldTagList extends Field implements ExportableField, ImportableField
 
     public function fetchAssociatedEntryCount($value)
     {
-        if (function_exists('cleanValue') === false) {
-            function cleanValue($val)
-            {
-                return Symphony::Database()->cleanValue($val);
-            }
-        }
-
-        $value = array_map("cleanValue", explode(',', $value));
+        $value = array_map(array($this, 'cleanValue'), explode(',', $value));
+        $value = implode("','", $value);
         $count = (int)Symphony::Database()->fetchVar('count', 0, sprintf("
-            SELECT COUNT(DISTINCT handle) AS `count`
+            SELECT COUNT(handle) AS `count`
             FROM `tbl_entries_data_%d`
             WHERE `handle` IN ('%s')",
             $this->get('id'),
-            implode("','", $value)
+            $value
         ));
 
         return $count;


michael-e:
My basic questions are: Should I __associate__ a section (using SBL field) with a Select Box or Tag List field in another section? Or should I simply __add__ a Select Box or Tag List field to a section? Which "associated count" should be displayed in which section (by clicking where in the section editor)?

nitriques:
I simply add a Select Box or Tag List field to a section WITH "associated count" checked, and go see the index page of the linked section.

I was always getting 0 if there was no link to that entry and always 1 if there was one **or more** links to that entry.

nitriques:
NOTE: Putting the `DISTINCT` outside the count also fixes the issue.

michael-e:
> I simply add a Select Box or Tag List field to a section WITH "associated count" checked, and go see the index page of the linked section.

I don't get it.

1. OK, I have a single section. Let's say I have an Input field there.
2. Now I add a Select Box or Tag List field to this section, right?
3. Where should I check "display count of entries in associated sections"? It's not there.
4. What do you mean by "linked section"?

nitriques:
1. OK, I have a single section (Call it `Input`). Let's say I have an Input field there. GOOD.
1. Now GO CREATE A NEW section (Call it `Link`), with a Select Box or Tag List, using data from `Input` as the "dynamic" values. Check yes to show association counts in the `Input` section.

> Where should I check "display count of entries in associated sections"? It's not there.

When creating the tag or select box field in `Link`, i.e.

image

> What do you mean by "linked section"?

Sorry, the proper word is "Suggestion list" for tags, "Dynamic values" for the select box. Those create a "link" to another section (To `input` for our case).

---

Create multiple entries in `Input`, then go create entries in `Link`, creating many `Link` entries that all points to the same `Input` entry. The count will only be one !

michael-e:
Thanks for taking the time to explain this. I will test tomorrow!

nitriques:
@michael-e no worries. Explaining it makes me double checks everything, which is good ;)

Issue #2650, opened 2017-03-06
Button "Create New" in entries' associations collides with title


mprofitl:
The sidebar `associations` show a title of each associated section and right beneath a button `Create New`. The sidebar is not wide enough for showing them beneath each other. I suggest to change the css of `symphony.associations.css` to 
```
.association-new {
   margin-top: 1rem;
   margin-left:0;
}
```
Then the button will be shown underneath the title without collision.

twiro:
> _The sidebar is not wide enough for showing them beneath each other._

That surely depends on the length oh your section-names ;)

> _I suggest to change the css of symphony.associations.css ... Then the button will be shown underneath the title without collision._

I don't think this will be a more satisfying solution than what we currently have. If you keep your section-names rather short, but have a lot of association-stuff going on in the drawer (and I often have lots of) you're glad that the "create new"-buttons don't take away any extra vertical space.

On top of that putting the buttons below the section-names doesn't add to the clearness/readability of the drawer.vI've attached a quick comparison-screenshot with a rather full associations-drawer to show what I mean.

One possible solution might be to simply shorte the section-titles if they're longer than a maximum of allowed characters. Or giving the drawer itself a little more width... or best combining both these methods. At least the Buttons should be placed above the section-title (z-index-wise)...

![symphony-associations-drawer](https://cloud.githubusercontent.com/assets/870227/23671917/45df1bc0-036d-11e7-8fac-b585e085237a.jpg)

mprofitl:
Your screenshots help a lot – I am adding one now.

Once a association exists, the collision rarely happens. However when no association exists yet, the text (at least with the backend in German langauge) is too long for text and button, e.g. _Keine Verknüpfungen für Neuigkeiten_ and _Neu erstellen_.

The increased z-index will make the button usable no mater how long the text. That would be a good amendment. 

The button on the right is indeed visually much clearer. The length of the section names don't seem to be troublesome, comparing yours and mine, it seems to be the translation. On the image attached I tried alternatives for the naming:

- _Anlegen_, _Erstellen_ or _Neu_ instead of _Neu erstellen_
- No differentation between the stated links existant or not: in both cases just _Verknüpfungen für…_ or even _Links für…_

Shortening the section titles would not help a lot as the space left to the button is already used up almost by the label.

![symphony associations in drawer](https://cloud.githubusercontent.com/assets/6146601/23692038/8fd02d00-03cc-11e7-9872-1ff7a29cce2b.png)

[...]

PR #2698, opened 2017-06-02
Associations - Fix Linked To informations

alexnantel88:
Add Title Linked To and Links in to allow more horizontal space for long names.
Change button from Create New to +.


brendo:
Not fond of the change to `+`, that convention does not exist anywhere in Symphony.

symphony/content/content.publish.php
```diff
@@ -1727,14 +1732,14 @@ private function prepareAssociationsDrawer($section)
                     ));
 
                     // Create new entries
-                    $create = new XMLElement('a', __('Create New'), array(
+                    $create = new XMLElement('a', '+', array(
```
nitriques: As @brendo said, can you please keep the 'Create New" label ? Thanks.

nitriques:
Re #2650

nitriques:
@alexnantel88 Can you send a screenshot of the new interface for @twiro to review quickly ?

Thanks!

alexnantel88:
image

@twiro

nitriques:
Great! And what happens if you make the button with a longer text ? And if the section's name is really really long ?

alexnantel88:
image
It would make an automatic ellipsis so the string stays on one line only.

nitriques:
Great!

nitriques:
@brendo @twiro last chance to review/comment...

twiro:
Great that this has been taken care of - I'm pretty busy at the moment and probably wouldn't have been able to send a PR soon enough for 2.7 (which seems to come along very well)! Thanks @alexnantel88 !

A few things though:

**1. The backward-arrow**

I'm not really a fan of using a backward-arrow to indicate links in the "linked in"-section. I understand the idea (The linked entry is associated to the current open entry, so from a "link-direction-point-of-view" it makes sense) but by clicking on that link we're still moving **forward** (in a general website/browser kinda sense) to that entry and not **back** (which that arrow somehow indicates to me). I totally can live with that if its the general consensus (which it seems to be) but in my eyes this doesn't improve things here - especially as this kind of logic isn't used anywhere else in the admin area.

**2. Association count**

I really think showing the number of associated entries for each section (as shown in my mockup) would be a great addition to the drawer... to hard to implement, hasn't this been thought of yet or are there good reasons against it?

**3. Font formatting & visual hierarchy**

In the mockup I built a few weeks ago I made the section-titles bold and the two main titles "links in" and "linked to" a little larger (14x) but in normal weight. In my eyes this improves the visual hierarchy of the drawer quite a lot as it is way easier to skim through the section-titles which - I think - are the more essential and useful data/information compared to the two main titles. These should be prominent enough to clearly separate the two types of associations, but then their job is done and they should move out of the way and don't look like the most important piece of information in the drawer - which currently is the case as they're the biggest **and** boldest piece of text in the drawer.

I also still think there's some space wasted in the listed links - reducing the top padding of each single listed link so that it matches the bottom padding would look cleaner, tighter and gain some valuable vertical space.

**4. Button wording**

As someone who uses the drawer a lot and is familiar with how it works I really would have prefered the simple, clean and less cluttered looking "+"-version instead of returning to the old  "Create New"-wording which in my eyes simply gives the button too much "weight" compared to the other data in the drawer. But I also understand @brendo's concerns regarding keeping up with Symphony's conventions. What about @nilshoerrmann's proposal of using a reduced wording ("New" or "Neu" in german)...? What does the thread opener think about it - @mprofitl?

----

My latest mockup from early may...

2ac1276c-2f22-11e7-8511-fea5506b1e53-1

nitriques:
1. Are there any arrows anywhere in the admin ? If not, let's removed them since they take up precious space
1. Yup, it should be doable [with this function](https://github.com/symphonycms/symphony-2/blob/3da2f53376b911cdf033696373aeef6f1131411f/symphony/lib/toolkit/class.field.php#L1828)
1. Yup good point
1. Agreed for the shorten version "New"

cc @alexnantel88

alexnantel88:
image

How about this @twiro @nitriques ?

nitriques:
@alexnantel88 

We are missing that point:

1. Yup, it should be doable [with this function](https://github.com/symphonycms/symphony-2/blob/3da2f53376b911cdf033696373aeef6f1131411f/symphony/lib/toolkit/class.field.php#L1828)

Also, I think we should add the `...` at the end of both title, as proposed by @twiro

alexnantel88:
image

@nitriques I put ':' instead of '...' at the end of the titles. Otherwise it looks like the strings are incomplete (which is accentuated by the fact that we use ellipsis for other long strings).

nitriques:
>  I put ':' instead of '...' at the end of the titles.

You're right. LGTM. Let me rebase it for fun

nitriques:
Merged as 4e3b973c0b...2da24023b0

mprofitl:
I'm late, the PR is closed, but may I comment on a small detail? I am happy to see how my little request has evolved and the solution is very neat in my eyes. Twiro's design did not underline the number of links in brackets; that looked better than it is implemented now. My other remarque is on the button's name in German (I'll open an issue there): "Create new" is meant as "create new link" but the translation "Neu erstellen" means "create newly".

Thanks @twiro for asking for my opinion. I very much like your solution.

nitriques:
@mprofitl It is never too late, we can always open new PR :) Never be afraid to share your ideas!

I understand that this may be a translation problem, since 'New' can have multiple different meanings. We have a length problem in French (new = nouveau) and the right word would be create (créer) but that felt weird in English.

As for the underline part I would be curious to see what's @twiro and @alexnantel88 opinion on this (or @nilshoerrmann's or anybody else's!)

nitriques:
> Twiro's design did not underline the number of links in brackets; that looked better than it is implemented now.

Finally, we went with usability: the count in clickable in the link (hence it's underlined)

twiro:
I'm a little late too and I'm really sorry... but I have to be a little nit-picky again:

---

> _Twiro's design did not underline the number of links in brackets; that looked better than it is implemented now._

Visually I also prefer my proposal without the underlined link-counter...

> _Finally, we went with usability: the count in clickable in the link (hence it's underlined)_

...but it's pretty obvious that it makes sense to consider the counter as part of the link and therefore underline it too. (Though I assume that it would not generate any "real world usability problems" to not do so and simply go with the smoother look ;) )

---

Regarding the styling of the two grouping associations-headlines ("Linked to", "Linked in") I'm still not happy with one detail - I know I talked about vertical space beeing scarce & valuable in the drawer, but in my eyes the margin below those headlines is too small to visually support the idea/meaning of these headlines:

They are meant to entitle and group multiple sections listed below them but the gap between the headline and the first "child"-section is noticable smaller (27px) than the gap between the first "child"-section and the second (or third, fourth, ...) "child"-section (32px or even 44px if one doesn't take the thin and light grey line into account).

I'd suggest to move from 27px to at least 32px or better 40px to improve the visual hierarchy.

symphony-associations-drawer-6

---

_Are there any arrows anywhere in the admin ? If not, let's removed them since they take up precious space_

I have to admit that I'm not happy with this change. We **do** use these arrows in the admin area (linked sections/entries in the section index, next/previous-buttons in the pagination) and in my eyes they did a pretty good job here in the drawer:

- They clearly marked the single entries as clickable **links**. Which is helpful because the entries in the drawer aren't underlined (at least not directly) as nearly every other linked content in the admin area).
- They supported the visual hierarchy in very clearly marking the entries as **listed children** of the **parent section** that precedes them (Think of a classical tree/folder-structure). This "listed look" and the indentation that happens through the arrows also helps the section names pop out more which again makes it easier to quickly scan the whole drawer.

So (at least in my eyes) the whole drawer looks much better organized with the arrows.

If horizontal space in the drawer really is a problem (for me it isn't as I guess 95% of my entries across projects fit in the available space and I really don't mind the other 5% getting shortened - but that surely differs from project to project and with different naming conventions) I would suggest to simply make the drawer a little wider - it has a fixed width of 300px right now and I would keep this as a minimum but allow for percent-based growing on larger screens. 350px already makes a big difference and I think other scenarios (documenter!) would benefit from a bigger drawer even more.

**tl;dr**

I don't wanna see all arrows disappear from the drawer just because [I didn't completely agree](https://github.com/symphonycms/symphony-2/pull/2698#issuecomment-307122360) with the backward-arrows. I'd really love to keep the arrows the way they were since the drawer was implemented, I'd be ok with the arrows pointing in both directions and indicating the "linked direction" if that's the general consenus but I think completely removing them is the worst option out of the 3.

---

Btw: Playing around with RC1 i noticed the [same problems](https://github.com/symphonycms/symphony-2/issues/2713) as @michael-e regarding the associations drawer... so as it seems to need some fixes anyway I'd be glad if my last suggestions for styling improvements could be considered for 2.7 :) Thanks for all your work guys!

twiro:
Oh - and one last thing:

If there is no linked entry in a section I'd suggest to simply use the wording "**_No links in [section]_**" and ditch the counter "**_(0)_**". Looks redundant and irritating in that context.

screen shot 2017-06-19 at 16 09 31

nilshoerrmann:
I agree with @twiro on the spacing issues and I also agree on the redundant zero counter.

> Are there any arrows anywhere in the admin ? If not, let's removed them since they take up precious space

Arrows are part of all associative UI elements. They are used to indicate relationships. Removing them removes this semantic. 

In my eyes, visual standards have to be considered when thinking about versioning: this is a breaking design change which removes a standard UI element. This has consequences for all association interfaces – something that should not be done in a minor release.

> I understand that this may be a translation problem, since 'New' can have multiple different meanings. We have a length problem in French (new = nouveau) and the right word would be create (créer) but that felt weird in English.

I understand the need to accommodate the interface to mobile devices (width) and different languages (text length). And I think going with the shorter `New` is the way forward. But again this is  a fundamental change in the way Symphony talks to the users that should not be done in a minor release. It should be part of Symphony 3/4 (not sure about the future numbering) and it should be done properly: short labels should become the new standard (new, edit, view).

nitriques:
Thanks guys for the feedback. Will check what can be done. I personally do not agree about the arrows: In the pagination, they do have a meaning, but if both directions are represented with the same "icon", they're just fancy bullet points.

I agree with the "breaking change" nature of things and will revert the things that do. But this will have to change in next versions.

nilshoerrmann:
> Thanks guys for the feedback. Will check what can be done. I personally do not agree about the arrows: In the pagination, they do have a meaning, but if both directions are represented with the same "icon", they're just fancy bullet points.

The original design used arrows in both directions:

- associations from the current entry to another pointed to the right (outgoing)
- associations from another entry to the current one pointed to the left (incomming)

If I remember correctly, the initial idea was also including bidirectional arrows, if there were relationships from both directions (different fields, pointing from different directions). Having different arrow directions was controversial, [see the discussion at that time](https://github.com/symphonycms/symphony-2/pull/1714).

This is a screenshot of the early design:

68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f36393236382f3332373331352f35343235363465632d396236662d313165322d386262622d6631346362376466343438642e706e67

> But this will have to change in next versions.

I completely agree!

nitriques:
Closing, see #2726 which rebases this PR.

PR #2701, opened 2017-06-06
Push deprecated warnings to log

### LINKED ISSUE #2693: [Suggestion]: Log deprecated usage
Would this helps extensions developers ?

nitriques:
I've logged all method calls. Defines are pretty hard to catch, so it's
a can't do. I've also fixed some docs along the way.

Fixes #2693 

It also adds stack trace info in deprecate message

### DIFF
--- a/symphony/lib/core/class.log.php
+++ b/symphony/lib/core/class.log.php
@@ -289,6 +289,10 @@ class Log
      *  The sprintf format to apply to $method
      * @param string $opts.alternative-format
      *  The sprintf format to apply to $alternative
+     * @param string $opts.removal-format
+     *  The sprintf format to apply to $opts.removal-version
+     * @param string $opts.removal-version
+     *  The Symphony version at which the removal is planned
      * @param boolean $opts.write-to-log
      *  If set to true, this message will be immediately written to the log. By default
      *  this is set to false, which means that it will only be added to the array ready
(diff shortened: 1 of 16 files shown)


symphony/lib/core/class.log.php
```diff
@@ -275,6 +275,65 @@ public function pushExceptionToLog($exception, $writeToLog = false, $addbreak =
...
+     * @param boolean $opts.append
+     *  If set to true, the given `$message` will be append to the previous log
+     *  message found in the `$_log` array
+     * @param boolean $opts.addtrace
+     *  If set to true, the caller of the function will be added. Defaults to true.
+     * @return boolean|null
+     *  If `$writeToLog` is passed, this function will return boolean, otherwise
+     *  void
+     */
+    public function pushDeprecateWarningToLog($method, $alternative = null, array $opts = array())
```
brendo: This is nice, only thing I can think of adding would be the removed version (if applicable). eg. `The method `blah` is deprecated and will be removed in v3.0.0. Please use `foo` instead.`

nitriques: Nice idea!

nitriques: @brendo See 	ca36cf0 and how using the "array of settings" pattern makes it simple to add it ;)

PR #2702, opened 2017-06-06
Group 404 checkboxes and shorten their labels

### LINKED ISSUE #2250: 404 Redirection
The checkbox labels are hard to read (length) and cannot be differentiated easily (similarity):

bildschirmfoto 2014-11-10 um 13 38 38

nitriques:
Trying to associate the first 2 checkbox with their respective textbox
while having a unrelated thrid checkbox is confusing. Also, all labels
were are hard to read (length) and cannot be differentiated easily
(similarity)

Fixes #2250

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -324,8 +324,15 @@ class contentBlueprintsDatasources extends ResourcesPage
 
         $fieldset->appendChild($group);
 
+        $this->Form->appendChild($fieldset);
+
+        $fieldset = new XMLElement('fieldset');
+        $this->setContext($fieldset, array('sections', 'system'));
+        $fieldset->appendChild(new XMLElement('legend', __('Page not found')));
+        $p = new XMLElement('p', __('Redirect to 404 page by enabling those criteria.'));
+        $p->setAttribute('class', 'help');
+        $fieldset->appendChild($p);
         $group = new XMLElement('div');
-        $group->appendChild(new XMLElement('label', __('Redirect to 404 page when:')));
 
         $label = Widget::Checkbox('fields[redirect_on_required]', $fields['redirect_on_required'], __('The required parameter is missing'));
         $group->appendChild($label);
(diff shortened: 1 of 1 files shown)


nitriques:
Without fieldset:
screen shot 2017-06-06 at 10 00 33 am

nitriques:
With a fieldset:
screen shot 2017-06-06 at 10 05 16 am

With this perspective, I think the wording must be adjusted. Conditions could become Parameters ?

symphony/content/content.blueprintsdatasources.php
```diff
@@ -324,22 +324,26 @@ public function __form()
 
         $fieldset->appendChild($group);
 
+        $this->Form->appendChild($fieldset);
+
+        $fieldset = new XMLElement('fieldset');
+        $this->setContext($fieldset, array('sections', 'system'));
+        $fieldset->appendChild(new XMLElement('legend', __('Page not found')));
+        $p = new XMLElement('p', __('Redirect to 404 page by enabling those criteria.'));
```
nilshoerrmann: This copy reads a bit strange. Maybe @brendo has an idea how to improve it.

nilshoerrmann:
Maybe the legends could read "Execution Rules" and "Redirect Rules". The second rule could then be changed to:

> # Redirect Rules
> Enabling these conditions will throw a 404 page not found error.
>
> - [ ] required parameter missing
> - [ ] forbidden parameter present
> - [ ] no results

nitriques:
I am not found of the "Redirection", since it is not a 30x status, but a real 404. But that's the engineer speaking ;)

symphony/content/content.blueprintsdatasources.php
nitriques: Yes, yes!

nilshoerrmann:
In this case "Error Rules" might be a engineer-friendly alternative ;)

nitriques:
I like it!
screen shot 2017-06-06 at 10 05 16 am

nitriques:
I would be good to merge as-is, if nobody is opposed.

brendo:
What about `Execution Conditions` and `Error Conditions`? Otherwise fine

nitriques:
Yes `conditions` is better (Since they are not rules *per se*)
@nilshoerrmann ?

nitriques:
@nilshoerrmann if you could approve the PR that would be neat! Thanks!

nitriques:
> I can’t approve it, though, I guess because it's already merged?

Yup ;)

BTW, the review thing is really great 👍

PR #2705, opened 2017-06-08
Make sure we always have a date and time format

### LINKED ISSUE #2704: Default date preferences
If – in the config – the date preferences are empty, Symphony doesn't default to any standard format. It just doesn't output dates anymore (showing `None`).

```php
'region' => array(
	'time_format' => null,
	'date_format' => null,
	'datetime_separator' => null,
	'timezone' => 'Europe/Berlin',
),
```

This caused me to think that something went terribly wrong and the database was corrupted. 

### Expected behaviour

I'd expect Symphony:

1. to log an error,
2. to show a warning in the interface,
3. to fall back to a default format.

nitriques:
If, somehow, the date_format or time_format config is missing, Symphony
stops displaying dates. This is problematic as we rely on date
formatting a lot.

This change adds a default values for date_format and time_format
directly in the DateTimeObj class. It also makes sure we have a valid
timezone. If not, the date.timezone php.ini value is used. If it is
still empty, we default to UTC.

This change made is possible to remove Configuration()->get() calls and
replace them with __SYM_DATE_FORMAT__ or __SYM_TIME_FORMAT__ constant
when possible, or uses DateTimeObj::getSetting() which shoulw always
return a value.

Finally, this change also adds page alerts when a missing date_format or
time_format is empty.

Fixes #2704

### DIFF
--- a/symphony/lib/core/class.datetimeobj.php
+++ b/symphony/lib/core/class.datetimeobj.php
@@ -17,10 +17,15 @@ class DateTimeObj
      * Holds the various settings for the formats that the `DateTimeObj` should
      * use when parsing input dates.
      *
+     * @since Symphony 2.7.0 it contains default values to prevent the case where
+     * no settings are set
      * @since Symphony 2.2.4
      * @var array
      */
-    private static $settings = array();
+    private static $settings = array(
+        'time_format' => 'g:i a',
+        'date_format' => 'm/d/Y',
+    );
 
     /**
      * Mapping PHP to Moment date formats.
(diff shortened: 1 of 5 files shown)


symphony/lib/core/class.datetimeobj.php
```diff
@@ -17,10 +17,15 @@ class DateTimeObj
...
+     * @since Symphony 2.7.0 it contains default values to prevent the case where
+     * no settings are set
      * @since Symphony 2.2.4
      * @var array
      */
-    private static $settings = array();
+    private static $settings = array(
+        'time_format' => 'g:i a',
+        'date_format' => 'm/d/Y',
+    );
```
nilshoerrmann: There also needs to be a default separator value.

nilshoerrmann: The install's default is a space character.

nitriques: Yup, it is already there at https://github.com/symphonycms/symphony-2/pull/2705/files#diff-5420cf877b2e40bf5839ca454c3544bcR92

PR #2717, opened 2017-06-20
[2.7.0RC1] Minor improvements

jensscherbl:
NPM wouldn’t install the dependencies because of an invalid version number. I also added the package-lock.json to the repository, as recommended by NPM, and updated the README to no longer suggest to install Grunt globally. The reason for the latter is that you could have different projects that depend on different versions of Grunt, so it’s better to use the version that comes with the project.


README.markdown
```diff
@@ -194,19 +194,13 @@ Follow the instructions below if you are updating from Symphony 2.0 (not from Gi
...
-		npm install -g grunt-cli
-
-2. Install all dependencies from the repository’s root:
+1. Install all dependencies from the repository’s root:
 
 		npm install
 
-3. Run the `watch` task:
-
-		grunt watch
```
nitriques: Why remove the `grunt watch` point ?

README.markdown
```diff
@@ -194,19 +194,13 @@ Follow the instructions below if you are updating from Symphony 2.0 (not from Gi
 
 Symphony uses [Grunt](http://gruntjs.com/) to build concatenated and minified assets:
 
-1. Install `grunt-cli` globally ([see the Grunt documentation](http://gruntjs.com/getting-started)):
-
-		npm install -g grunt-cli
```
nitriques: If we remove this, don't we need to add the `grunt-cli` package into package.json ? I think the idea behind having the cli vs the lib is that it allows you to have a specific grunt version per project, but keep a single cli tool around that calls in the right version. From http://gruntjs.com/getting-started: 

> Note that installing grunt-cli does not install the Grunt task runner! The job of the Grunt CLI is simple: run the version of Grunt which has been installed next to a Gruntfile. This allows multiple versions of Grunt to be installed on the same machine simultaneously.

nitriques:
Thanks BTW :)

README.markdown
jensscherbl: > I think the idea behind having the cli vs the lib is that it allows you to have a specific grunt version per project

True, I did only read the manual after submitting the change. I still think it’s better not to rely on anything to be installed globally, but my original argument is no longer valid. 

Feel free to ignore this and only fix the version number.

> If we remove this, don't we need to add the grunt-cli package into package.json ?

No, for the exact same reasons you mentioned. It's only a "remote control" for the actual lib in the project. If we use the lib in the project directly, we no longer need the remote.

README.markdown
jensscherbl: If I have to “build concatenated and minified assets” for whatever reason, the most straight forward way is to just run grunt. Watch is only useful while actively working on the assets during development. I guess we could/should actually mention and explain both, but the watch task is not the default or most convenient way to trigger a single build.

jensscherbl:
@nitriques GitHub says “changes requested”. Do you still want me to undo the “controversial” changes, or can it be merged now that the reasoning is more clear?

README.markdown
nitriques: Cool! I'll cherry pick this!

nitriques:
> Do you still want me to undo the “controversial” changes,

No I'll deal with this thanks! I have "special" bookmark with filters already applied so it is easy to close them in bunch when I do clean up. That way, I am sure I do not have "stale" issues. That's also why issues are not closed ;)

The "request changes" is github's new review workflow. I still did not properly test your branch, so I won't update my review status until I've properly done it (again, easy to find with filters). I like that workflow a lot, it simplifies tracking the whole project.

The only thing I'd like you to do is to review when asked :) Or even when not asked, I thing you can start a review, and this become a "reviewer". If you do not have the time to do it, it is ok, I generally wait for at least one review before merging. Thanks!

jensscherbl:
> No I'll deal with this thanks!

Thanks, and sorry for the mess. Will try to do more fine-grained commits and pull requests in the future again.

nitriques:
> Thanks, and sorry for the mess. Will try to do more fine-grained commits and pull requests in the future again.

No problems ;)

vendor/composer/ClassLoader.php
```diff
@@ -313,29 +334,34 @@ public function loadClass($class)
      */
     public function findFile($class)
     {
-        // work around for PHP 5.3.0 - 5.3.2 https://bugs.php.net/50731
-        if ('\\' == $class[0]) {
-            $class = substr($class, 1);
-        }
```
nitriques: Symphony technically still is compatible with PHP 5.3. I think I'll wait for a major version before dropping that line...

jensscherbl: This has been auto-generated by composer.

README.markdown
nitriques: Ok good point

vendor/composer/ClassLoader.php
nitriques: Ok so I'll wait for a major version to merge this than.

jensscherbl: No, it’s not a breaking change. According to the comment, it’s a workaround for an issue in PHP `5.3.0`-`5.3.2`, so I’d guess that the issue has been fixed in `5.3.3`. The last PHP `5.3.x` release was `5.3.29`.

vendor/composer/autoload_static.php
```diff
@@ -0,0 +1,184 @@
...
+    );
+
+    public static function getInitializer(ClassLoader $loader)
+    {
+        return \Closure::bind(function () use ($loader) {
+            $loader->classMap = ComposerStaticInit5f867b58cbaae676dc2f623c94dcd5ad::$classMap;
+
+        }, null, ClassLoader::class);
+    }
+}
```
nitriques: Do we still need the vendor/composer/autoload_classmap.php file if we have this one instead ?

jensscherbl: Everything in `vendor` is auto-generated by Composer, so yes.

nitriques: > Only thing to discuss would be if we should gitignore the vendor folder entirely and only commit the composer.json and composer.lock, thus require Symphony users to download Composer and run composer install locally in order to use Symphony.

This is what @brendo and I wanted for next release. The vendor folder would be in the zip archive only. Do you agree on that proposal ?

jensscherbl: > Do you agree on that proposal ?

Thinking about the best approach for a while now, with no conclusion. Both approaches have their pros and contras. Feel free to do what you think makes most sense for the project.

jensscherbl: > The vendor folder would be in the zip archive only.

Actually, I think that’s a pretty good solution.

nitriques: :+1:

nitriques:
I've picked 256e8d26c6b19e9a7587309781c6c62236aaa07e and 33be3b83dc1ca9fe17014f515749334f24e682e7 as the other stuff is already fixed in my 2.7.0-release branch. Thanks!

jensscherbl:
👍

PR #2721, opened 2017-06-27
Make sure buildSortingSelectSQL() does not crash

### LINKED ISSUE #2718: [2.7.0-RC1] Unknown column 'ed.value' in 'field list'
Upgrading an existing site to `2.7.0-RC1` resulted in the following error in the backend:

```sql
SELECT SQL_CACHE DISTINCT `e`.`id`, `e`.section_id, `e`.`author_id`, `e`.`modification_author_id`, `e`.`creation_date` AS `creation_date`, `e`.`modification_date` AS `modification_date` , `ed`.`value` 
FROM `sym_entries` AS `e` 
WHERE 1 AND `e`.`section_id` = 3 
ORDER BY ( 
    SELECT `ed`.value 
    FROM sym_entries_data_42 AS `ed` 
    WHERE entry_id = e.id 
) 
ASC LIMIT 0, 20
```

I'm using the latest code from the `2.7.0` branch, which includes a few fixes added after the pre-release. The update script run without any errors.

nitriques:
The default implementation of `buildSortingSelectSQL()` is well suited
for Fields that do not provide a custom `buildSortingSQL()`
implementation. But this can make Fields with a custom `buildSortingSQL()`
implementation without a `buildSortingSelectSQL()` crash if they do not
have the 'value' column.

~~This commits adds an extra SQL query to first check that the colunm
exist. For performance reason, this result is memoized.~~

^ This idea is stupid. It does not even fixes the bug for real. We only need to check if the thing we want to add in the select exists in the sort expression.

Fixes #2718

### DIFF
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1562,7 +1562,11 @@ class Field
      * sorting of this field. By default this will sort the results by
      * the entry id in ascending order.
      *
+     * Extension developers should always implement both `buildSortingSQL()`
+     * and `buildSortingSelectSQL()`.
+     *
      * @uses Field::isRandomOrder()
+     * @see Field::buildSortingSelectSQL()
      * @param string $joins
      *  the join element of the query to append the custom join sql to.
      * @param string $where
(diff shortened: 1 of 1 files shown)


jensscherbl:
> This seems to fix my issues just fine.

Yep, well done!

nitriques:
The only way it could fail, would be to do a `ORDER BY` with a sub-query, that aliases the table to `ed` and uses the string "`ed`.`value`" in it. Never saw this in the wild.

Issue #2710, opened 2017-06-18
Copy text on Blueprints > Data Sources should be improved


michael-e:
In https://github.com/symphonycms/symphony-2/pull/2702 changes to the data source GUI were introduced. Now we have:

> Enabling these conditions will send a 404 page not found error.

Please, read this aloud.

**Enabling the conditions** will throw an error?

We also have:

> Leaving these fields empty will always execute the data source.

The negation makes this hard to understand. Leaving something empty will do something…

We should improve this. Regarding the "Error Conditions", I would prefer "Error 404 Conditions". 

Maybe we don't need additional text at all?

@nils, @brendo

nitriques:
The wording has been discussed in length in [the the PR](https://github.com/symphonycms/symphony-2/pull/2702#issuecomment-306506725). Maybe the sentence does not make sense because the <code> format is not that apparent.

> Enabling these conditions will send a 404 page not found error.

It should be understood as 

Enabling these conditions will send a "404 page not found" error.

> We also have:

>> Leaving these fields empty will always execute the data source.

[This wording has been introduced 4 years ago](https://github.com/symphonycms/symphony-2/blame/d1be054abbe03638b78c7d12fdb005d31c442d49/symphony/content/content.blueprintsdatasources.php)

In understand your concerns, but that  one is not related to #2702

michael-e:
> It should be understood as: Enabling these conditions will send a "404 page not found" error.

No, this is semantically wrong. Please think about it. "Enabling conditions" won't send an error—the conditions must also be met when a datasource is executed. It would be much more correct to say: "If enabled conditions are met, a 404 error will be sent".

This should be fixed, really.

---

>> Leaving these fields empty will always execute the data source.

> This wording has been introduced 4 years ago

Ok, in this case this is nothing that must be fixed fast. Somehow the new layout made it more prominent to me.

nitriques:
@brendo since you are the writer behind that sentence and since you are a native English speaker, I'd like to get your feedback on this.

I personally find @michael-e proposition to be clearer, but I would put the whole error name in there.

> If enabled conditions are met, a `404 page not found` error will be sent

jensscherbl:
> Error 404 Conditions

I agree that the current wording is not perfect, and I like @michael-e’s suggestion as well, but this can always be improved in a patch release, so I wouldn't consider this to be a blocker for the `2.7.0` release.

nitriques:
I'll leave this open for the 

[...]

PR #2723, opened 2017-06-29
Better wording for the Error Conditions

nitriques:
Re: #2710

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -331,7 +331,7 @@ class contentBlueprintsDatasources extends ResourcesPage
         $fieldset = new XMLElement('fieldset');
         $this->setContext($fieldset, array('sections', 'system'));
         $fieldset->appendChild(new XMLElement('legend', __('Error Conditions')));
-        $p = new XMLElement('p', __('Enabling these conditions will send a <code>404 page not found</code> error.'));
+        $p = new XMLElement('p', __('If enabled conditions are met, a <code>"404 page not found"</code> error will be sent.'));
         $p->setAttribute('class', 'help');
         $fieldset->appendChild($p);
         $group = new XMLElement('div');
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -331,7 +331,7 @@ class contentBlueprintsDatasources extends ResourcesPage
         $fieldset = new XMLElement('fieldset');
         $this->setContext($fieldset, array('sections', 'system'));
         $fieldset->appendChild(new XMLElement('legend', __('Error Conditions')));
-        $p = new XMLElement('p', __('Meeting one of these conditions will send a <code>404 Not Found</code> error.'));
+        $p = new XMLElement('p', __('Meeting one of these conditions will cause a <code>404 Not Found</code> response.'));
         $p->setAttribute('class', 'help');
         $fieldset->appendChild($p);
         $group = new XMLElement('div');
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -331,7 +331,7 @@ class contentBlueprintsDatasources extends ResourcesPage
         $fieldset = new XMLElement('fieldset');
         $this->setContext($fieldset, array('sections', 'system'));
         $fieldset->appendChild(new XMLElement('legend', __('Error Conditions')));
-        $p = new XMLElement('p', __('If enabled conditions are met, a <code>"404 page not found"</code> error will be sent.'));
+        $p = new XMLElement('p', __('Meeting one of these conditions will send a <code>404 Not Found</code> error.'));
         $p->setAttribute('class', 'help');
         $fieldset->appendChild($p);
         $group = new XMLElement('div');


symphony/content/content.blueprintsdatasources.php
```diff
@@ -331,7 +331,7 @@ public function __form()
         $fieldset = new XMLElement('fieldset');
         $this->setContext($fieldset, array('sections', 'system'));
         $fieldset->appendChild(new XMLElement('legend', __('Error Conditions')));
-        $p = new XMLElement('p', __('Enabling these conditions will send a <code>404 page not found</code> error.'));
+        $p = new XMLElement('p', __('If enabled conditions are met, a <code>"404 page not found"</code> error will be sent.'));
         $p->setAttribute('class', 'help');
```
nilshoerrmann: What about:

> Meeting one of these conditions will return a `404 Not Found` error.

brendo: enabled is awkward here, could just be `If conditions are met...`

@nilshoerrmann's suggestion also works, feels more Symphonylike too.

nitriques: 1. > Do we need the quotes if the response is marked up as code?

No, but on low constrast screens, it made the error name pop out.

2. I also like @nilshoerrmann proposition. The "if" part is not necessary.

Also thanks @michael-e for pointing out the fact the the error name does not include the word 'page'. I will never get enough of HTTP status code geekness.

nitriques: I am not found of the word 'return'. The proper word IMHO would be 'sent'. Or maybe that's my French specking brain ?

michael-e: To be really picky, it is just a _response_. So it might read:

Meeting one of these conditions will yield a `404 Not Found` response.

Or:

Meeting one of these conditions will result in a `404 Not Found` response.

michael-e:
I was too late. :-(

symphony/content/content.blueprintsdatasources.php
nitriques: (The following not meant to be understood as a criticism!  I like to understand and learn!)

Ok @michael-e I want to be picky ;)

1. Result is related to the causality.
2. Yeild feels more like a side-effect thing.

> It is a response.

Yup. Responses are SENT and RECEIVED.

Random example in the spec: https://tools.ietf.org/html/rfc7231

> The 100 (Continue) status code indicates that the initial part of a
   request has been received and has not yet been rejected by the
   server.  The server intends to send a final response after the
   request has been fully received and acted upon.

Boy am I having fun 👍

nitriques:
It's never too late.

symphony/content/content.blueprintsdatasources.php
nitriques: 'generated' would also seems to be a good term, as per spec ;)

michael-e: I agree that responses are sent and received. This is done by the server resp. "the software", normally. But currently we say that "meeting a condition" will send a response, which sounds really strange to my ears. How can "meeeting a condition" actively send something? It can only cause the software to do it. (That's why I used the verb "yield".)

I would like to hear what native speakers think about it.

nitriques: Ah! I see your point!

> I would like to hear what native speakers think about it.

@brendo seemed fine with return, then. I'll change it.

Thanks for playing with me @michael-e

michael-e: Return is not much better, it still sounds like "taking action". Maybe @brendo will follow my additional thoughts and come up with something better?

michael-e: Oh, what about "causing a response"?

Meeting one of these conditions will cause a `404 Not Found` response.

michael-e:
There is no reason for a UI copy to be technically disputable, unless you have to wrap it up really quick (which is not true for Symphony 2.7). On the contrary, I think that a UI copy that is readable and technically correct alike is a very good advertisment. As you say: The UI is what people will see!

nilshoerrmann:
Yes, but from a user's point of view, the options just tell: “This will display my error page.” Nothing more, nothing less.

michael-e:
I don't like this discussion. Symphony has taught me so much because the people who built it thought about every detail. And now you say "Let's just make it sloppy, who cares!".

nilshoerrmann:
That's not what I mean – and you know that. I'm just pointing out a different perspective.

michael-e:
I know, but let's rather spend our time on a correct and readable copy text.

What do you guys think about my latest proposal [from here](https://github.com/symphonycms/symphony-2/pull/2723/commits/eb2eb40d5479155871c6880f9299973bb446bb61#r125090733):

Meeting one of these conditions will cause a `404 Not Found` response.

nitriques:
> Meeting one of these conditions will cause a 404 Not Found response.

Sold!

michael-e:
Thanks a lot!

nitriques:
I'll still wait before merging and maybe get @brendo's approval. 
@nilshoerrmann is it ok with you ? I feel like we have the right balance, and the technicality I care so much about is safe.

Issue #2737, opened 2017-08-21
Creating a new entry while using prepopulate throws PHP error


jensscherbl:
Affected Symphony version(s) : `2.7.0`
PHP version(s) : `5.6.30`
OS(es) : `OS X 10.11.6`

![screen shot 2017-08-21 at 14 19 28](https://user-images.githubusercontent.com/1640033/29521060-26414244-8684-11e7-937c-fa1d55234936.png)

![screen shot 2017-08-21 at 14 18 00](https://user-images.githubusercontent.com/1640033/29521064-2ba012f6-8684-11e7-9e8c-dc917e3ec16b.png)

![screen shot 2017-08-21 at 14 18 15](https://user-images.githubusercontent.com/1640033/29521071-30829078-8684-11e7-85ce-013d3a007e9d.png)

nitriques:
Confirmed. Thanks Jens.

PR #2738, opened 2017-08-21
Fixes an issue where creating a new entry while using prepopulate throws a PHP error

jensscherbl:
Fix for #2737 

[EntryManager::fetch()][1] is expected to [return an array][2], but instead [returns a boolean][3] in some cases. I changed the method to always return an array as documented. Also adapted the handling of the return value in all places.

[1]: https://github.com/symphonycms/symphony-2/blob/2f9dc2dcf345fa6b0859f4c18931a86475f7cb1c/symphony/lib/toolkit/class.entrymanager.php#L379
[2]: https://github.com/symphonycms/symphony-2/blob/2f9dc2dcf345fa6b0859f4c18931a86475f7cb1c/symphony/lib/toolkit/class.entrymanager.php#L375
[3]: https://github.com/symphonycms/symphony-2/blob/2f9dc2dcf345fa6b0859f4c18931a86475f7cb1c/symphony/lib/toolkit/class.entrymanager.php#L385

The change makes sure that the method actually works as expected and documented, so I’d consider this to be a bugfix and not a breaking change.

### DIFF
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -1778,9 +1778,10 @@ class contentPublish extends AdministrationPage
                     $header = new XMLElement('header');
 
                     // Get the search value for filters and prepopulate
-                    $filter      = '';
+                    $filter = '';
                     $prepopulate = '';
-                    if ($entry = current(EntryManager::fetch($entry_id))) {
+                    $entry = current(EntryManager::fetch($entry_id));
+                    if ($entry) {
                         $search_value = $relation_field->fetchAssociatedEntrySearchValue(
                             $entry->getData($as['parent_section_field_id']),
                             $as['parent_section_field_id'],
@@ -1789,7 +1790,7 @@ class contentPublish extends AdministrationPage
                         if (is_array($search_value)) {
                             $search_value = $entry_id;
                         }
-                        $filter      = '?filter[' . $relation_field->get('element_name') . ']=' . $search_value;
+                        $filter = '?filter[' . $relation_field->get('element_name') . ']=' . $search_value;
                         $prepopulate = '?prepopulate[' . $as['child_section_field_id'] . ']=' . $search_value;
                     }
 
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -1189,7 +1189,8 @@ class contentPublish extends AdministrationPage
 
         EntryManager::setFetchSorting('id', 'DESC');
 
-        if (!$existingEntry = EntryManager::fetch($entry_id)) {
+        $existingEntry = EntryManager::fetch($entry_id);
+        if (empty($existingEntry)) {
             Administration::instance()->throwCustomError(
                 __('Unknown Entry'),
                 __('The Entry, %s, could not be found.', array($entry_id)),
@@ -1397,7 +1398,8 @@ class contentPublish extends AdministrationPage
         $entry_id = intval($this->_context['entry_id']);
 
         if (is_array($_POST['action']) && (array_key_exists('save', $_POST['action']) || array_key_exists('done', $_POST['action']))) {
-            if (!$ret = EntryManager::fetch($entry_id)) {
+            $ret = EntryManager::fetch($entry_id);
+            if (empty($ret)) {
                 Administration::instance()->throwCustomError(
                     __('The Entry, %s, could not be found.', array($entry_id)),
                     __('Unknown Entry'),
@@ -1504,7 +1506,8 @@ class contentPublish extends AdministrationPage
 
                 redirect(SYMPHONY_URL . '/publish/'.$this->_context['section_handle'].'/');
             } else {
-                if (is_array($ret = EntryManager::fetch($entry_id))) {
+                $ret = EntryManager::fetch($entry_id);
+                if (!empty($ret)) {
                     $entry = $ret[0];
                     $this->addTimestampValidationPageAlert($this->_errors['timestamp'], $entry, 'delete');
                 }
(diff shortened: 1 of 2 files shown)


symphony/content/content.publish.php
```diff
@@ -1775,19 +1778,21 @@ private function prepareAssociationsDrawer($section)
...
-                    $search_value = $relation_field->fetchAssociatedEntrySearchValue(
-                        $entry->getData($as['parent_section_field_id']),
-                        $as['parent_section_field_id'],
-                        $entry_id
-                    );
-                    if (is_array($search_value)) {
-                        $search_value = $entry_id;
+                    $filter      = '';
+                    $prepopulate = '';
+                    if ($entry = current(EntryManager::fetch($entry_id))) {
```
nitriques: Is this construct still encouraged ? I mean, it is not super clear that it is an assignment and not a boolean test.

symphony/content/content.publish.php
```diff
@@ -1775,19 +1778,21 @@ private function prepareAssociationsDrawer($section)
...
+                        $search_value = $relation_field->fetchAssociatedEntrySearchValue(
+                            $entry->getData($as['parent_section_field_id']),
+                            $as['parent_section_field_id'],
+                            $entry_id
+                        );
+                        if (is_array($search_value)) {
+                            $search_value = $entry_id;
+                        }
+                        $filter      = '?filter[' . $relation_field->get('element_name') . ']=' . $search_value;
+                        $prepopulate = '?prepopulate[' . $as['child_section_field_id'] . ']=' . $search_value;
```
nitriques: Is this alignment recommended by PSR ? IHMO, it just makes the line longer

symphony/lib/toolkit/class.entrymanager.php
```diff
@@ -391,7 +391,7 @@ public static function fetch($entry_id = null, $section_id = null, $limit = null
 
         $section = SectionManager::fetch($section_id);
         if (!is_object($section)) {
-            return false;
+            return array();
```
nitriques: I find it really surprising that the doc did not mentioned the boolean return. I got caught by this many many times, as you can tell. ;)

Even though, technically, it is not a breaking fix, it think it has the potential of breaking some extensions. We'll see ;)

michael-e: The function's doc block says that the function returns an array. So it should never return `false`, should it? If this breaks an extension, the extension's author hasn't read the docs.

When using the `fetch` method , you should always do `if (EntryManager::fetch())`, never `if (EntryManager::fetch() === false)`.

nitriques: > If this breaks an extension, the extension's author hasn't read the docs.

Ahah I've read the doc multiple time... but when debugging, I'd rather code with what the function is giving me.

`if (array())` returns false ? I've never tried it to be honest...

nitriques: BTW I think other managers have the same problem.

michael-e: Oh yes, it does! See http://php.net/manual/en/language.types.boolean.php#language.types.boolean.casting

This is why `!empty($var)` is the same as `isset($var) && $var`(see http://php.net/manual/en/function.empty.php#refsect1-function.empty-parameters) — and I prefer the latter, because it is more explicit.

michael-e: Heaven or hell: PHP's "type juggling". :-)

michael-e: > BTW I think other managers have the same problem.

So I wouldn't oppose fixing them. I really think that (most) functions should have a single return type.

nitriques: > I prefer the latter, because it is more explicit.

I prefer empty because "it's magic". Thanks Michael!

> Heaven or hell: PHP's "type juggling". :-)

I am still pretty formatted javascript/C/c++/java/.NET even tho I've been using PHP since version 3.

symphony/content/content.publish.php
jensscherbl: I don’t think that it was ever encouraged. Not sure if it’s specifically discouraged, though. I don't have a strong opinion about it, can change it to separate lines if you find it more readable that way.

symphony/content/content.publish.php
jensscherbl: PSR has no opinion about vertical alignment. Personally, I find it more readable this way, separates the variables and the values more clearly.

symphony/lib/toolkit/class.entrymanager.php
jensscherbl: > If this breaks an extension, the extension's author hasn't read the docs.

I agree with @michael-e on this. If we had used proper (PHP 7) return types instead of just documenting the return type, the method would have thrown an exception. Extensions should never rely on bugs, accidental or undocumented behavior. If fixing a bug breaks an extension, there’s something wrong with the extension.

symphony/content/content.publish.php
nitriques: Yes I think it is more readable on two lines. Thanks.

symphony/content/content.publish.php
nitriques: Hum, I'd have to disagree on that. But in the current codebase, I think we do not use this more than we use it.

symphony/lib/toolkit/class.entrymanager.php
nitriques: > Extensions should never rely on bugs, accidental or undocumented behavior.

Indeed. But in the wild, that's not the case. Remember how hidden files in UNIX are actually a bug, that became a feature because people relied on it.

That being said, I do not mind breaking extensions haha. I justed to document that it can be the case (and I'll merge the PR as soon as Jens can do the simple fixes asked in the review).

> Or a nullable return type

Yes I do not consider nullable as a separate  type either.

symphony/content/content.publish.php
michael-e: I personally use this alignment style only if I have more consecutive lines of parameters. If it's only 2 lines like here, I wouldn't do it. But that's just my personal taste. And if PSR has no opinion about it…

symphony/content/content.publish.php
jensscherbl: Done.

symphony/content/content.publish.php
jensscherbl: Done.

nitriques: Thanks.

PR #2741, opened 2017-08-21
Removes empty parent association sections from associations drawer

jensscherbl:
I find the empty parent association sections in the redesigned associations drawer quite noisy since the [2.7.0-release][2] (see screenshots). As far as I can tell, they also provide no real value, so I propose to remove them entirely. Not sure if this is something that should be changed in a patch release or better be cherry-picked for the [3.0.x-branch][1], but I’m working with the [2.7.0-code][2] at the moment, therefore submitting this to the [2.7.x-branch][3].

[1]: https://github.com/symphonycms/symphony-2/tree/3.0.x
[2]: https://github.com/symphonycms/symphony-2/tree/2.7.0
[3]: https://github.com/symphonycms/symphony-2/tree/2.7.x 

screen shot 2017-08-21 at 19 36 08

screen shot 2017-08-21 at 19 36 48

screen shot 2017-08-21 at 19 38 08

screen shot 2017-08-21 at 19 37 57

### DIFF
--- a/symphony/content/content.publish.php
+++ b/symphony/content/content.publish.php
@@ -1703,35 +1703,30 @@ class contentPublish extends AdministrationPage
                             'title' => strip_tags($aname),
                         ));
 
+                        if (!$has_entries) {
+                            unset($field);
+                            continue;
+                        }
+
                         $element = new XMLElement('section', null, array('class' => 'association parent'));
                         $header = new XMLElement('header');
+                        $header->appendChild(new XMLElement('p', $a->generate()));
                         $element->appendChild($header);
 
-                        if ($has_entries) {
-                            $element = new XMLElement('section', null, array('class' => 'association parent'));
-                            $header = new XMLElement('header');
-                            $header->appendChild(new XMLElement('p', $a->generate()));
-                            $element->appendChild($header);
+                        $ul = new XMLElement('ul', null, array(
+                            'class' => 'association-links',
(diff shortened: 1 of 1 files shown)


twiro:
Looks like sections without actual associations weren't listed in the "linked to"-part of the association drawer before 2.7.0 and I guess this change more or less "happened" while refactoring the drawer and wasn't really introduced by purpose.

So I'm all for Jens' proposal and remove these links in the next patch-release again.

jensscherbl:
@twiro Thanks for the detailed feedback, much appreciated.

nitriques:
The goal was to add information about the possible links. Without it, you loose information about all the possible associations. It is also weird to have empty sections in one way, but not the other way around.

I do not know if that fact comes from our data modelling. But I sure do know that Symphony should not care about one's data modelling, so I am tempted to add "Yet Another Configuration Setting"<sup>TM</sup>.

I'll think about it, but with 4 votes against 1 right now, you guys might expect me to merge it in the next days ;)

jensscherbl:
> It is also weird to have empty sections in one way, but not the other way around.

The difference is that having it for the other direction offers additional functionality because of the “new” button.

nitriques:
> The difference is that having it for the other direction offers additional functionality because of the “new” button.

Yeah I understand that ;) Still feels weird to me!

nitriques:
Well, even @alexnantel88 is for it, so I guess democracy wins. This time ;)

PR #2743, opened 2017-08-22
Fixes a breaking change in a hotfix

### LINKED ISSUE #2742: Deleting section throws PHP error
Affected Symphony version(s) : `2.7.0`
PHP version(s) : `5.6.30`
OS(es) : `OS X 10.11.6`

Deleting a section throws an error on the edit section view (not via bulk delete on the sections index view). The issue is caused by a new method [contentBlueprintsSections::validateTimestamp()][1] that has been introduced in `2.7.0`, which only expects a single section ID, but is instead given an array of section IDs in [content.blueprintssections.php#L892][2].

[1]: https://github.com/symphonycms/symphony-2/blob/2f9dc2dcf345fa6b0859f4c18931a86475f7cb1c/symphony/content/content.blueprintssections.php#L982
[2]: https://github.com/symphonycms/symphony-2/blob/2f9dc2dcf345fa6b0859f4c18931a86475f7cb1c/symphony/content/content.blueprintssections.php#L892

screen shot 2017-08-22 at 16 40 32

screen shot 2017-08-22 at 16 40 42

jensscherbl:
Fix for #2742

@nitriques 

Apparently you already fixed #2742. There wasn’t an issue for it, so I only discovered the fix after rebasing my own `2.7.0`-based hotfix on `2.7.x`. However, I think that your fix introduces a breaking change, so I’m submitting my original hotfix as well.

In `2.7.0`, we pass an array to the delegate by reference, which means that the array can be manipulated by the delegate. Your fix only passes the array value by reference, which is not the same, at least from my understanding.

For example, the delegate could decide to remove the section from the array in `2.7.0`, which would then not delete the section. Or add another section, which would then delete this section as well. I think that your proposed fix basically just bypasses the manipulation that could happen in the delegate.

### DIFF
--- a/symphony/content/content.blueprintssections.php
+++ b/symphony/content/content.blueprintssections.php
@@ -892,6 +892,8 @@ class contentBlueprintsSections extends AdministrationPage
             $canProceed = $this->validateTimestamp($section_id);
 
             if ($canProceed) {
+                $section_ids = array($section_id);
+
                 /**
                  * Just prior to calling the Section Manager's delete function
                  *
@@ -902,9 +904,11 @@ class contentBlueprintsSections extends AdministrationPage
                  * @param array $section_ids
                  *  An array of Section ID's passed by reference
                  */
-                Symphony::ExtensionManager()->notifyMembers('SectionPreDelete', '/blueprints/sections/', array('section_ids' => array(&$section_id)));
+                Symphony::ExtensionManager()->notifyMembers('SectionPreDelete', '/blueprints/sections/', array('section_ids' => &$section_ids));
 
-                SectionManager::delete($section_id);
+                foreach ($section_ids as $section) {
+                    SectionManager::delete($section);
+                }
 
                 redirect(SYMPHONY_URL . '/blueprints/sections/');
             } else {
--- a/symphony/content/content.blueprintssections.php
+++ b/symphony/content/content.blueprintssections.php
@@ -891,33 +891,34 @@ class contentBlueprintsSections extends AdministrationPage
             $section_id = $this->_context[1];
             $canProceed = $this->validateTimestamp($section_id);
 
-            if ($canProceed) {
-                $section_ids = array($section_id);
-
-                /**
-                 * Just prior to calling the Section Manager's delete function
-                 *
-                 * @delegate SectionPreDelete
-                 * @since Symphony 2.2
-                 * @param string $context
-                 * '/blueprints/sections/'
-                 * @param array $section_ids
-                 *  An array of Section ID's passed by reference
-                 */
-                Symphony::ExtensionManager()->notifyMembers('SectionPreDelete', '/blueprints/sections/', array('section_ids' => &$section_ids));
-
-                foreach ($section_ids as $section) {
-                    SectionManager::delete($section);
-                }
-
-                redirect(SYMPHONY_URL . '/blueprints/sections/');
-            } else {
+            if (!$canProceed) {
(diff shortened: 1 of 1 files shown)


symphony/content/content.blueprintssections.php
```diff
@@ -891,29 +891,34 @@ public function __actionNew()
...
-
-                redirect(SYMPHONY_URL . '/blueprints/sections/');
-            } else {
+            if (!$canProceed) {
                 $this->addTimestampValidationPageAlert(
                     $this->_errors['timestamp'],
                     SectionManager::fetch($section_id),
                     'delete'
                 );
+                return;
```
nitriques: I also prefer early returns. 👍  Let's try to use this as much as possible.

jensscherbl:
> Are you trying to add more sections to delete ? 

No, that was just an example of what an extension could do with the delegate.

> I do not see the point of creating and array with a single item and do a for loop on it. 

Agreed, but that’d be a change for 3.0.0.

> I understand that the interface of the delegate must not change, and I thought that this is what I did...

I think you changed it by accident. If you pass an array by reference, the delegate can change what’s in this array (e.g. add items, remove items, assign different values to existing keys, etc.), and the method then operates on that array (i.e. does something with every item of that array).

I made [a Gist to illustrate the problem][1], just run in from the command line.

Again, this is not about what makes sense or what we think should be done with the context data in a delegate, it’s just about not breaking existing stuff in a patch release.

> I am asking for a commit message

Agreed. I think we need to practice this a bit more, and try to figure out how to write better and more useful commit messages.

[1]: https://gist.github.com/jensscherbl/6563e124766b1c419b3e26e0caaf0787

nitriques:
Ah damn, I am missing a nested array, that's the problem ? (your proposed patched)

Because it felt like a security patch: Extension should not be allow to "add sections to delete" should they ? Was this ever documented ?

jensscherbl:
> Ah damn, I am missing a nested array, that's the problem ? (your proposed patched)

No. The problem is that the delegate passes an array by reference, which allows an extension to manipulate the array before the core operates on this array. Your proposed fix completely changed the logic here.

Take a look at [the Gist][1]: The `delegate`-function illustrates some extension. It relies on [documented functionality][2]. The code starting at [line 27][3] illustrates how the core works in `2.7.0`. In the example, it traverses the array (after passing it to the extension for further manipulation) and echoes its values (`5` and `42`). The code starting at [line 41][4] illustrates how the core would work in `2.7.1` after applying your patch. It would just echo `23` and ignore the array and what the extension does with the array.

[1]: https://gist.github.com/jensscherbl/6563e124766b1c419b3e26e0caaf0787
[2]: http://www.getsymphony.com/learn/api/2.7.0/delegates/#SectionPreDelete
[3]: https://gist.github.com/jensscherbl/6563e124766b1c419b3e26e0caaf0787#file-2742-php-L27
[4]: https://gist.github.com/jensscherbl/6563e124766b1c419b3e26e0caaf0787#file-2742-php-L41

nitriques:
Thanks Jens for this thorough explanation. I appreciate it.

> No. The problem is that the delegate passes an array by reference, which allows an extension to manipulate the array before the core operates on this array. 

Yeah, that was the whole idea !

> Extensions in Symphony can do all kinds of crazy shit, that’s what makes it so flexible and powerful.

Indeed.

> This is definitely not a security thing.

It is if it's a secret feature, with I thought it was. But yeah, it written `passed by reference` in the docs, so it is not hidden.

Thanks again!

PR #2744, opened 2017-08-25
Restore the 2.6.x field behaviour for empty arrays

### LINKED ISSUE #2733: Cannot delete content from "optional" fields
Apologies in advance if this has already been addressed, however, the behaviour is still present in the 2.7.x branch and I've not located a fix or existing issue. Appears this bug has been around since 18th of May.

**Affected Symphony version(s) :** Since PR #2664 

This relates to code in EntryManager line 119 (https://github.com/symphonycms/symphony-2/blob/2.7.x/symphony/lib/toolkit/class.entrymanager.php#L119)

### Expected Behaviour

If there is a field flagged as optional, editing it to have an empty value should cause anything that was there to be deleted.

### Current Behaviour

The original value persists.

### Steps to reproduce

1. Create a new section.
2. Add a text input field to that section and set it to optional
3. Create a new entry in that section, ensuring to add a value to the text input. e.g. "APPLE"
4. Go back to the entry to edit it. Delete the value "APPLE" from the text input field, then save the entry.
5. Observe the value "APPLE" persists.

### Reason

When an entry is saved, field data is passed through this method. For an optional field with no value, the data looks something like this:

```
int(113)
array(0) {
}
```

This code is executed near the top of `EntryManager::saveFieldData()`:

```
        // Check if we have field data
        if (!is_array($field) || empty($field)) {
            return;
        }
```

When this code is executed, it will skip over fields with an empty value. Consider the previous code https://github.com/symphonycms/symphony-2/blob/0dd2c1238a62aa620d7f7e5737843380a31c8b6a/symphony/lib/toolkit/class.entrymanager.php#L186

Note that a call to `Symphony::Database()->delete(...)` is triggered first. This allows the desired behaviour to occur. i.e. removing a value from an optional field. 

### Solution

Delete field data before checking for an empty state.

nitriques:
Symphony always treated empty arrays as being null values, i.e. deleting
old data without doing an insert statement afterwards.

The refactor made in 2.7.x used the code for new as the base, instead of
edit. new never had to deal with old data and simply did not produce a
DELETE statement, but edit did.

I think it is safe to change both, to always issue a DELETE statement
and produce an INSERT when needed.

Fixes #2733

### DIFF
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -115,8 +115,8 @@ class EntryManager
             return;
         }
 
-        // Check if we have field data
-        if (!is_array($field) || empty($field)) {
+        // Ignore when not an array
+        if (!is_array($field)) {
             return;
         }
 
@@ -158,7 +158,10 @@ class EntryManager
                 $fields[$index] = array_merge($data, $field_data);
             }
 
-            Symphony::Database()->insert($fields, $table_name);
+            // Insert only if we have field data
+            if (!empty($fields)) {
+                Symphony::Database()->insert($fields, $table_name);
+            }
         } catch (Exception $ex) {
             Symphony::Log()->pushExceptionToLog($ex, true);
         }


nitriques:
Already tested by @animaux

PR #2747, opened 2017-09-05
Update field.upload.php

byura:
Add SVG image support

### DIFF
--- a/symphony/lib/toolkit/fields/field.upload.php
+++ b/symphony/lib/toolkit/fields/field.upload.php
@@ -15,7 +15,8 @@ class FieldUpload extends Field implements ExportableField, ImportableField
         'image/jpeg',
         'image/pjpeg',
         'image/png',
-        'image/x-png'
+        'image/x-png',
+		'image/svg+xml'
     );
 
     public function __construct()


symphony/lib/toolkit/fields/field.upload.php
```diff
@@ -15,7 +15,8 @@ class FieldUpload extends Field implements ExportableField, ImportableField
         'image/jpeg',
         'image/pjpeg',
         'image/png',
-        'image/x-png'
+        'image/x-png',
+		'image/svg+xml'
```
michael-e: Please do not use tabs to indent code. Use spaces.

jensscherbl:
Just to make everyone aware, there’s discussion about this regarding potential security concerns. From what I recall, the issue is that an SVG could contain JavaScript that would be automatically executed by browsers. See https://github.com/w3c/svgwg/issues/266 and [Active Content Injection with SVG Files][1].

[1]: https://www.owasp.org/images/0/03/Mario_Heiderich_OWASP_Sweden_The_image_that_called_me.pdf

nitriques:
@jensscherbl is right. I do not want that in the core. If you want to be able to use it @byura I would recommend using a [custom validator file: manifest/validators.php](https://github.com/DeuxHuitHuit/symphonycms-template/blob/master/manifest/validators.php).

Issue #2748, opened 2017-09-05
Ampersand in static XML DS


jurajkapsz:
Affected Symphony version(s) : `2.6.11`
PHP version(s) : `5.6.x`
OS(es) : `Linux`

I have a static content with `ampersands` in a static XML data source. Ampersands are written as entities, tried out  `&#38;`, `&amp;` or `&amp;amp;`

**Issue:**
After save, DS editor returns those ampersands as `&` characters, which upon repeated save returns `XML is invalid.` error message.

**Expected:**
To not to need to repeatedly escape ampersands.

nitriques:
I can confirm the issue still exists in the latest 2.7.x branch.

PR #2749, opened 2017-09-07
Escape ampersand before validation

nitriques:
Fix for #2748


michael-e:
@nitriques: The "fix regression" commit has little to nothing to do with the issue. In this case it is not a big thing, but generally I would really love to have separate PRs for thematically different commits. This makes it much easier to follow (and to find things later).

nitriques:
>  but generally I would really love to have separate PRs

You are right. Sorry.

Thanks guys!

PR #2750, opened 2017-09-09
Fix for static XML data source

jensscherbl:
Fix for #2748. Also see #2749 for reference.

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -181,6 +181,11 @@ class contentBlueprintsDatasources extends ResourcesPage
                         } else {
                             $fields['static_xml'] = trim($existing->grab());
                         }
+                        $fields['static_xml'] = htmlspecialchars(
+                            $fields['static_xml'],
+                            ENT_XML1|ENT_COMPAT,
+                            'UTF-8'
+                        );
                         break;
                     default:
                         $fields['filter'][$fields['source']] = $existing->dsParamFILTERS;


symphony/content/content.blueprintsdatasources.php
```diff
@@ -181,6 +181,11 @@ public function __form()
                         } else {
                             $fields['static_xml'] = trim($existing->grab());
                         }
+                        $fields['static_xml'] = htmlspecialchars(
+                            $fields['static_xml'],
+                            ENT_XML1|ENT_COMPAT,
+                            'UTF-8'
+                        );
```
nitriques: I would personally move that assignment to line 949, because that where we need it (?)

nitriques: And maybe dropping the `General::sanitize` call is needed to fix the issue raised by @michael-e ?

michael-e: Unfortunately, removing the call to `General::sanitize` (when creating the `Widget::Textarea`) does not help at all with my issue.

michael-e:
I am afraid that, somehow, posted values are treated differently from saved values. (Only valid XML gets saved, so in case of invalid XML the posted value must be returned to the backend form element.) 

Or is invalid XML treated differently from valid XML?

### Case 1: valid XML

```xml
<data>
  <item>&amp;</item>
</data>
```

Saved value:

```php
    public $dsParamSTATIC = '<data>
  <item>&amp;</item>
</data>';
```

Value in the backend form (HTML):

```xml
<textarea name="fields[static_xml]" rows="12" cols="50" class="code" placeholder="<static>content</static>">&lt;data&gt;
&lt;item&gt;&amp;amp;&lt;/item&gt;
&lt;/data&gt;</textarea>
```

### Case 2: invalid XML

```xml
<data>
  <item>&amp;&</item>
</data>
```

Nothing gets saved in this case.

Value in the backend form (HTML):

```xml
<textarea name="fields[static_xml]" rows="12" cols="50" class="code" placeholder="<static>content</static>">&lt;data&gt;
&lt;item&gt;&amp;&amp;&lt;/item&gt;
&lt;/data&gt;</textarea>
```

symphony/content/content.blueprintsdatasources.php
nitriques: :(

nitriques:
Thanks @michael-e I'll dig deeper on this issue.

nitriques:
> Unfortunately, removing the call to General::sanitize (when creating the Widget::Textarea) does not help at all with my issue.

I think I've fixed it by NEVER changing the value in the $fields array.

I've removed the `sanitize` call, which would double escape things.

Valid XML gets saved. Invalid XML is returned to the browser AS IS. The invalid value does not change in the textbox, which is what we want amirite ?

nitriques:
Check 409f000c924477a094118517666923025cff52bc

michael-e:
@nitriques' [pr-2750 branch](https://github.com/DeuxHuitHuit/symphonycms/tree/pr-2750) looks good to me, test-wise. It does what I expect.

Still, I would like to be sure that @jensscherbl has no concerns.

nitriques:
Great

symphony/content/content.blueprintsdatasources.php
jensscherbl: > Unfortunately, removing the call to General::sanitize (when creating the Widget::Textarea) does not help at all with my issue.

I think it does?

jensscherbl:
409f000 is perfect.

nitriques:
Thanks @jensscherbl

nitriques:
Merged as d2c9b62789

PR #2751, opened 2017-09-12
Prepopulate fields that supports it

nitriques:
When prepopulating the field, we only checked if there was no POST
data for the field. But we also need to make sure the field supports
prepoluation. If not, it can create bugs.


nitriques:
@michael-e do you have any objections ?

michael-e:
No, sorry, I just missed that this was awaiting my review.

nitriques:
Thanks!

PR #2754, opened 2017-09-16
Update admin.css

byura:
fix css property


nitriques:
@jensscherbl we can now do it ourself via gihub!

image

jensscherbl:
> we can now do it ourself via gihub!

Ah, perfect!

PR #2755, opened 2017-09-16
Update symphony.min.css

byura:
fix css property


jensscherbl:
Thanks for #2754. Closing this because the minified assets will be generated automatically as part of our build process.

PR #2756, opened 2017-09-19
Do not hide exceptions thrown by the database

### LINKED ISSUE #2752: If the DB user can't lock the table, we get a "Entry saved" message
Affected Symphony version(s) : 2.7.0
PHP version(s) : ALL
OS(es) : ALL

@michael-e was that intended ? Thank god I was able to find it in the log.

Just stumbled upon this case where the DB user did not had the `LOCK TABLES` privilege. Should we deal with it ? (That same user did not had `ALTER` either and when I updated extension, I did received the MySQL error message).

nitriques:
If there was a problem with a SQL statement, the current code you log
the exception, but silence it, resulting in a weird user experience.

This commits make sure to throw any errors that would occur during the
saving process.

Fixes #2752

### DIFF
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -121,6 +121,7 @@ class EntryManager
         }
 
         $did_lock = false;
+        $exception = null;
         try {
 
             // Check if table exists
@@ -163,11 +164,16 @@ class EntryManager
                 Symphony::Database()->insert($fields, $table_name);
             }
         } catch (Exception $ex) {
+            $exception = $ex;
             Symphony::Log()->pushExceptionToLog($ex, true);
         }
 
         if ($did_lock) {
             Symphony::Database()->query('UNLOCK TABLES');
+        }
+
+        if ($exception) {
+            throw $exception;
         }
     }
 


nitriques:
I've been thinking about it and I feel like we should try to display a nicer message to the user, instead of a nasty screen of death.

In the extensions page, when installing or updating, if there is a mysql error, it is shown in a notice. Should we do this here as well ?

michael-e:
I think that the "screen of death" is appropriate for serious problem. :-)

nitriques:
Ok. I might wanna change that for version 3.0.0 tho.

michael-e:
If you change this, think about saving entries from the frontend as well!

symphony/lib/toolkit/class.entrymanager.php
```diff
@@ -163,12 +164,17 @@ protected static function saveFieldData($entry_id, $field_id, $field)
                 Symphony::Database()->insert($fields, $table_name);
             }
         } catch (Exception $ex) {
+            $exception = $ex;
             Symphony::Log()->pushExceptionToLog($ex, true);
         }
 
         if ($did_lock) {
```
brendo: What if this was in a `finally` block?

symphony/lib/toolkit/class.entrymanager.php
```diff
@@ -163,12 +164,17 @@ protected static function saveFieldData($entry_id, $field_id, $field)
...
             Symphony::Database()->query('UNLOCK TABLES');
         }
+
+        if ($exception) {
+            throw $exception;
```
brendo: This would then move into the catch.

michael-e: `finally` is PHP 5.5 and later only.

brendo: Ah, ignore me then. Forgot we're still supporting PHP 5.3 :)

nitriques: Yup @brendo I've made that mistake to often. And `[]` also. Urg!

PR #2758, opened 2017-09-21
add delegates to allow sorting to be manipulated by extensions

jonmifsud:
Sorting has so-far been always managed in config. Given that different users might want to sort by different values this allows extensions to be added so that sorting can be tailored by user if needed.

[Author Sort](https://github.com/mazedigital/author_sort) is an example of an extension which would do user based sorted if required for testing. All data is stored in database rather than in config through the extension, thus could also be useful in environments where one might want the config to be in read-only mode.

### DIFF
--- a/symphony/lib/toolkit/class.section.php
+++ b/symphony/lib/toolkit/class.section.php
@@ -92,7 +92,16 @@ class Section
      */
     public function getSortingField()
     {
-        $result = Symphony::Configuration()->get('section_' . $this->get('handle') . '_sortby', 'sorting');
+        $result = null;
+        
+        Symphony::ExtensionManager()->notifyMembers('getSortingField', '/publish/', array(
+            'section-handle' => $this->get('handle'),
+            'field' => &$result,
+        ));
+
+        if($result == null){
+            $result = Symphony::Configuration()->get('section_' . $this->get('handle') . '_sortby', 'sorting');
+        }
 
         return (is_null($result) ? $this->getDefaultSortingField() : $result);
     }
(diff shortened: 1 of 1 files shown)


symphony/lib/toolkit/class.section.php
```diff
@@ -92,7 +92,16 @@ public function getDefaultSortingField()
      */
     public function getSortingField()
     {
-        $result = Symphony::Configuration()->get('section_' . $this->get('handle') . '_sortby', 'sorting');
+        $result = null;
+        
+        Symphony::ExtensionManager()->notifyMembers('getSortingField', '/publish/', array(
```
nitriques: Can you please add the required php doc for all `notifyMembers` calls ? It is required in order to generate the doc for the website.

symphony/lib/toolkit/class.section.php
```diff
@@ -92,7 +92,16 @@ public function getDefaultSortingField()
...
+            'section-handle' => $this->get('handle'),
+            'field' => &$result,
+        ));
+
+        if($result == null){
```
nitriques: PSR requires a space between `if (` and `) {`

symphony/lib/toolkit/class.section.php
```diff
@@ -123,10 +139,19 @@ public function getSortingOrder()
      */
     public function setSortingField($sort, $write = true)
     {
-        Symphony::Configuration()->set('section_' . $this->get('handle') . '_sortby', $sort, 'sorting');
+        Symphony::ExtensionManager()->notifyMembers('setSortingField', '/publish/', array(
+            'section-handle' => $this->get('handle'),
+            'sort' => $sort,
+            'updated' => $updated,
```
nitriques: 1. I think 'sort' should be 'field' for consistency. I would also simply pass the $order by ref and drop the 'updated' key.

nitriques:
Merged as 3001c91613

nitriques:
BTW @jonmifsud I've changed the delegates name and some parameter names, make sure to update your extension as well !

PR #2760, opened 2017-09-21
Use the full width when the primary col is empty

nitriques:
If a developer creates a section with fields only in the sidebar, the
fields won't have any size since they are float. This yeiled bad UX.

This commits adds a CSS selector for secondary columns that are not preceded
by a primary column.


nitriques:
> Please consider switching to flexbox and grid in the next major release. 

Yup we already did it ;) I was never a float fan (I prefer inline-block).

Thanks guys for all those reviews.

Issue #2711, opened 2017-06-18
[2.7.0RC1] Symphony polluting the logs


michael-e:
RC1 is polluting the log file with the following message:

> Deprecated Warning: The method `Datasource::grab()` is deprecated. It will be removed in Symphony 3.0.0. Please use `Datasource::execute()` instead. Called from `FrontendPage::processDatasources()` in class.frontendpage.php at line 895

If s.th. is deprecated, Symphony itself should not use it, right?

michael-e:
Here's a second one which sounds more severe:

> Deprecated Warning: The use of `SHA1::hash()` is strongly discouraged due to severe security flaws. It will be removed in Symphony 3.0.0. Please use `PBKDF2::hash()` instead. Called from `Author::createAuthToken()` in class.author.php at line 157

nitriques:
> If s.th. is deprecated, Symphony itself should not use it, right?

No, because it break compatibility with older setups.

> Here's a second one which sounds more severe:

Hashing passwords with SHA1 is not considered secure anymore. How come your users never got upgraded to PBKDF2 ?
---

Re #2701

michael-e:
> Hashing passwords with SHA1 is not considered secure anymore. How come your users never got upgraded to PBKDF2 ?

This error occured in a new test setup. I installed Symphony 2.6.11 with the default workspace, then upgraded to Symphony 2.7.0RC1.

nitriques:
Hum I wonder why then... Let me test that.

nitriques:
I've found the problem: The SHA-1 is used to generate the remote login token.

We have 2 options:

1. A more secure one, would be to switch to PBKDF2 token (and BTW make them way bigger).
2. Do not use the SHA1 class and use directly php's function.

Votes are open!

nitriques:
And I get that error with the workspace: The Section, 1, associated with the Data source, <code>Archive</code>, could not be found. on line 482 of file symphony\lib\toolkit\data-sources\class.datasource.section.php

michael-e:
What do you mean by PHP'S function? Something like `password_hash('test', PASSWORD_DEFAULT)`?

Why is the remote login token a hash at all? I would expect it to be "simply random", at the same time as random as possible.

nitriques:
> What do you mean by PHP's function? Something like password_hash('test', PASSWORD_DEFAULT)

Yup. Or `sha1()`, which would prevent the token from changing when updating from lower than 2.7.0

[...]

PR #2762, opened 2017-09-26
Add configurable log filtering

nitriques:
Since this is a LTS version, it seem counter intuitive to log deprecated
calls. We do need the possibility to log them, to help migration
efforts. But since there are a lot of deprecated calls, it then becomes
as performance issue.

Re #2711

### DIFF
--- a/install/includes/config_default.php
+++ b/install/includes/config_default.php
@@ -31,6 +31,7 @@
         'log' => array(
             'archive' => '1',
             'maxsize' => '102400',
+            'filter' => E_ALL ^ E_DEPRECATED,
         ),
         ########
 
--- a/symphony/lib/core/class.log.php
+++ b/symphony/lib/core/class.log.php
@@ -50,6 +50,13 @@ class Log
     private $_archive = false;
 
     /**
+     * The filter applied to logs before they are written.
+     * @since Symphony 2.7.1
+     * @var integer
+     */
+    private $_filter = -1;
+
+    /**
      * The date format that this Log entries will be written as. Defaults to
      * Y/m/d H:i:s.
      * @var string
@@ -119,7 +126,19 @@ class Log
      */
     public function setMaxSize($size)
     {
-        $this->_max_size = $size;
+        $this->_max_size = General::intval($size);
+    }
+
+    /**
+     * Setter for the `$_filter`.
+     *
+     * @since Symphony 2.7.1
+     * @param mixed $filter
+     *  The filter used on log $type parameter.
+     */
+    public function setFilter($filter)
+    {
+        $this->_filter = General::intval($filter);
     }
 
     /**
@@ -185,7 +204,8 @@ class Log
      * @param string $message
      *  The message to add to the Log
      * @param integer $type
-     *  A PHP error constant for this message, defaults to E_NOTICE
+     *  A PHP error constant for this message, defaults to E_NOTICE.
+     *  If null or 0, will be converted to E_ERROR.
      * @param boolean $writeToLog
      *  If set to true, this message will be immediately written to the log. By default
      *  this is set to false, which means that it will only be added to the array ready
@@ -202,6 +222,10 @@ class Log
      */
     public function pushToLog($message, $type = E_NOTICE, $writeToLog = false, $addbreak = true, $append = false)
     {
+        if (!$type) {
+            $type = E_ERROR;
+        }
+
         if ($append) {
             $this->_log[count($this->_log) - 1]['message'] =  $this->_log[count($this->_log) - 1]['message'] . $message;
         } else {
@@ -209,7 +233,7 @@ class Log
             $message = DateTimeObj::get($this->_datetime_format) . ' > ' . $this->__defineNameString($type) . ': ' . $message;
         }
 
-        if ($writeToLog) {
+        if ($writeToLog && ($this->_filter === -1 || ($this->_filter & $type))) {
             return $this->writeToLog($message, $addbreak);
         }
     }
--- a/symphony/lib/core/class.symphony.php
+++ b/symphony/lib/core/class.symphony.php
@@ -228,7 +228,8 @@ abstract class Symphony implements Singleton
 
         self::$Log = new Log($filename);
         self::$Log->setArchive((self::Configuration()->get('archive', 'log') == '1' ? true : false));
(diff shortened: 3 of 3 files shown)


nitriques:
@michael-e If you could spare 2 minutes to review this, I would appreciate it. Thanks.

install/includes/config_default.php
```diff
@@ -31,6 +31,7 @@
         'log' => array(
             'archive' => '1',
             'maxsize' => '102400',
+            'filter' => E_ALL ^ E_DEPRECATED,
```
michael-e: Doesn't this say "all or deprecated", which is the same as "all"? I would expect `E_ALL & ~E_DEPRECATED` instead.

symphony/lib/core/class.symphony.php
```diff
@@ -228,7 +228,8 @@ public static function initialiseLog($filename = null)
 
         self::$Log = new Log($filename);
         self::$Log->setArchive((self::Configuration()->get('archive', 'log') == '1' ? true : false));
-        self::$Log->setMaxSize(intval(self::Configuration()->get('maxsize', 'log')));
```
brendo: Removing this `intval` check assumes that log files will always be written by Symphony, I know I've definitely edited these files through nano or so in the past, so it might be good to keep the training wheels here for safety (given it's LTS).

symphony/lib/core/class.log.php
```diff
@@ -185,7 +204,8 @@ public function popFromLog()
      * @param string $message
      *  The message to add to the Log
      * @param integer $type
-     *  A PHP error constant for this message, defaults to E_NOTICE
+     *  A PHP error constant for this message, defaults to E_NOTICE.
+     *  If null or 0, will be converted to E_ERROR.
```
brendo: This is overcomplicating it for me. What problem was solved with this? I think we should encourage people to specify something, or be happy with the `E_NOTICE` default.

install/includes/config_default.php
brendo: I'm pretty sure it's a XOR, meaning E_ALL or E_DEPRECATED, but not both. I'm not strong with bitwise though, always have to look it up!

symphony/lib/core/class.symphony.php
michael-e: This has been moved inside the `setMaxSize` function.

brendo: Woops, missed that. Disregard! Sorry!

install/includes/config_default.php
michael-e: Ah, thanks.

michael-e: The resulting integer value is the same, BTW.

nitriques: It is all "minus" deprecated. ie (a && !b) || (!a && b) on bits

> The resulting integer value is the same, BTW.

Yes because in that case, it's always (a && !b) that wins!

symphony/lib/core/class.log.php
nitriques: > What problem was solved with this?

Exceptions. By default, they have a code == 0

nitriques: So we have plenty of unknown error in the logs.

nitriques:
> It's a shame that this configuration setting is not consistent with the main error reporting setting. 

I do not agree. It is nice to log them on a production system where you definitively do not want to crash.

>  I noticed that many calls to Log::pushToLog don't use the right function signature

- [ ] Thanks will fix

symphony/lib/core/class.log.php
brendo: Gotcha.

nitriques:
> I noticed that many calls to Log::pushToLog don't use the right function signature

I could not find any. Maybe I missed something @michael-e ?

symphony/lib/core/class.log.php
nitriques: @brendo wanna approve the PR? I like green check marks.

michael-e:
> I could not find any. Maybe I missed something @michael-e ?

I think so. Currently `pushToLog` has 5 parameters. Searching the project for `->pushToLog` you will find many calls using less parameters:

* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/content/content.ajaxlog.php#L15
* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/content/content.publish.php#L504
* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/lib/core/class.errorhandler.php#L274
* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/lib/core/class.errorhandler.php#L455
* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/lib/core/class.log.php#L232
* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/lib/toolkit/class.emailgateway.php#L28
* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/lib/toolkit/class.extensionmanager.php#L1001
* https://github.com/symphonycms/symphony-2/blob/2.7.0/symphony/lib/toolkit/class.lang.php#L282

nitriques:
> I think so. Currently pushToLog has 5 parameters. 

Yeah but only the first one is required. The remaining 4 are all optional (since they have a default value in the function signature), so they are all ok !!

PHP does not let you call function will less then the minimum number of parameters. But it lets you call function on which you pass more arguments then needed.

michael-e:
> The remaining 4 are all optional (since they have a default value…)

My oversight. Sorry, and thanks!

nitriques:
Np ;)

nilshoerrmann:
#2711 was driving me crazy and it tooks a while for me to figure out that there is a new settings to handle these endless entries of deprecation warnings in the logs. 

Symphony doesn't take care of this issue when updating: it neither adds the setting nor informs the user that it exists. Ideally it would do both, add the most common setting and notify the user about the change.

Furthermore, some documentation with common usage examples would be helpful: `E_ALL ^ E_DEPRECATED` is hardcore programmer talk. Personally, I don't know what this line mean exactly so I don't know how to setup this correctly besides copying the the default value :)

nitriques:
Sorry about that @nilshoerrmann ! Would adding the text and the default value for 2.7.8 be reasonable ?

Issue #2763, opened 2017-10-01
SMTP password saving fails


tiloschroeder:
Affected Symphony version(s) :2.6.11, 2.7.0

I've tried to use a password with a "less then" character, e.g. '123<456', in the Symphony Preferences for a SMTP Gateway.  Only '123' will be stored.

michael-e:
Confirmed. This is a bug. (I really wonder why nobody ever experienced it before.)

jurajkapsz:
> (I really wonder why nobody ever experienced it before.)

Interesting, that I never had a `<` symbol in my passwords, not even in generated ones and all my passwords that I use or have ever used are unique.

nitriques:
Yeah wow super find, never had this problem also !

I this is strip_tags that does this. It is really hard to sanitize passwords.

michael-e:
Why does Symphony use `strip_tags` for the Email Gateway preferences?

I tried the same (`<` in a password) for senders in the Email Newsletter Manager (which also uses the email gateways). The "sender" page will break, but just because the email gateways don't sanitize the values when building the input widgets.

So my first suggestion is to sanitize the values when buikding the `Widget::Input` stuff. This would at least fix issues with the Email Newsletter Manager. Then one could try and fix Symphony's preferences.

@nitriques: I can send a PR for the sanitizing, if you like. Should I?

nitriques:
> @nitriques: I can send a PR for the sanitizing, if you like. Should I?

Yes please.

We should not touch the password when saving it (only `'` escaping is needed).

BTW, returning passwords to the user is kind of a non sense to me.

michael-e:
My pull request #2769 prevents invalid HTML if any value contains a `<` character, for example. All values are sanitized (which is good practice, IMHO). However, it does not fully solve the issue in this thread. The problem that is left is **saving** the values.

nitriques:
Fixed in #2781 which got merged into `2.7.x` branch

PR #2769, opened 2017-10-25
Sanitize email gateway values (#2763)

michael-e:
Sanitizing prevents invalid HTML output of email gateways in the Symphony backend.

### DIFF
--- a/symphony/content/content.systempreferences.php
+++ b/symphony/content/content.systempreferences.php
@@ -198,7 +198,12 @@ class contentSystemPreferences extends AdministrationPage
         Symphony::ExtensionManager()->notifyMembers('CustomActions', '/system/preferences/');
 
         if (isset($_POST['action']['save'])) {
+            // Sanitize all values
             $settings = filter_var_array($_POST['settings'], FILTER_SANITIZE_STRING);
+            // Reset (un-sanitize) the password of the SMTP email gateway. RE: #2763
+            if (isset($_POST['settings']['email_smtp']['password'])) {
+                $settings['email_smtp']['password'] = $_POST['settings']['email_smtp']['password'];
+            }
 
             /**
              * Just prior to saving the preferences and writing them to the `CONFIG`
--- a/symphony/lib/toolkit/email-gateways/email.sendmail.php
+++ b/symphony/lib/toolkit/email-gateways/email.sendmail.php
@@ -167,12 +167,12 @@ class SendmailGateway extends EmailGateway
 
         $label = Widget::Label(__('From Name'));
         $label->setAttribute('class', 'column');
-        $label->appendChild(Widget::Input('settings[email_sendmail][from_name]', $this->_sender_name, 'text', $readonly));
+        $label->appendChild(Widget::Input('settings[email_sendmail][from_name]', General::sanitize($this->_sender_name), 'text', $readonly));
         $div->appendChild($label);
 
         $label = Widget::Label(__('From Email Address'));
         $label->setAttribute('class', 'column');
-        $label->appendChild(Widget::Input('settings[email_sendmail][from_address]', $this->_sender_email_address, 'text', $readonly));
+        $label->appendChild(Widget::Input('settings[email_sendmail][from_address]', General::sanitize($this->_sender_email_address), 'text', $readonly));
         $div->appendChild($label);
 
         $group->appendChild($div);
--- a/symphony/lib/toolkit/email-gateways/email.smtp.php
+++ b/symphony/lib/toolkit/email-gateways/email.smtp.php
@@ -341,7 +341,7 @@ class SMTPGateway extends EmailGateway
         $readonly = array('readonly' => 'readonly');
 
         $label = Widget::Label(__('HELO Hostname'));
-        $label->appendChild(Widget::Input('settings[email_smtp][helo_hostname]', $this->_helo_hostname, 'text', $readonly));
+        $label->appendChild(Widget::Input('settings[email_smtp][helo_hostname]', General::sanitize($this->_helo_hostname), 'text', $readonly));
         $div->appendChild($label);
 
         $group->appendChild($div);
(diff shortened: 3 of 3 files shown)


nitriques:
Thanks @michael-e this is better than nothing. I confirm that sanitize does not seem to break the password.

I also think that Widget should auto-sanitize things, but that's an other sorry (and there a lots of hacks that would break). Should we open an issue for this ?

---

As discussed in #2763, the REAL fix would be to stop sending clear text password in responses. Even for sites with https. We could use the empty value or 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' as a place holder, detect it, and revert back to the old value on save, while keeping the sanitize call. Should we open a new issue for this ?

Since that would be a breaking change, we still need a better fix for the LTS release line.

---

Luckly, I DID found the culprit, i.e. FILTER_SANITIZE_STRING, which does as advertised:

> Strip tags, optionally strip or encode special characters.

So, should we "capture" password before this ? I mean, it is still valid for all other inputs.

---

Here's my test script

```php
<?php

function sanitize($s)
{
	return htmlspecialchars($s, ENT_COMPAT, 'UTF-8', false);
}
function filter($s)
{
	return filter_var_array([$s], FILTER_SANITIZE_STRING);
}
$p = 'test<test';
var_dump(sanitize($p));
var_dump(sanitize(sanitize($p)));
var_dump(filter($p));
```

michael-e:
We have three things now:

### Widgets and sanitizing

> I also think that Widget should auto-sanitize things, but that's an other sorry (and there a lots of hacks that would break). Should we open an issue for this ?

Yes, the Widget methods should sanitize the input. So you really think it would break "hacks"? At least sanitizing too often doesn't do any harm (see your test script, for example).

I will open an issue for further discussion.


### Stop sending clear test passwords

> As discussed in #2763, the REAL fix would be to stop sending clear text password in responses.

No, it wouldn't help at all, because the issue arises during saving. And as you said, it would be a breaking change. Feel free to open an issue as a reminder (for later Symphony versions).


### Stripping characters when saving

> Luckly, I DID found the culprit, i.e. FILTER_SANITIZE_STRING, which does as advertised

Thanks! Indeed all posted values are filtered, which has been introduced by @brendo [two years ago](https://github.com/symphonycms/symphony-2/commit/651e150091c61fb60ad1dff2bc2166185a83d9d6).

I am not sure if being so strict about these values is actually necessary, but anyway — nobody ever complained (except for the password thingie). So yes, "capturing" the password would be cool. Something like this:

```php
// Sanitize all values
$settings = filter_var_array($_POST['settings'], FILTER_SANITIZE_STRING);

// The password is special
if (isset($_POST['settings']['email_smtp']['password'])) {
    $settings['email_smtp']['password'] = $_POST['settings']['email_smtp']['password'];
}
```


I just don't know if and how the password itself should be sanitized (inside the `if` condition)...

nitriques:
> I just don't know if and how the password itself should be sanitized (inside the if condition)...

We don't: simply escape slashes. What else could break a php string ?

> So you really think it would break "hacks"?

Not for LTS, but for 3.0.0 or more yes.

> Something like this:

Maybe more like this?

```php
foreach ($email_gateways as $gateway) {
    if (isset($_POST['settings'][$gateway]['password'])) {
        $settings[$gateway]['password'] = $_POST['settings'][$gateway]['password'];
    }
}
```

It solves the bug, but it's clearly a hack and should be revisited. I am also not aware of any other password field tho.

michael-e:
Should I add a commit to my pull request? Or do you want to care for the password logic yourself?

nitriques:
@michael-e Please add a commit ! We'll discuss it ;)

But I wonder if we should but use strip_tags instead... see http://php.net/manual/en/filter.filters.sanitize.php#118186

michael-e:
The comment you linked to is actually misleading. The behaviour only differs when there is a whitespace after the `<`. I other cases, `<` is stripped anyway.

Honestly, I don't know why we sanitize the strings (i.e. strip tags) at all. @brendo, do you remember why you [introduced this](651e150091c61fb60ad1dff2bc2166185a83d9d6)?

I will add a commit as discussed above, resetting the password value for email gateways to the original string.

michael-e:
Unfortunately it seems impossible to loop over all email gateways [as proposed](#issuecomment-339841537), because the array keys are hardcoded into the gateway classes, and there is no API to read them. Although there seems to be a "silent convention" (the Sendmail gateway uses `email_sendmail`, the SMTP gateway uses `email_smtp`), this is nothing we should rely on IMHO.

So all we can do is reset the password for the the SMTP gateway only:

```php
if (isset($_POST['settings']['email_smtp']['password'])) {
    $settings['email_smtp']['password'] = $_POST['settings']['email_smtp']['password'];
}
```

This is not nice. But I don't see a better fix as long as we sanitize all values before saving (see above).

nitriques:
> The comment you linked to is actually misleading.

Damn.

> Unfortunately it seems impossible to loop over all email gateways as proposed,

Double damn.

> I don't know why we sanitize the strings

Me neither. Looks like @brendo was trying to fix XSS, but did it in the wrong place.

2.7.x takes care of serializing the config properly. We just need to make sure that when output in html, values are sanitized.

michael-e:
> 2.7.x takes care of serializing the config properly. We just need to make sure that when output in html, values are sanitized.

So you think that the "sanitizing when saving the prefs" should be removed altogether?

nitriques:
> So you think that the "sanitizing when saving the prefs" should be removed altogether?

Yeah, that's my guess... I think @brendo wanted to fix from injection problems, but fix it at the wrong place. One might actually WANT html to be saved in the config file...

nitriques:
But we would also need to make sure `sanitize` is called EVERYWHERE we output config value in the html.

michael-e:
> But we would also need to make sure sanitize is called EVERYWHERE we output config value in the html

This is hard to ensure, isn't it?

nitriques:
No indeed. But it's a manual, error-prone process.

michael-e:
I would really appreciate this, but I can't do it. If you don't find the time either, we should merge the current pull request (which fixes the email password issue) — and maybe open a new issue as a reminder for later.

nitriques:
I'll try to do it ASAP.

nitriques:
See #2781

PR #2772, opened 2017-11-02
Prevent robots from indexing login and install

nitriques:
The title says it all!


nitriques:
> Gee, I'd hope people are removing these prior to deployment!

Me too! But you never know...

PR #2773, opened 2017-11-07
Add alter and xml escaping of leading digits

### LINKED ISSUE #2765: Include a count of entries in associated sections doesn't work when a section starts with a number
Affected Symphony version(s) : 2.7.0 and previous versions?
PHP version(s) : 5.6
OS(es) : OS X

Include a count of entries in associated sections doesn't work when the associated section starts with a number.

```  
error parsing attribute name in Entity, line: 894
/data/sites/web/moovne/www/symphony/lib/toolkit/class.xsltprocess.php:141
```

nitriques:
Fixes #2765 for the LTS release.

In next release, leading digit should be disallowed.

### DIFF
--- a/symphony/lib/toolkit/class.datasource.php
+++ b/symphony/lib/toolkit/class.datasource.php
@@ -180,7 +180,7 @@ class Datasource
      */
     public function execute(array &$param_pool = null)
     {
-        $result = new XMLElement($this->dsParamROOTELEMENT);
+        $result = new XMLElement($this->escapeFirstCharIfDigit($this->dsParamROOTELEMENT));
 
         try {
             $result = $this->execute($param_pool);
@@ -264,7 +264,7 @@ class Datasource
     public function emptyXMLSet(XMLElement $xml = null)
     {
         if (is_null($xml)) {
-            $xml = new XMLElement($this->dsParamROOTELEMENT);
+            $xml = new XMLElement($this->escapeFirstCharIfDigit($this->dsParamROOTELEMENT));
         }
 
         $xml->appendChild($this->__noRecordsFound());
@@ -284,7 +284,7 @@ class Datasource
     public function negateXMLSet(XMLElement $xml = null)
     {
         if (is_null($xml)) {
-            $xml = new XMLElement($this->dsParamROOTELEMENT);
+            $xml = new XMLElement($this->escapeFirstCharIfDigit($this->dsParamROOTELEMENT));
         }
 
         $xml->appendChild($this->__negateResult());
@@ -534,6 +534,38 @@ class Datasource
     public static function removeEscapedCommas($string)
     {
         return preg_replace('/(?<!\\\\)\\\\,/', ',', $string);
+    }
+
+    private static $digitReplacements = array(
+        '1' => 'one',
+        '2' => 'two',
+        '3' => 'three',
+        '4' => 'four',
+        '5' => 'five',
+        '6' => 'six',
+        '7' => 'seven',
+        '8' => 'eight',
+        '9' => 'nine',
+        '0' => 'zero',
+    );
+    /**
+     * XML does not allow tags or attribute to begin with a number.
+     * This function will replace the leading digit with its english equivalent.
+     *
+     * @param string $value
+     * @return string
+     *  The escaped name
+     */
+    public function escapeFirstCharIfDigit($value)
+    {
+        foreach (static::$digitReplacements as $d => $r) {
+            if (strpos($value, $d) === 0) {
+                $value = substr($value, 1);
+                $value = lcfirst($value);
+                return "$r-$value";
+            }
+        }
+        return $value;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


symphony/lib/toolkit/class.datasource.php
```diff
@@ -536,6 +536,38 @@ public static function removeEscapedCommas($string)
...
+        '1' => 'one',
+        '2' => 'two',
+        '3' => 'three',
+        '4' => 'four',
+        '5' => 'five',
+        '6' => 'six',
+        '7' => 'seven',
+        '8' => 'eight',
+        '9' => 'nine',
+        '0' => 'zero',
```
michael-e: Although PHP (still) allows this, we shouldn't have a trailing comma on the last array element. It's just nothing we should encorage, IMHO.

symphony/lib/toolkit/class.datasource.php
```diff
@@ -536,6 +536,38 @@ public static function removeEscapedCommas($string)
...
+     * @param string $value
+     * @return string
+     *  The escaped name
+     */
+    public function escapeFirstCharIfDigit($value)
+    {
+        foreach (static::$digitReplacements as $d => $r) {
+            if (strpos($value, $d) === 0) {
+                $value = substr($value, 1);
+                $value = lcfirst($value);
```
michael-e: Why should this function deal with upper/lower case letters? That's literally "out if scope" IMHO.

symphony/lib/toolkit/class.datasource.php
froschdesign: @michael-e 
> lthough PHP (still) allows this

Why "still"? More is coming: https://wiki.php.net/rfc/trailing-comma-function-calls 😉

symphony/lib/toolkit/class.datasource.php
```diff
@@ -536,6 +536,38 @@ public static function removeEscapedCommas($string)
         return preg_replace('/(?<!\\\\)\\\\,/', ',', $string);
     }
 
+    private static $digitReplacements = array(
```
froschdesign: `static` is not needed, because it's only used in the method `escapeFirstCharIfDigit`. Therefore is `static` also wrong here!

symphony/lib/toolkit/class.datasource.php
michael-e: Oh. :-(

So what is "good practice" then? I haven't found anything in the PSR-2 coding style guide.

froschdesign: > I haven't found anything in the PSR-2 coding style guide.

Right, it's not included.

> So what is "good practice" then?

It's included in the coding standards of many frameworks and IDE's also support automatic adding.

michael-e: Sorry if my question was unclear. I'd like to know which style — trailing comma or not — is the preferred version in PHP and/or Symphony. I know that both styles work, but we should have a preferred one.

froschdesign: > The comma after the last array element is optional and can be omitted. This is usually done for single-line arrays, i.e. array(1, 2) is preferred over array(1, 2, ). For multi-line arrays on the other hand the trailing comma is commonly used, as it allows easier addition of new elements at the end.

http://de2.php.net/manual/en/language.types.array.php

> Add a comma after each array item in a multi-line array, even after the last one;

https://symfony.com/doc/current/contributing/code/standards.html#structure

> ZendCodingStandard.Arrays.TrailingArrayComma (requires comma after last element in multiline arrays)

https://github.com/zendframework/zend-coding-standard/pull/1

And I prefer the trailing comma! 😄

nitriques: > It's just nothing we should encorage, IMHO.

I like it. It makes diff of adding/removing element a lot nicer.

We should also encourage early returns.

symphony/lib/toolkit/class.datasource.php
nitriques: static here saves memory. We de create a lot of DataSource object and do not need multiple copies of the data.

symphony/lib/toolkit/class.datasource.php
nitriques: Hum I think this is an error.

nitriques:
I kind of agree with @michael-e.

I thought it looked cool on the night I've made it. But I think it is wack now.

I like the reaction to it tho ¯\_(ツ)_/¯

symphony/lib/toolkit/class.datasource.php
froschdesign: @nitriques 
> static here saves memory.

Do you have some benchmarks, which illustrates the problem? I found nothing in this repository.
If it is only your feeling, then remove the `static`. 😉

nitriques: > Do you have some benchmarks, which illustrates the problem?

No, but I am pretty confident ;)

PR #2774, opened 2017-11-08
Issue 2765 for 2.7.x

nitriques:
Fix for #2765
Re #2773

Two things:

- Disallow data source that starts with a digit.
- Prepend `x-` to section names that begins with a digit.

Both situation crashes the whole frontend and are considered a bug fixes.

I think sections name/events beginning with a digit should be avoided and disallowed in 3.0.0

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -1168,6 +1168,8 @@ class contentBlueprintsDatasources extends ResourcesPage
             $this->_errors['name'] = __('This is a required field');
         } elseif (strpos($fields['name'], '\\') !== false) {
             $this->_errors['name'] = __('This field contains invalid characters') . ' (\\)';
+        } elseif (!preg_match('/^[a-z]/i', $fields['name'])) {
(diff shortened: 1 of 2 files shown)


symphony/content/content.blueprintsdatasources.php
```diff
@@ -1168,6 +1168,8 @@ public function __formAction()
             $this->_errors['name'] = __('This is a required field');
         } elseif (strpos($fields['name'], '\\') !== false) {
             $this->_errors['name'] = __('This field contains invalid characters') . ' (\\)';
+        } elseif (!preg_match('/^[a-z]/i', $fields['name'])) {
+            $this->_errors['name'] = __('The name of the data-source must begin with a letter.');
```
michael-e: Most of the time it's written "data source" (unhyphenated) in Symphony.

nitriques: Good catch!

nitriques:
@michael-e is it ok now ? I have a branch with only typo fixes in there too, which normalizes the spelling across the codebase.

PR #2775, opened 2017-11-08
Call addslashes on configuration group

nitriques:
This is needed because if a group containing a ' is created it will
create a syntax error in our config file


nitriques:
> Assume that we already handle removing the slashes when reading the configuration?

This is done by PHP itself when it parses the config file.

PR #2778, opened 2017-11-10
Restrict CSS resize to vertical on <textarea>

cylkee:
Enabled by default in nearly all browsers, `<textarea>` form fields can be resized.  This rule restricts that resize ability to `vertical`, to help this browser feature play nice with Symphony's column layout.  [CSS resize support](https://caniuse.com/#feat=css-resize) is good, except for IE, naturally!


nitriques:
LGTM ! Waiting for more reviewers before merging.

PR #2781, opened 2017-11-24
Remove automatic escaping of preferences values

### LINKED ISSUE #2763: SMTP password saving fails

nitriques:
With this change, preference values are now not escaped on POST.
They are already escaped properly when saving the config and this PR adds output escaping to preferences controlled in the core.

Fixes #2763
Closes #2769

### DIFF
--- a/symphony/content/content.systempreferences.php
+++ b/symphony/content/content.systempreferences.php
@@ -198,8 +198,7 @@ class contentSystemPreferences extends AdministrationPage
         Symphony::ExtensionManager()->notifyMembers('CustomActions', '/system/preferences/');
 
         if (isset($_POST['action']['save'])) {
-            $settings = filter_var_array($_POST['settings'], FILTER_SANITIZE_STRING);
-
+            $settings = filter_var_array($_POST['settings'], FILTER_UNSAFE_RAW, FILTER_FLAG_STRIP_LOW);
             /**
              * Just prior to saving the preferences and writing them to the `CONFIG`
              * Allows extensions to preform custom validation logic on the settings.
--- a/symphony/lib/toolkit/email-gateways/email.sendmail.php
+++ b/symphony/lib/toolkit/email-gateways/email.sendmail.php
@@ -167,12 +167,12 @@ class SendmailGateway extends EmailGateway
 
         $label = Widget::Label(__('From Name'));
         $label->setAttribute('class', 'column');
-        $label->appendChild(Widget::Input('settings[email_sendmail][from_name]', $this->_sender_name, 'text', $readonly));
+        $label->appendChild(Widget::Input('settings[email_sendmail][from_name]', General::sanitize($this->_sender_name), 'text', $readonly));
         $div->appendChild($label);
 
         $label = Widget::Label(__('From Email Address'));
         $label->setAttribute('class', 'column');
-        $label->appendChild(Widget::Input('settings[email_sendmail][from_address]', $this->_sender_email_address, 'text', $readonly));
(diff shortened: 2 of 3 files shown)


nitriques:
@michael-e So we then have no-way of fixing this problem in the LTS release ?

nitriques:
> do what you do in this pull request — breaking change

I do not agree that it is a breaking change. Values that were saved with the old behaviour will still be saved with the new code. I understand that it may brings a new XSS vector in extension, which should be fixed. 

Maybe @jensscherbl could help us classify this better. I feel ok merging this into the LTS release.

nitriques:
@michael-e Do you still oppose to this getting merged in 2.7.x ?

michael-e:
Unfortunately @jensscherbl has not joined the discussion. I personally still feel that this is a breaking change. But: It might be wise to accept that semantic versioning is a great idea that sometimes can not be followed without side effects and headache. :-)

I agree that this would improve the LTS version. At the same time we would have to open issues for all extensions that use the Symphony prefs, correct?

BTW: How severe is the attack vector at all? It's just an "internal backend attack", right?

nitriques:
> How severe is the attack vector at all? It's just an "internal backend attack", right?

Yes. You can only XSS yourself as long as the frontend is not allowed to set values in the config.php file, which it should never do.

I would do a thorough audit of extension I use to make sure they are patched.

>  It might be wise to accept that semantic versioning is a great idea that sometimes can not be followed without side effects and headache. :-)

Indeed.

michael-e:
OK, so I don't oppose anymore. Go ahead!

nitriques:
Great! I'll make sure to patch all extensions before merging this in.

michael-e:
What about extension compatibility? Do we have to be careful here? (I mean: What happens if output escaping is performed twice, because the new extension version is used with an older Symphony version?)

nitriques:
I would make them 2,7,1+ only to be sure.

michael-e:
Sounds good.

nitriques:
ETA:

- [ ] JIT Image Manipulation  (PR open)
- [ ] Anti Brute Force
- [ ] Members (maybe)

nitriques:
@michael-e I've fixed or open a PR for all extensions. Can you please review them ? I'll deal with the merge and release. Thanks!

michael-e:
Done. I requested to change compatibility information as disussed above, because I am scared of what will happen if values are escaped twice.

michael-e:
OK, I did a quick test. It seems that no double-encoding takes place, because `General::sanitize` uses

```php
htmlspecialchars($test, ENT_COMPAT, 'UTF-8', false)
```

so the last parameter (`$double_encode`) is set to false.

So extensions should indeed be backward-compatible.

nitriques:
> So extensions should indeed be backward-compatible.

Great! Let me fix the PRs

nitriques:
There is only https://github.com/symphonycms/jit_image_manipulation/pull/161 and https://github.com/symphonists/documenter/pull/29 left to review ! I've made the changes.

michael-e:
I found another two extensions which add HTML to the Preferences page:

* [ ] Maintenance Mode (https://github.com/symphonycms/maintenance_mode)
* [ ] Entity Diagram (https://github.com/symphonists/entity_diagram)

nitriques:
https://github.com/symphonists/entity_diagram/releases/tag/1.5.3 and https://github.com/symphonycms/maintenance_mode/releases/tag/1.9.4

Thanks!

PR #2784, opened 2017-12-13
Remove session.save_handler from php settings

### LINKED ISSUE #2783: PHP 7.2 & session save handler.
Affected Symphony version(s) : 2.7.x latest
PHP version(s) : 7.2.0RC6
OS(es) : Ubuntu 17.10

Trying to use Symphony with PHP 7.2.0 gives the following error:

`ini_set(): Cannot set 'user' save handler by ini_set() or session_module_name()`

I don't have much understanding of what is going on.

screenshot-2017-11-27 symphony error

nitriques:
This has never been required in PHP 5.6 and causes a fatal crash with PHP 7.2

Fixes #2783

### DIFF
--- a/symphony/lib/core/class.session.php
+++ b/symphony/lib/core/class.session.php
@@ -57,7 +57,6 @@ class Session
             }
 
             if (session_id() == '') {
-                ini_set('session.save_handler', 'user');
                 ini_set('session.use_trans_sid', '0');
                 ini_set('session.use_strict_mode', '1');
                 ini_set('session.use_only_cookies', '1');


brendo:
Interesting, what makes you say it's not required by PHP 5.6?

From the [docs](http://php.net/manual/en/session.configuration.php#ini.session.save-handler) the default value is _files_, so removing it may cause different behaviour? It goes on to say you can use `session_set_save_handler` (which we do), but I'm not clear it that has the same effect as setting `save-handler` to `user`.

petertron:
@brendo I can tell you from my own test that the `read` and `write` functions within Symphony's `Session` class will be invoked without having to use `ini_set` and the `user` option, and this is with PHP 5.6.

brendo:
Perfect, thanks @petertron!

nitriques:
@brendo the docs says nothing about 'user' ;) And it looks like php sets it to user whenever `session_set_save_handler` is called https://github.com/php/php-src/blob/83e495e0fdc0c59b449bd173d0c8df1999239634/ext/session/session.c#L1942

PR #2787, opened 2018-01-19
Fix field duplication without field settings

### LINKED ISSUE #2736: Field gets repeated with default settings on save of a section
Affected Symphony version(s) : 2.7.0
PHP version(s) : 5.6.31
OS(es) : MacOS

This is a really strange one, which I've come across twice now so I thought I should log it incase I'm not the only one. Only problem is I can't replicate it or find a pattern to it.

Issue: When creating or editing a section, on save a field is duplicated. I've experienced it with both the Expression Field, and the Textbox Field which is why I'm posting the bug here instead of with the extension. 

The original field is saved with all the parameters fine, the duplicated field has the same label assigned to it but default field parameters. 

Then when I try to edit the duplicated field, it doesn't retain any of the changes. If I remove it, it remains on save until you refresh the browser.

Like I said, it's a weird one which I can't find a pattern to for testing :(

Anyone else experienced this? I would bet on it being an issue with the duplicator.

nitriques:
Some fields do not create any records in their settings table because they do not
implement a proper `Field::commit()` method.
The base implementation of the commit function only deals with the "core"
`tbl_fields` table.
The problem with this approach is that it can lead to data corruption when
saving a field that got deleted by another user.
The only way a field can live without a commit method is if it does not store any
settings at all.
But current version of Symphony assume that the `tbl_fields_$handle` table exists
with at least a `id` and `field_id` column, so field are required to at least create
the table to make their field work without SQL errors from the core.

Fixes #2736

cc @michael-e @twiro 

This patch works when

1. Instance of the field exists *without* entries in the setting table
2. Instance of the field exists *with* entries in the setting table

So migration should always work, regardless. WDYT ?

### DIFF
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1810,7 +1810,13 @@ class Field
      * Checks that we are working with a valid field handle and field id, and
      * checks that the field record exists in the settings table.
      *
+     * @since Symphony 2.7.1 It does check if the settings table only contains
+     *   default columns and assume those fields do not require a record in the settings table.
+     *   When this situation is detected the field is considered as valid even if no records were
+     *   found in the settings table.
+     *
      * @since Symphony 2.7.0
+     * @see Field::tableExists()
      * @return boolean
      *   true if the field id exists in the table, false otherwise
      */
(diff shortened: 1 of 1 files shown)


michael-e:
Please give me one or two days. I have to find the time to dive into it.

nitriques:
No problems @michael-e 

I tested different approaches and this one was the

1. easiest to test
2. smallest
3. easiest to get rid of in next versions
4. impact on only one file

Other possibilities may look less hackish, but did not add all those points.

nitriques:
If @twiro could test it also, it would be MUCH appreciated ;)

nitriques:
@michael-e I feel pretty confident about it, so if you do not have the time to check it, I would go ahead with the merge and the 2.7.1 release.

twiro:
Sorry for the late feedback, I hope I'll find the time to give it a go tomorrow…

nitriques:
Damn @michael-e hope you'll get better soon. Sending as many good vibes as I can.

michael-e:
Thanks a lot!

nitriques:
@twiro I'll release this today, open a new issue if there are any bugs with it.

Issue #2792, opened 2018-02-11
Profiler delivers wrong "zero" durations under certain conditions


michael-e:
Affected Symphony version(s) : all
PHP version(s) : all
OS(es) : all

Symphony fills a `Profiler::$_samples` array during page build. (This array is then used by the Profiler extension in order to display the information to the developer.)

Under certain conditions I experienced wrong "zero" durations in certain samples, and finally I found the reason. It will probably be more clear if you look at the fixes that I will send, but I will try to explain anyway:

The `Profiler::sample` function can be called with a type of `PROFILE_RUNNING_TOTAL` (which is the function's default) or with `PROFILE_LAP`. If the latter is used, the function will call `Profiler::retrieveLast()` in order to find the last sample element and calculate a duration from that point in time. This is dangerous if you don't know which is the last element, of course. If, for example, delegate calls have added sample elements in the meantime, you may refer to a an unexpected sample.

Using `PROFILE_LAP` is only reliable if you are really sure about the last sample.

Another issue (see the second commit in my pull request): `Profiler::retrieveTotalRunningTime` shouldn't use `Profiler::retrieveLast()`, because again we can't be sure which sample this will be. Furtunately, retrieving the total running time can be achieved in a simler and more reliable way.

nitriques:
Closed in  e80ac53

PR #2793, opened 2018-02-11
Fix profiler samples

michael-e:
Fix for #2792

### DIFF
--- a/symphony/lib/toolkit/class.frontendpage.php
+++ b/symphony/lib/toolkit/class.frontendpage.php
@@ -257,6 +257,8 @@ class FrontendPage extends XSLTPage
             $output = parent::generate();
             $this->_param = $backup_param;
 
+            Symphony::Profiler()->sample('XSLT Transformation', PROFILE_LAP);
+
             /**
              * Immediately after generating the page. Provided with string containing page source
              * @delegate FrontendOutputPostGenerate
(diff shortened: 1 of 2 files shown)


michael-e:
At first sight the whole thing is a mess. For example, "XML built" is just a part of the "Page built" process, so it's not calculated from the start but as an own "slice" duration. It's similar with "XML generation" and "XSLT transformation" — both are calculated as slices. And even "Page built" is just a slice of "Page creation complete". 

I never understood the meaning of those numbers until I read the code. But I am hesitant about how to improve this.

It seems much simpler to save the "time from the start" with every sample and do all the math later. But this will probably mean more samples than we have now (because we will need samples for the start as well as the end of certain processes, e.g. a single data source).

So we have to be careful. It may be best to continue collecting two kinds of samples (like we do now):

* Durations from the start
* Slice durations (for datasources, for example)

I have never seen any decent definition of the Profiler values. If we had one, it would be much easier to talk about the inner workings of profiling. In simple words: It's hard to say how to do it right if you don't know what you want to achieve. :-)

nitriques:
> At first sight the whole thing is a mess. 

Indeed.

> So we have to be careful. It may be best to continue collecting two kinds of samples (like we do now):

> - Durations from the start
> - Slice duration (for datasources, for example)

Yes. I think I could build something easy to use AND precise ;) I'll sleep on this.

PR #2801, opened 2018-02-13
Make the default max log size to 10Mb

### LINKED ISSUE #2757: Make the default log size bigger
Affected Symphony version(s) : 3.0.0, possibly 2.7.1

We now log more things, so the default log size should be increased to prevnent

1) loss of information
2) possible errors when unlinking the log file

cc @wdebusschere

nitriques:
This is to prevent over rotating the logs

Fixes #2757

---

Should we update the value if the default is configured when updating ?

### DIFF
--- a/install/includes/config_default.php
+++ b/install/includes/config_default.php
@@ -30,7 +30,7 @@
         ###### LOG ######
         'log' => array(
             'archive' => '1',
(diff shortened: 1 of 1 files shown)


nitriques:
Cool ! Thanks guys !

PR #2802, opened 2018-02-14
Properly set and pass the new author's id

nitriques:
When we commit a new author, the new id was not set on the Author
object.
This prevented delegates to run based on the new id in the PostAuthorCreate.

Also, a direct reference to the author id has been added in the context
of the PostAuthorCreate delegate.

### DIFF
--- a/symphony/content/content.systemauthors.php
+++ b/symphony/content/content.systemauthors.php
@@ -615,6 +615,8 @@ class contentSystemAuthors extends AdministrationPage
                      * '/system/authors/'
                      * @param Author $author
                      *  The Author object that has just been created
+                     * @param integer $author_id
+                     *  The ID of Author ID that was just created
                      * @param array $fields
                      *  The POST fields
                      *  This parameter is available @since Symphony 2.7.0
@@ -625,6 +627,7 @@ class contentSystemAuthors extends AdministrationPage
                      */
                     Symphony::ExtensionManager()->notifyMembers('AuthorPostCreate', '/system/authors/', array(
                         'author' => $this->_Author,
+                        'author_id' => $author_id,
                         'field' => $fields,
                         'errors' => &$this->_errors,
                     ));
--- a/symphony/lib/toolkit/class.author.php
+++ b/symphony/lib/toolkit/class.author.php
@@ -272,6 +272,7 @@ class Author
      * `$this->_fields` values and adds them to the database using either the
      * `AuthorManager::edit` or `AuthorManager::add` functions. An
      * existing user is determined by if an ID is already set.
+     * When the database is updated successfully, the id of the author is set.
      *
      * @see toolkit.AuthorManager#add()
      * @see toolkit.AuthorManager#edit()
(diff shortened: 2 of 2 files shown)


symphony/lib/toolkit/class.author.php
```diff
@@ -292,7 +293,11 @@ public function commit()
                 return false;
             }
         } else {
-            return AuthorManager::add($this->get());
+            $id = AuthorManager::add($this->get());
+            if ($id) {
+                $this->set('id', $id);
+            }
+            return $id;
         }
```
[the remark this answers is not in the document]
nitriques: >  So setting the author ID on the author object has no effect on AuthorPostCreate at all.

Well, if you use the author object in post create, it does not have an author id set on it... I think that's a problem. But yeah, I've seen that we were also passing the 'author_id' key in the context elsewhere, so I've added it also. 

> At this point I just think that it is useless for what you want to achieve.

Generally true. But would you oppose this merge ?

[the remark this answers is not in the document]
michael-e: No, I don't oppose this merge. You are right that it's generally good to have a "complete" author object, i.e. including the ID.

symphony/content/content.systemauthors.php
```diff
@@ -625,6 +627,7 @@ public function __actionNew()
                      */
                     Symphony::ExtensionManager()->notifyMembers('AuthorPostCreate', '/system/authors/', array(
                         'author' => $this->_Author,
+                        'author_id' => $author_id,
```
brendo: Is this still required with the changes to `Author::commit`? `$this->_Author` will now include the ID.

nitriques: No it is not.

This is only to make it possible to reuse the same code as other delegates. I often find myself reusing the same function for multiple delegates and not having to look into the $context is a good win.

symphony/lib/toolkit/class.author.php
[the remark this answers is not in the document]
nitriques: Great!

symphony/content/content.systemauthors.php
michael-e: On the other hand, doesn't a 3.0 release allow to clean stuff up? So what about removing the ID from all delegates? :-)

brendo: Ah consistency makes sense!

nitriques: @michael-e I thought about it, and it felt weird when deleting the author: do we really want to load an Author object from the DB before deleting it ? If so, I would not mind removing the 'author_id'!

michael-e: Talking about consistency… :-)  …take a look at `content.publish.php`, for example (and there are more examples, IMHO). There you will find the same old scheme: create and edit delegates provide the object, but the delete delegate provides the ID only. 

So changing the author class only is actually bad if we talk about consistency. Shoudn't we have comparable delegates for authors, sections, entries etc.?

nitriques: > Shouldn't we have comparable delegates for authors, sections, entries etc.?

I think we should. Since they were all added in a "ad-hoc" fashion, simply filling the goal of the original developer.

PR #2804, opened 2018-02-16
Remove parameters as arguments

### LINKED ISSUE #2338: Parameters as arguments (using $param in XSLT without declaring it) should be removed
There has been a [discussion about best practice](https://github.com/symphonycms/symphony-2/issues/2141) in XSLT templating regarding parameters some time ago. In this discussion, @allen [explained](https://github.com/symphonycms/symphony-2/issues/2141#issuecomment-50145118) the technical concepts and [suggested to remove "parameters as arguments"](https://github.com/symphonycms/symphony-2/issues/2141#issuecomment-50146270) from Symphony.

I am all for it. Clean that stuff up, finally. Removing `$param` type parameters will, of course, break XSLT stylesheets that do not yet adhere to the "good practice", but developers can fix this easily.

Related: https://github.com/symphonycms/symphony-2/issues/2332#issuecomment-72176600

nitriques:
i.e. using $param in XSLT without declaring it.

Fixes #2338

### DIFF
--- a/symphony/lib/toolkit/class.frontendpage.php
+++ b/symphony/lib/toolkit/class.frontendpage.php
@@ -539,7 +539,6 @@ class FrontendPage extends XSLTPage
                '</xsl:stylesheet>';
 
         $this->setXSL($xsl, false);
-        $this->setRuntimeParam($this->_param);
 
         Symphony::Profiler()->seed($start);
         Symphony::Profiler()->sample('Page Built', PROFILE_LAP);
--- a/symphony/lib/toolkit/class.general.php
+++ b/symphony/lib/toolkit/class.general.php
@@ -129,7 +129,7 @@ class General
 
             </xsl:stylesheet>';
 
-            $xsltProcessor->process($_data, $xsl, array());
+            $xsltProcessor->process($_data, $xsl);
 
             if ($xsltProcessor->isErrors()) {
                 $errors = $xsltProcessor->getError(true);
--- a/symphony/lib/toolkit/class.xsltpage.php
+++ b/symphony/lib/toolkit/class.xsltpage.php
@@ -30,20 +30,6 @@ class XSLTPage extends Page
     protected $_xsl;
 
     /**
-     * An array of all the parameters to be made available during the XSLT
-     * transform
-     * @var array
-     */
-    protected $_param = array();
-
-    /**
-     * An array of the PHP functions to be made available during the XSLT
-     * transform
-     * @var array
-     */
-    protected $_registered_php_functions = array();
-
-    /**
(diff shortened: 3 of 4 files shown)


michael-e:
Wow, this is a major change. Have you tested this? Should I? (From looking at the code it's impossible for me to see what can happen.)

nitriques:
> Have you tested this?

Yes I have. I've also made lots of find in folder to make sure the methods that got removed were not used anymore.

> Should I?

Yes, you always should if you can ! ;) 

> From looking at the code it's impossible for me to see what can happen.

Does it helps to read 0614ff5's commit message ?

michael-e:
> Does it helps to read 0614ff5's commit message ?

It helps to understand the intention, but how could it ever reveal unplanned "side effects"? :-)

Will try and test on the weekend.

nitriques:
> but how could it ever reveal unplanned "side effects"? :-)

Haha. I would say, let's try and see ! But I feel pretty confident about it since it's only a redistribution of responsibilities. And it's 3.0.0, we are allowed to break thing.

michael-e:
If we have the time, just give me one or two days to install and test this. Will do my best.

nitriques:
Great. Thanks.

nitriques:
1. Yes you are right.
  - [ ] We need to open an issue and fix the problem ;)
2. Yes you are right.
  - [ ] We need to open an issue and fix the problem ;)
3. This is not true! Instead of `$context['page']->registerPHPFunction()` you would do `$context['page']->Proc->registerPHPFunction()` :) You just have to access the right object ! Does this fixes your problem ?

michael-e:
Re 3.: Ah, I see. So I will do more testing! Unfortunately I won't find the time before Sunday. :-(

nitriques:
No problems, I can wait until Monday !

nitriques:
@michael-e I do not want to put any pressure, but this needs to land right after 2.7.5 is merged because I need to rebase 3.0.x. You can trust me on this, and if you ever loose a feature, I'll make sure it gets addressed.

nitriques:
> Sorry for all the open reviews.

Not a problem.

> I noticed that with the changes in this PR, the Email Template Manager extension will cause errors when a data source gets deleted in the backend. Don't ask me why. 

I can not *not* ask why! I do not see how this can be related to data sources !!

> But will I ever be able to fix the ETM for Sym 3.0?

I am here to help if you need so. But we need to move forward.

michael-e:
> I can not *not* ask why!

Sure, but I really can't tell. I am under pressure currently, so I don't find the time to debug this.

> I am here to help if you need so.

Thanks!

> But we need to move forward.

Agreed on that!

nitriques:
Great, thanks @michael-e

PR #2805, opened 2018-02-16
Make DS params in XML optional

### LINKED ISSUE #784: Make DS params in XML optional
I would like to see a setting to turn on and off the parameter output to XML in `class.frontendpage.php` starting on line 444.  

**Use case**: when outputting a `$ds-datasource-parameter` such as IDs to be used as a filter in another data source, there can be unnecessary overhead when the number of IDs is significant. 

What do you guys think?

nitriques:
This commits add a new parameter in data sources to control if the
output parameters should be added in the xml output.

By default, data sources that do not have the `dsParamPARAMXML` property
set will include their output parameters in the XML, since this is their
current behavior. If the `dsParamPARAMXML` exists on the object, it only
needs to be different from 'yes' to be excluded from the XML.

Fixes #784

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -75,6 +75,7 @@ class contentBlueprintsDatasources extends ResourcesPage
             'required_url_param' => null,
             'negate_url_param' => null,
             'param' => null,
+            'paramxml' => null,
         );
 
         if (isset($_POST['fields'])) {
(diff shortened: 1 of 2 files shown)


michael-e:
In the corresponding discussion (#784) we said that it would be useful if data sources were able to output system IDs only. I will send a PR for this.

nitriques:
Thanks @michael-e !

PR #2806, opened 2018-02-19
Data Sources: Allow to output system IDs only

michael-e:
As discussed in #784, it may be useful to output system IDs only (i.e. without the need to add a field to DS output).


nitriques:
And you would keep the fact that empty selection does not output anything ?

We already have the id attribute on the `<entry>` node, why not just output it on empty selection ?

michael-e:
Why should we introduce magic here? Selecting nothing means that I want nothing. Selecting `system:id` means that I want IDs.

It's also a performance thingie. If you only need param output, why should Symphony build XML output as well?

nitriques:
> It's also a performance thingie. If you only need param output, why should Symphony build XML output as well?

Good point. I would have liked to not output a node under `<entry>` just for the id, but your the output parameter argument is stronger ;)

PR #2810, opened 2018-03-02
Ignore errors generated by unlink for delete methods

fhamon:
Make sure the file exist before deleting it or return true otherwise to fool Symphony. :)
Ignore errors of file_exists, is_dir and unlink.
It fixes the error thrown at us the first time we open Symphony to config the installation.

👋


froschdesign:
> Ignore errors…

…is always a bad idea. It's like closing your eyes and crossing the street. :see_no_evil:

nitriques:
@froschdesign We do not ignore them, we silence them ;) The end result will always be available for the API caller.

PR #2811, opened 2018-03-05
core patch to avoid checking all instances of fetch function to ensur…

jonmifsud:
Changes `is_int` to `is_numeric` to eliminate the addition of sorting to `entry_id` when these are numeric strings not integers. 

If unchanged this may cause some odd scenarios if the join table has no data, specifically when sorting by a datetime field.


nitriques:
Merged as 829b329a00

Thanks Jon!

PR #2813, opened 2018-03-16
Add missing Resources delegates

nitriques:
Both events and data sources were missing the PostDelete delegates.

No delegates were fired when the resources were deleted from the index
page.

The handle of the resource is added to PreDelete also.

Re https://github.com/symphonists/globalresourceloader/issues/13

### DIFF
--- a/symphony/lib/toolkit/class.resourcespage.php
+++ b/symphony/lib/toolkit/class.resourcespage.php
@@ -316,6 +316,8 @@ abstract class ResourcesPage extends AdministrationPage
     public function __actionIndex($resource_type)
     {
         $manager = ResourceManager::getManagerFromType($resource_type);
+        $resource_name = str_replace('Manager', '', $manager);
+        $delegate_path = strtolower($resource_name);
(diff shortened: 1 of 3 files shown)


symphony/lib/toolkit/class.resourcespage.php
```diff
@@ -347,7 +349,29 @@ public function __actionIndex($resource_type)
                         // Don't allow Extension resources to be deleted. RE: #2027
                         if (stripos($path, EXTENSIONS) === 0) {
                             continue;
-                        } elseif (!General::deleteFile($path)) {
+                        }
+                        
+                        /**
+                         * Prior to deleting the Resource file. Target file path is provided.
+                         *
```
fhamon: Any way we can add the `@delegate` in the comment block? I assume it's with this parameter that the online doc auto generate ?

nitriques: No since it is dynamic in this page. But we only need the `@delegate` bloc once and it's there already (in the content page)

nitriques:
> I'd prefer to see both, 

@michael-e Indeed!

PR #2815, opened 2018-03-16
Fix the privilege escalation bug

fhamon:
As of right now, no validation was made on the user_type so anyone who is logged in the the cms could edit their account (or create a new one) and modify the form to set their type for developer.

### DIFF
--- a/symphony/content/content.systemauthors.php
+++ b/symphony/content/content.systemauthors.php
@@ -348,7 +348,11 @@ class contentSystemAuthors extends AdministrationPage
             }
 
             $label->appendChild(Widget::Select('fields[user_type]', $options));
-            $div->appendChild($label);
+            if (isset($this->_errors['user_type'])) {
+                $div->appendChild(Widget::Error($label, $this->_errors['user_type']));
+            } else {
+                $div->appendChild($label);
+            }
         }
 
         $group->appendChild($div);
@@ -574,8 +578,7 @@ class contentSystemAuthors extends AdministrationPage
             }
 
             if (Symphony::Author()->isManager() && $fields['user_type'] !== 'author') {
-                $this->pageAlert(__('The user type is invalid. You can only create Authors.'), Alert::ERROR);
-                return;
+                $this->_errors['user_type'] = __('The user type is invalid. You can only create Authors.');
             }
 
             $this->_Author = new Author();
@@ -685,6 +688,11 @@ class contentSystemAuthors extends AdministrationPage
         if (@array_key_exists('save', $_POST['action']) || @array_key_exists('done', $_POST['action'])) {
             $authenticated = false;
 
+            if (!$isOwner && !$canEdit) {
+                $this->pageAlert(__('You don\'t have the rights to edit an author.'), Alert::ERROR);
+                return;
+            }
+
             if ($fields['email'] != $this->_Author->get('email')) {
                 $changing_email = true;
             }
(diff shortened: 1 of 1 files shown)


symphony/content/content.systemauthors.php
```diff
@@ -566,6 +566,18 @@ public function __actionNew()
...
+            if (!$canCreate) {
+                $this->pageAlert(__('You don\'t have the rights to create a new author.'), Alert::ERROR);
+                return;
+            }
+
+            if (Symphony::Author()->isManager() && $fields['user_type'] !== 'author') {
+                $this->pageAlert(__('The user type is invalid. You can only create Authors.'), Alert::ERROR);
+                return;
+            }
+
```
nitriques: I would rather use the `$this->_errors['user_type']` mechanism instead of page alerts.

symphony/content/content.systemauthors.php
```diff
@@ -695,7 +707,15 @@ public function __actionEdit()
...
-            } elseif ((Symphony::Author()->isDeveloper() || Symphony::Author()->isManager()) && isset($fields['user_type'])) {
+            } elseif (Symphony::Author()->isManager() && isset($fields['user_type'])) { // Manager can only change user type for author and managers
+
+                if ($fields['user_type'] !== 'author' && $fields['user_type'] !== 'manager') {
+                    $this->pageAlert(__('The user type is invalid. You can only set authors and manager.'), Alert::ERROR);
+                    return;
+                }
+
+                $this->_Author->set('user_type', $fields['user_type']);
+            } elseif (Symphony::Author()->isDeveloper() && isset($fields['user_type'])) {
```
nitriques: I would also add this check if an author wants to edit someone and throw if is does.

symphony/content/content.systemauthors.php
```diff
@@ -566,6 +570,17 @@ public function __actionNew()
         if (@array_key_exists('save', $_POST['action']) || @array_key_exists('done', $_POST['action'])) {
             $fields = $_POST['fields'];
 
+            $canCreate = Symphony::Author()->isDeveloper() || Symphony::Author()->isManager();
+
+            if (!$canCreate) {
+                $this->pageAlert(__('You don\'t have the rights to create a new author.'), Alert::ERROR);
+                return;
+            }
```
nitriques: We can get rid of the page alert since we block this already

symphony/content/content.systemauthors.php
```diff
@@ -673,6 +688,11 @@ public function __actionEdit()
         if (@array_key_exists('save', $_POST['action']) || @array_key_exists('done', $_POST['action'])) {
             $authenticated = false;
 
+            if (!$isOwner && !$canEdit) {
+                $this->pageAlert(__('You don\'t have the rights to edit an author.'), Alert::ERROR);
+                return;
+            }
```
nitriques: Same here

nitriques:
@brendo Yes, the UI was ok, so no UX problem. But security must be enforced !

symphony/content/content.systemauthors.php
fhamon: `we block this already`
Nope. This is where the validation is made.

symphony/content/content.systemauthors.php
fhamon: Same here 😁

symphony/content/content.systemauthors.php
nitriques: See https://github.com/symphonycms/symphony-2/pull/2815/files#diff-9375ed02f5d8084262cfb796bef2eb9fR218 (Line 218). The use should never reached those lines. The problem is that actions are processed before the views: This is why it is needed to valide access when changing the value.

symphony/content/content.systemauthors.php
nitriques: Same here :pick:

symphony/content/content.systemauthors.php
fhamon: As discussed we will keep the validation in the post and I will change the page alert for a `throwCustomError` That way we will have the validation on the GET request and the POST

nitriques: :+1:

PR #2816, opened 2018-03-19
Fix createHandle + limitWords to work for non-space breaking characters

jonmifsud:
General::createHandle was broken with the release of symphony 2.7.x and this is an attempt to fix the issue. @brendo & @nitriques made a nice job in having words cut out at the right place, except we forgot one use case; this was also used to trim handles.

I'm suggesting to add a character variable which is by default a space but could be replaced, for example with a `-` to be used for handles. 

In most cases this was hard to find/spot; as most handles used are at 255 characters and most use-cases are indeed shorter. For easier testing; I suggest checking with lower limits - the current behaviour completely trims the string out to leave an empty string `""` so releasing this fix should be somewhat of a priority.

### DIFF
--- a/symphony/lib/toolkit/class.general.php
+++ b/symphony/lib/toolkit/class.general.php
@@ -318,7 +318,7 @@ class General
 
         // Trim it
         if ($max_length > 0) {
-            $string = General::limitWords($string, $max_length);
+            $string = General::limitWords($string, $max_length,null,'-');
         }
 
         // Replace spaces (tab, newline etc) with the delimiter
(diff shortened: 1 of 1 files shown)


nitriques:
@jonmifsud Good find ! But instead of adding "yet another parameter", I think we could `strpos` for a space, and if `=== -1` use '-'.

I am not against using a better thing then space for a word boundary... (like any punctuation...)

symphony/lib/toolkit/class.general.php
```diff
@@ -1388,11 +1388,14 @@ public static function countWords($string)
      *  true if the ellipses should be appended to the result in circumstances
      *  where the result is shorter than the input string. false otherwise. this
      *  defaults to false.
+     * @param string $character (optional)
```
fhamon: Is there any way we can had a `@since Symphony 2.7` only for a param? @nitriques

jonmifsud:
@nitriques I'm fine with your suggestion :) - I thought of matching any whitespace but `strpos` only takes a single character. Unless someone sees any added benefit with my suggestion having less parameters makes it simpler/cleaner.

symphony/lib/toolkit/class.general.php
nitriques: Yes we do, but I would try to avoid adding a parameter.

nitriques:
@michael-e 

>  Just comment, or approve blindly

Just comment ;) Thanks !

nitriques:
@jonmifsud This is how I would do it (untested)

```php
$workBreaks = array(' ', '-', ',', '.');
$pos = max(array_map(function ($wb) {
   return strrpos($string, $wb, $maxChars - $original_length);
}, $workBreaks));
```

nitriques:
Merged as fd8c0bf9df

PR #2821, opened 2018-04-08
Prevent empty session creation

### LINKED ISSUE #2820: Empty Session creation
Affected Symphony version(s) : 2.7.0+
PHP version(s) : 5.5.9
OS(es) : Ubuntu

When creating a session, there is a [short circuit](https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/core/class.session.php#L186) to prevent saving "empty" sessions in the database by checking if `Session::read` returns null. [Since this commit](https://github.com/symphonycms/symphony-2/commit/fd4c0d3d96e96af5dd0af79490d0e8c73550a469), the `Session::read` casts it's output to a string, meaning the check will never fulfil, and empty sessions are likely created.

Fixing this will likely improve performance and reduce unnecessary session creation. This is a regression in the 2.7.x branch.

brendo:
Fixes #2820

### DIFF
--- a/symphony/lib/core/class.session.php
+++ b/symphony/lib/core/class.session.php
@@ -183,7 +183,7 @@ class Session
         // in the database. This prevents empty Sessions from being created, but
         // allows them to be nulled.
         $session_data = Session::read($id);
-        if (is_null($session_data)) {
+        if ($session_data === '') {
             $empty = true;
             $unserialized_data = Session::unserialize($data);
 
--- a/symphony/lib/core/class.session.php
+++ b/symphony/lib/core/class.session.php
@@ -183,7 +183,7 @@ class Session
         // in the database. This prevents empty Sessions from being created, but
         // allows them to be nulled.
         $session_data = Session::read($id);
-        if ($session_data === '') {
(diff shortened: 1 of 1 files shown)


symphony/lib/core/class.session.php
```diff
@@ -183,7 +183,7 @@ public static function write($id, $data)
         // in the database. This prevents empty Sessions from being created, but
         // allows them to be nulled.
         $session_data = Session::read($id);
-        if (is_null($session_data)) {
+        if ($session_data === '') {
```
nitriques: Why not do `!$session_data` ? I would cover both cases ?

brendo: Just didn't think of it :) Based on the current implementation, `Session::read` can't return anything _but_ a string.

nitriques: Yes but it would prevent future breakage ;) I really hate having `''` in my code, maybe I am crazy.

brendo: Heh, ok, updated 👍

PR #2822, opened 2018-04-13
Double encode value in displayPublishPanel()

### LINKED ISSUE #2819: Textareas do not always preserve content
Affected Symphony version(s) : since 2.3
PHP version(s) : all
OS(es) : all

Symphony's textareas do not always preserve the content.

This has to do with the usage of `General::sanitize` when preparing the content for the browser. The sanitize function has been changed a long time ago in order to prevent double encoding: https://github.com/symphonycms/symphony-2/commit/444475fb76399a1e3d725e923cb3366cd6826e21

This may have solved certain problems with double-encoding (I haven't done any research on this), but it has created an issue for textareas. For example, save the following:

```html
<div>&gt; hello</div>
```

After saving, the textarea will display:

```html
<div>> hello</div>
```

How bad. You can not save this again without fixing it.

I think that textareas should **not** use `General::sanitize` at all in `displayPublishPanel`, because the task here is not simply sanitizing but *displaying the original content in the browser*. So my suggestion is to replace `General::sanitize($value)` with `htmlspecialchars($value, ENT_COMPAT, 'UTF-8', true)` in `fieldTextarea::displayPublishPanel`.

Any thoughts on this are welcome.

nitriques:
This makes sure the user will actually see the entity.

Fixes #2819

### DIFF
--- a/symphony/lib/toolkit/class.general.php
+++ b/symphony/lib/toolkit/class.general.php
@@ -14,6 +14,7 @@ class General
      * Symphony 2.3, this function assumes UTF-8 and will not double
      * encode strings.
      *
+     * @uses htmlspecialchars()
      * @param string $source
      *  a string to operate on.
      * @return string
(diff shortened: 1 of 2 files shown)


symphony/lib/toolkit/class.general.php
```diff
@@ -26,6 +27,24 @@ public static function sanitize($source)
...
+     * @since Symphony 2.7.5
+     * @uses htmlspecialchars()
+     * @param string $source
+     *  a string to operate on.
+     * @return string
+     *  the fully encoded version of the string.
+     */
+    public static function sanitizeDouble($source)
+    {
+        $source = htmlspecialchars($value, ENT_COMPAT, 'UTF-8', true);
```
michael-e: `$value` is wrong, it must be `$source` — this is a serious bug.

fhamon: fixed in #2823

PR #2823, opened 2018-04-17
Renamed the undefined var

fhamon:
Fixes empty textareas in publish view even if the content is present in the database.

### DIFF
--- a/symphony/lib/toolkit/class.general.php
+++ b/symphony/lib/toolkit/class.general.php
@@ -40,7 +40,7 @@ class General
      */
     public static function sanitizeDouble($source)
     {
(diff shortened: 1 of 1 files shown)


symphony/lib/toolkit/class.general.php
```diff
@@ -40,7 +40,7 @@ public static function sanitize($source)
      */
     public static function sanitizeDouble($source)
     {
-        $source = htmlspecialchars($value, ENT_COMPAT, 'UTF-8', true);
```
nitriques: Why did I put `$value` in there... Even after 15 years of professional experience, it baffles me that I still do so many "rookie" mistakes. Thanks @fhamon

fhamon: ¯\_(ツ)_/¯

PR #2827, opened 2018-04-23
PDO Number 5

nitriques:
Oh boy, here it is. *The worst PR ever.* I do not even hope any complete review on this. I will merge it very fast. But I wanted to have place where it is possible to have discussion on it!

After 2 years, it finally happen: We have a new PDO implementation. We also have the following:

1. Redesigned Database API
1. 99% Compatibility with the old API (only delete without where are not retro-compatible)
1. InnoDB by default
1. Transactions
1. Unit/integration tests
1. Travis CI
1. Proper (it is a start) documentation directly in the code

There will be bugs. Please open new issues for it. But I feel like its the only way to go forward.

I was also able to keep the [original PDO implementation commit](https://github.com/symphonycms/symphony-2/commit/fee6f9020edd3457fbb9ef6ecb1f5654e668f806) from @nils-werner and @brendo in the history. Thanks guys for your work, it help me a lot.


brendo:
This is amazing. I didn't even know this was in the works! 

> There will be bugs. Please open new issues for it. But I feel like its the only way to go forward.

💯I agree. I feel this will be like Symphony 1.7 to Symphony 2 in that way. The changes are fundamental though, and they go a long way to bringing Symphony into the new world where people can use different database backends, build on CI infrastructure and just generally start to adopt modern standards.

Hell, just for maintainers on this project it means we can finally start to have automated tests for our changes instead of 🙏that things are good or doing manual QA. 

Right now Travis fails, but it doesn't look like we're attempting the CI install yet. Is that intended?

Definitely feels like a merge and fix forward though!

nitriques:
> Is there a configuration setting?

[Yes](https://github.com/symphonycms/symphony-2/pull/2827/files#diff-9406c7875b31e3f2f4699e9fba79eb06)

> Including re-creating the indexes?

Yes that would be required. That's why I did not include it in the upgrade: this is not something that can be done automatically: developer have to understand the risk associated with it.

michael-e:
Pretty cool! Well, yes, I think that an update script for the database would be great. I could maybe test it with a copy of my "big system" DB. :-)

.docs/api/DATABASE.md
```diff
@@ -0,0 +1,410 @@
...
+Note that if there was a SQL error, `PDO` is configured to throw.
+We wrap does errors in a `DatabaseStatementException` in order to provide more context.
+
+Here is the list of all factory methods, the type of object they create and their basic usage:
+
+### `select()` -> `DatabaseQuery`
+```php
+Symphony::Database()
+    ->select(['t.*'])
+    ->from('table')
```
fhamon: I think we should use metasyntactic variables in the documentation to make it easier to read and differentiate what is database jargon and users code.

wdyt @nitriques ? 😁

nitriques: I would not do it for all example, but for the most complex one, why not !

nitriques:
481e5e2 brings running the unit tests on a Windows CI !! and it works !

.docs/api/DATABASE.md
fhamon: Good for me ! 🍰

nitriques:
@froschdesign

We would indeed need to do more work to support multiple DBMS. Adapters would need to be created.

But the new API allows us to work in that direction (the previous one, not so)

As for the SQL_CACHE directive, it got migrate, but [it is marked as deprecated](https://github.com/DeuxHuitHuit/symphonycms/blob/5bb5790f74286318798f55ac812aa344556c68ea/symphony/lib/toolkit/class.database.php#L275)

PR #2829, opened 2018-04-27
Actually make auth tokens longer (40 chars)

### LINKED ISSUE #2724: Make the auth token LONGER
This is a breaking change, as it would invalidate all current auth token url.

But this change is needed.

Current implementation uses 8 chars: I propose 40 (SHA 1 = 160 bits, not truncated, which means 40 characters in HEX)

I propose to use sha256 also.

Re #2711

nitriques:
Fixes #2724

Both reset password token and auth token are 40 chars long.
Database is changed to take 255 long token (for the future)

Auth token are generated on demand. Deactivating the remote login deletes the token. Renabling it creates a new token.

### DIFF
--- a/symphony/lib/toolkit/class.authormanager.php
+++ b/symphony/lib/toolkit/class.authormanager.php
@@ -82,6 +82,43 @@ class AuthorManager
     }
 
     /**
+     * Undocumented function
+     *
+     * @param [type] $token
+     * @return void
+     */
+    public function fetchByPasswordResetToken($token)
+    {
+        if (!$token) {
+            return null;
+        }
+        return $this->select()
+            ->innerjoin('tbl_forgotpass')->alias('f')
+            ->on(['a.id' => '$f.author_id'])
+            ->where(['f.expiry' => ['>' => DateTimeObj::getGMT('c')]])
+            ->where(['f.token' => $token])
+            ->execute()
+            ->next();
+    }
+
+    /**
+     * Undocumented function
+     *
+     * @param [type] $token
+     * @return void
+     */
+    public function fetchByAuthToken($token)
+    {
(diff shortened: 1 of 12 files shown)


symphony/lib/toolkit/class.authormanager.php
```diff
@@ -81,6 +81,43 @@ public static function delete($id)
             ->success();
     }
 
+    /**
+     * Undocumented function
```
michael-e: "Undocumented function"? Looks like an oversight to me.

michael-e: Ah, you did it later!

nitriques: ;)

nitriques:
Yes @michael-e there are no hurry to merge it anyways. I thought it would be small.. but you know how it goes ;)

michael-e:
Oh yes, I know. Features are like icebergs. You only see a small part…

symphony/lib/toolkit/class.cryptography.php
```diff
@@ -96,4 +96,31 @@ public static function generateSalt($length)
...
+            throw new Exception('Can not generate less than 16 random bytes');
+        }
+        if (function_exists('random_bytes')) {
+            return bin2hex(random_bytes($length / 2));
+        } elseif (function_exists('openssl_random_pseudo_bytes')) {
+            return bin2hex(openssl_random_pseudo_bytes($length / 2));
+        }
+        return self::generateSalt($length);
+    }
 }
```
michael-e: I like the way this function works. Checking `random_bytes` exists is a good idea.

nitriques:
Thanks @michael-e !

PR #2832, opened 2018-05-01
Some love for XMLElement (bis)

nitriques:
This commit makes several changes to the XMLElement class.

First, it makes all operations returns the XMLElement instance, so that operation can be chained.

Second, it splits the XMLElement into two class by adding a new XMLDocument class.
This changes removes the need to keep a copy of properties that are only needed for the top level node in the tree. It should reduce the memory usage when dealing with large XML trees.

Third, it deprecates both convert functions. There new implementation should be faster and saner.

Finally, tests were created before the big splits, so hopefully no bugs should have been introduce (or we are missing a test!)


nitriques:
@michael-e :)

PR #2836, opened 2018-05-09
Make section, field and event start with a letter

### LINKED ISSUE #2765: Include a count of entries in associated sections doesn't work when a section starts with a number

nitriques:
This is needed because XML does not allow node names to start with a
letter.

Fixes #2765

### DIFF
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1001,19 +1001,21 @@ class Field implements ArrayAccess
         $parent_section = $this->get('parent_section');
         $label = $this->get('label');
         $element_name = $this->get('element_name');
-        $valid_name = preg_match('/^[\p{L}]([0-9\p{L}\.\-\_]+)?$/u', $element_name);
 
         if ($label === '') {
             $errors['label'] = __('This is a required field.');
         } elseif (strtolower($label) === 'id') {
             $errors['label'] = __('%s is a reserved name used by the system and is not allowed for a field handle. Try using %s instead.', array('<code>ID</code>', '<code>UID</code>'));
+        // Check label starts with a letter
+        } elseif (!preg_match('/^\p{L}/u', $label)) {
+            $errors['label'] = __('The label of the field must begin with a letter.');
         }
 
         if ($element_name === '') {
             $errors['element_name'] = __('This is a required field.');
         } elseif ($element_name === 'id') {
             $errors['element_name'] = __('%s is a reserved name used by the system and is not allowed for a field handle. Try using %s instead.', array('<code>id</code>', '<code>uid</code>'));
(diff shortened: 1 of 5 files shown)


nitriques:
> /^\p{L}/u

Are we ok with the fact that we would now require Unicode support all the time ?

> But: Since the issue is about "XML-valid" handles,

I case of fields and section, I would agree, since the label is user controlled.

But for events (same goes for data sources), they are not. But again, I do not like the idea of validating sometimes the name sometimes the handle.

Result: I would vote for the `/^\p{L}/u` check and require unicode support.

michael-e:
> Are we ok with the fact that we would now require Unicode support all the time ?

I think that we might do it in Symphony 3. But what exactly does this mean? `mbstring` module? If I look at the documentation regarding the `u` modifier we need here, I don't see any "requirement":
http://php.net/manual/en/reference.pcre.pattern.modifiers.php (look for PCRE_UTF8).

So I am a bit confused about what we actually need.

michael-e:
It looks to me that we don't need any additional requirements. `u` is a PCRE pattern modifier, and according to http://php.net/manual/en/pcre.installation.php, PCRE is a default extension of PHP, so it is always enabled.

This means that `/^\p{L}/u` shouldn't cause any issues.

nitriques:
:+1:

symphony/lib/toolkit/class.field.php
```diff
@@ -990,23 +990,20 @@ public function checkFields(array &$errors, $checkForDuplicates = true)
...
+        } elseif (!preg_match('/^\p{L}/u', $label)) {
+            $errors['label'] = __('The label of the field must begin with a letter.');
         }
 
         if ($element_name === '') {
             $errors['element_name'] = __('This is a required field.');
         } elseif ($element_name === 'id') {
             $errors['element_name'] = __('%s is a reserved name used by the system and is not allowed for a field handle. Try using %s instead.', array('<code>id</code>', '<code>uid</code>'));
-        } elseif (!$valid_name) {
+        } elseif (!preg_match('/^[a-z]/i', $element_name)) {
```
michael-e: I am not sure if this is actually needed. I think that `/^\p{L}/u` would suffice for valid XML. It may be a matter of taste or "best practice", however.

nitriques: I was not comfortable with removing it. Like I've said I am tempted to add to check everywhere the handle is user defined.

PR #2839, opened 2018-05-15
Remove unset $id (introduced by copy/paste)

### LINKED ISSUE #2838: "Undefined variable" notice caused by copy/paste typo
Affected Symphony version(s) : > 2.7
PHP version(s) : 7.x
OS(es) : all

This looks like a typo, and it throws "undefined variable" notices:

https://github.com/symphonycms/symphony-2/commit/5ddc2bd23f18b8d67435be241d85a1bc8e2a60b8#diff-c067dcaba874f8f2ad6c60d8d67af7a3R372

michael-e:
Fixes #2838

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -401,7 +401,7 @@ class contentBlueprintsDatasources extends ResourcesPage
             $li->setAttribute('data-type', 'system:id');
             $li->appendChild(new XMLElement('header', '<h4>' . __('System ID') . '</h4>'));
             $label = Widget::Label(__('Value'));
-            $input = Widget::Input('fields[filter]['.$section_id.'][system:id]', General::sanitize($id));
+            $input = Widget::Input('fields[filter]['.$section_id.'][system:id]');
             $input->setAttribute('data-search-types', 'parameters');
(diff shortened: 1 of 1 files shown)


nitriques:
Wow this is a 3,5 years old bug you just found, from 5ddc2bd23f1 !

nitriques:
Ah I just the link in the issue haha

michael-e:
Yep. :-)

As you may have heard, I started using Symphony 2.7 with `error_reporting_all` set to `yes`. I would really like to get rid of all `E_NOTICE`s in order to make the code more "bullet-proof" (think of  future versions of PHP). But I should probably open a deidcated issue for this and send a single (big) PR in the end.

michael-e:
I don't know how long this will take. There are tons of notices, and I am still in the backend… the frontend is even worse, as far as I have seen. Additionally, I need some advice how to fix things properly. (For example, when a variable that is supposed to be a string must be defined as "empty", should I make it an empty string or `null`?)

So go ahead!

nitriques:
> should I make it an empty string or null?)

It depends on the public signature (in the doc). But I would prefer to see always the same type (so empty strings) for returns, but always `empty()` (or similar) in if tests

Issue #2598, opened 2016-06-01
Ensure session id is regenerated after each request


nitriques:
According to https://www.owasp.org/index.php/PHP_Security_Cheat_Sheet#Session_Fixation it would be necessary.

I know work as already been done on session in `3.0.0`, but it can be done for `2.7.0`

@michael-e Do you see anything that could break with this change ?

michael-e:
I am not sure, would browsers handle the session ID correctly across multiple tabs/windows?

nitriques:
> would browsers handle the session ID correctly across multiple tabs/windows?

Yes: Cookie jars are shared across tabs since they are domain bounds. I've tested it and it works. But it could break other code that simply reads the cookie instead of using the core code.

michael-e:
I wouldn't care too much for hardcoded stuff that doesn't use Symphony's public API. :-)

nitriques:
Perf ;)

nitriques:
cc @hyp3rlinx

nitriques:
Pushed down to 2.8.0

PR #2843, opened 2018-05-20
WIP: Set session name to prevent collisions on subdomains. Closes #2842

### LINKED ISSUE #2842: Session fails to set on subdomains.
Affected Symphony version(s) : 2.7.6
PHP version(s) : 5.5.19 (haven't tested 7 yet)

Encountered a scenario today that was a little tricky to repeat:

- Main domain, `https://www.example.org`
- Dev domain, `http://dev.example.org`

1. Login to Symphony on the main domain ✔️ 
2. Login to Symphony on the dev domain 🙅‍♂️ 
3. Logout of Symphony on the main domain
4. Login to Symphony on the dev domain ✔️ 
5. Login to Symphony on the main domain 🙅‍♂️ 

The reason behind this weirdness appears to be the [`session.name`](http://us1.php.net/manual/en/function.session-name.php) setting. We don't set it in Symphony at all, so by default it is `PHPSESSID`.

When we attempt to login on the dev domain, it won't replace the cookie that was set previously (not too sure why?), so the login fails. You can see in the database that the session is actually created correctly, but the cookie just won't be set.

To fix it I set the `session.name` (just hardcoded it to a different string).

For a long term fix, perhaps we should set this value using the `getDomain()` method? The return value would need to be cleaned up to be [alphanumeric](http://us1.php.net/manual/en/session.configuration.php#ini.session.name) but it should solve the issue...

brendo:
This worked for me, I think others could benefit too!

This _may_ break some subdomain setups if it's expecting to use the parent cookie. I'm not sure how that would be possible given that we create a cookie that's scoped to the domain, but it could be a risk.

See #2842

### DIFF
--- a/symphony/lib/core/class.session.php
+++ b/symphony/lib/core/class.session.php
@@ -63,6 +63,9 @@ class Session
                 ini_set('session.gc_maxlifetime', $lifetime);
                 ini_set('session.gc_probability', '1');
                 ini_set('session.gc_divisor', Symphony::Configuration()->get('session_gc_divisor', 'symphony'));
+                if ($name = static::getName($domain)) {
+                    ini_set('session.name', $name);
+                }
             }
 
             session_set_save_handler(
@@ -105,7 +108,9 @@ class Session
      * with setups with special characters in their paths.
      *
      * @since Symphony 2.7.0
-     **/
+     * @param string $path
+     * @return string
+     */
     protected static function createCookieSafePath($path)
     {
         $path = array_filter(explode('/', $path));
@@ -137,6 +142,22 @@ class Session
         }
 
         return null;
+    }
+
+    /**
+     * Returns the session name, based off the domain. Used to help prevent issues
+     * with sessions on subdomains.
+     *
+     * @link http://php.net/manual/en/function.session-name.php
+     * @since Symphony 2.7.7
+     * @param  string $domain
+     * @return string
+     */
+    public static function getName($domain = null)
+    {
+        $domain = $domain ?: static::getDomain();
+
+        return preg_replace('/\W/', null, $domain);
     }
 
     /**


brendo:
Seeing some oddities, doesn't look like alphanumeric is acceptable. Alpha characters only seem to be working today 🤷‍♂️

nitriques:
@brendo why not just be stricter and restrict the cookie to the proper domain used ? Sharing this cookie across domains does not seems logical to me.

brendo:
This PR doesn't have anything to do with modifying the domain that the cookie is valid for. That's already correctly handled by [Session](https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/core/class.session.php#L80). This PR just changes the _name_ of the cookie set 👍

nitriques:
@brendo yeah, understood, but it would not be needed to have a configuration option for this IF we would be stricter...

nitriques:
@brendo Finally, I could not find a way to remove the leading dot from the cookie domain. Seems like it is php that adds it. Setting it to null, make the cookie valid for the whole domain, which is not good either (accept for localhost).

I'll rewrite the code to prevent adding a public static method.

brendo:
Moving to SSL on the subdomain fixed all this! Lesson learned.

nitriques:
@brendo @michael-e Thanks guys!!!

For 3.0.0, I would vote for "Set cookie valid for the current domain only", which is the safest thing security wise. 

@michael-e How do you currently configure your multiple domains ?

michael-e:
In order to achieve the third from the above list, I replaced [this line](https://github.com/symphonycms/symphony-2/blob/2.7.x/symphony/lib/core/class.session.php#L136) in `Session::getDomain` with:

```php
preg_match('/[^.]+\.[^.]+$/', HTTP_HOST, $matches);
$domain = $matches[0];
return $domain;
```

nitriques:
@michael-e This would open a security hole for TLD with 2 parts, like .uk.co or .qc.ca. :(

nitriques:
@michael-e Regeneration as been implemented. see #2598

michael-e:
> This would open a security hole for TLD with 2 parts, like .uk.co or .qc.ca. :(

Yep, indeed. Fortunately, I don't have any of these "pseudo TLDs"!

nitriques:
I do !!!

But I am not against solution 3. Would it be ok if it was a delegate (since it can not be a setting) ?

michael-e:
I think that **solution 2** should remain the default, simply because it works the same in all browsers. (However, Symphony does not add the leading dot to the domain — that should be fixed.)

You implemented **solution 1** using a config setting (`session_strict_domain`); but shouldn't we warn Symphony users that this does not work as intended in Internet Exploder? As cited above:

> Unfortunately, Internet Explorer appears to treat this as the host name along with any subdomains.

I suggest to add a comment to the code at least (and not make this the default). (Of course, we should verify the IE issue beforehand and test with Edge as well.)

Regarding **solution 3**: It's so counter-intuitive to have a config setting to switch between 1 and 2, but have a delegate if you need 3. If the config setting must be boolean (why?), then solution 3 can still be implemented as a hack (if ever needed). But if you say "all 3 solutions should be possible without hacking Symphony", then we shouldn't use a boolean config setting in the first place.

Anyway, the cookie domain behaviour needs documentation.

nitriques:
> IE

I seriously do not care about IE anymore. I would be a good idea to test Edge tho.

> I think that solution 2 should remain the default, 

I disagree. It is less secure and causes problems.

> However, Symphony does not add the leading dot to the domain — that should be fixed.

Agreed.

> Anyway, the cookie domain behaviour needs documentation.

Indeed.

> It's so counter-intuitive to have a config setting to switch between 1 and 2, but have a delegate if you need 3.

Indeed again. 

> If the config setting must be boolean

It does not. What do you have in mind ? Set the domain manually in the config ? Don't you need multiple domains ?

michael-e:
> I seriously do not care about IE anymore. I would be a good idea to test Edge tho.

Well, IE is still significantly bigger in market share than Edge... (see https://en.wikipedia.org/wiki/Usage_share_of_web_browsers)

> Set the domain manually in the config ?

No, please, not manually! But I admit that I don't have any idea which values could be understood (and changed) simply by reading the config. :-(

But, I can't think of a good solution to also detect "pseudo-TLDs" like `co.uk`. Do you have an idea? If we can't solve this, we shouldn't implement solution 3 at all.

michael-e:
> How did you hack it to make it work in your system ? Simply the regexp ?

See my comment above: https://github.com/symphonycms/symphony-2/pull/2843#issuecomment-396394745

> Would a 'default', 'strict-domain', 'dot-2', 'dot-3', 'mydomain' setting improve things ?

Thinking about it, I would really like to keep the config file as self-explanatory as possible. So we should keep your "boolean" solution and don't implement the third option at all. I have no problem with rebasing my custom branch on top of a current Symphony version, so I don't even need a delegate for this.

> The original goal was indeed counter intuitive, but the mindset was, the config is ok 98% of the time, lets create a delegate for the 2%.

Drop the 2% and we are fine! :-)

nitriques:
@michael-e 

Great! So nothing needs to be done than ;)

And FYI, in the mean time, I've found out how browsers can tell what is the tld and what is not: they have a hardcoded list, i.e. https://publicsuffix.org/list/public_suffix_list.dat. Check out https://publicsuffix.org/ for more infos.

michael-e:
Now that is a cool list! I didn't know that there are so many!

PR #2844, opened 2018-05-22
Output null checkbox as null values

### LINKED ISSUE #2692: Core checkbox UI filtering on 'No' fails if field added retrospectively
Steps to reproduce in v2.6.11 and workaround:

1. create a section with a text field or two
2. create several entries in this section
3. edit the section and add a checkbox field
4. add another entry (leave checkbox unticked)
5. filter entries in the UI on that checkbox and select No from the drop-down suggestion list
6. the filter only finds those entries created _after_ the checkbox was added to the section
7. remove the filter and observe all entries say No in checkbox column
8. Select all entries and use the With selected... drop-down to apply No to the checkbox field you added
9. repeat step 5 and now all other entries appear in the filter as expected

nitriques:
Fixes #2692


nitriques:
@michael-e Are you still in GDPR hell ?

michael-e:
Actually yes. GDPR means a lot of work if you need a lot of data. :-) But I like it, because privacy is supposed to work in a way people would expect. Anyway, I hope to find the time to work on open Symphony issues soon!

nitriques:
@michael-e I did it because of [this comment](https://github.com/symphonycms/symphony-2/issues/2692#issuecomment-305908326)

Did I misunderstood it ?

---

Yes, as a user, GDPR is good, and this is a good thing :+1:

michael-e:
Yes, I am afraid that you misunderstood my comment. Symphony did not output anything, so there is no sense in again making in output nothing. :-)

nitriques:
@michael-e Ok I see. So my new if does not have any purpose! (?)

michael-e:
Yes. It does not change anything. But I do not understand *why* it works even without your commit. I hoped that you would know. (I just tested it, but I don't understand it.)

nitriques:
> But I do not understand why it works even without your commit.

I bet would be on [those lines](https://github.com/symphonycms/symphony-2/blob/2.7.x/symphony/lib/toolkit/class.entrymanager.php#L546-L548). If the field data table is empty, we skip it.

In 3.0.x, that [check is still there (it got moved to another file tho)](https://github.com/symphonycms/symphony-2/blob/3.0.x/symphony/lib/toolkit/class.entryqueryresult.php#L104-L106).

nitriques:
@michael-e @fhamon I've rebased and removed 4dd0b168771362fa560c11e5c4a7539a1fb34b97 from the pr.

Issue #2605, opened 2016-07-01
No way to subvert behaviour of GenericExceptionHandler::shutdown


pointybeard:
Once a shutdown method, in this case `GenericExceptionHandler::shutdown` (https://github.com/symphonycms/symphony-2/blob/integration/symphony/lib/core/class.genericexceptionhandler.php#L288), as been set using `register_shutdown_function`, it cannot be removed. 

This is problematic as whenever there is a fatal error, this method is called and sometimes Symphony is not being used in the browser (e.g. on the command line or in a JSON api). 

Currently there is no way to remove a shutdown method in PHP once it has been set.

I propose additional logic in this shutdown method, based on the `$enabled` flag, that will allow developers to turn off Symphony's generic exception handler in favour of their own, thus no longer having to worry about this particular shutdown method from throwing out HTML on a fatal error.

A solution would be to add the following code at the top of the `shutdown` method:

``` php
        if(!self::$enabled) {
            return;
        }
```

I am still checking if there is any unwanted side effects to this, and hope to submit a PR soon.

nitriques:
@pointybeard I've already started to work on that. See https://github.com/DeuxHuitHuit/symphony-2/tree/errorhandler

The biggest problem is that Symphony treats the uninitialized state wrongly.

pointybeard:
Agreed. I think there is 2 problems. 

First, as you say, Symphony is not treating the uninitialised state correctly. I see you've made good progress on this.

And secondly, currently the shutdown method being set with `register_shutdown_function` forces HTML output, regardless of context, when there is an `E_ERROR`. This wouldn't be such a problem if there was a way to remove methods added this way, however it is not possible.

I have, however, found a potential workaround. [According to the manual](<http://php.net/manual/en/function.register-shutdown-function.php>):

> Multiple calls to register_shutdown_function() can be made, and each will be called in the same order as they were registered. **If you call exit() within one registered shutdown function, processing will stop completely and no other registered shutdown functions will be called.**

I will see about adding my own shutdown method before Symphony adds its. That method will just `exit()`, thus preventing Symphony's shutdown method from ever being called. It's a little dirty, and I think there should be a better solution in the future, but hopefully the work you're doing @nitriques will address it.

I'll post here with my findings.

nitriques:
Thanks Alistair!

PR #2845, opened 2018-05-23
ExceptionHandler overhaul

nitriques:
I wanted to have a CLI friendly ExceptionHandler for some time now, and thought that fixing #2605 would be the thing do to. I know this is (again) a big commit and do not expect full review on it, but I will gladly accept partial reviews!

It should at least make it simpler to understand and easier to maintain.

It essentially splits the ExceptionHandler responsibilities into two classes: ExceptionHandler and ExceptionRenderer.

Key commits are:

- Rename/Extract Exception/Error handlers 12d69f6
- Make it possible to disable the shutdown function c766337
- Render shutdown errors using their own exception 48d3fdc
- Add a text based exception output  48b3235


nitriques:
So can I merge and we will see ?

michael-e:
Yes, sure. But I must say that I have no plans to use Symphony 3 anytime soon, so I can't help with "Michael E.'s chaos testing" a.k.a. *I-just-wanted-to-do-x-and-something-strange-happened*.

(Actually I won't start any new projects. And the running projects — *big system* and *little sister* — are not even updated to 2.7 yet…)

nitriques:
No problem, I've got a lot of people testing it. I hope you do get the chance to at least try to break it ;)

symphony/lib/core/class.exceptionrenderer.php
```diff
@@ -0,0 +1,282 @@
...
+     *  This function works with both Exception and Throwable
+     * @since Symphony 3.0.0
+     *  This function support both html (default) and text (cli) format
+     *
+     * @param Throwable $e
+     *  The Throwable object
+     * @return string
+     *  An formatted message string
+     */
+    public static function render($e)
```
brendo: Can see a really nice opportunity here to tap into the format of a page and render 'json' or 'xml' etc.

fhamon:
This is to much for me to review. Sorry @nitriques 🤸‍♂️ . I say merge and pray for this one. hehe 😎

symphony/lib/core/class.exceptionhandler.php
```diff
@@ -0,0 +1,205 @@
...
+ */
+
+/**
+ * ExceptionHandler will handle any uncaught exceptions thrown in
+ * Symphony. Additionally, all errors in Symphony that are raised to Exceptions
+ * will be handled by this class.
+ * It is possible for Exceptions to be caught by their own `ExceptionRenderer` which can
+ * provide the `render` function so that it can be displayed to the user appropriately.
+ */
+final class ExceptionHandler
```
froschdesign: The entire class is useless, because all properties and methods are static.

I see two options here:

1. remove the class and create real functions under a namespace
2. remove the "static" keyword from all properties and methods; create a real object and inject this object in the classes `Symphony` and `SymphonyExceptionRenderer`

symphony/lib/core/class.exceptionhandler.php
```diff
@@ -0,0 +1,205 @@
...
+{
+    /**
+     * Whether the `ExceptionHandler` should handle exceptions. Defaults to true.
+     * @since Symphony 3.0.0
+     *  When disabled, exception are now rendered using the fatalerror.disabled template,
+     *  to prevent leaking debug data.
+     * @var boolean
+     */
+    public static $enabled = true;
```
froschdesign: Each value can be set here.

symphony/lib/core/class.exceptionhandler.php
```diff
@@ -0,0 +1,205 @@
...
+     *
+     * @since Symphony 3.0.0
+     *  The method is final
+     *
+     * @param Throwable $e
+     *  The Throwable object
+     * @return string
+     *  The result of the Throwable's render function
+     */
+    final public static function handler($e)
```
froschdesign: The entire class is already final.

symphony/lib/core/class.exceptionhandler.php
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * Writes an error to stdout.
+     * This is used as the last attempt to display and error to the end user,
+     * if our custom error handling code failed.
+     *
+     * @param Throwable $e
+     * @return void
+     */
+    final public static function echoRendererError($e)
```
froschdesign: Same here.

symphony/lib/core/class.exceptionrenderer.php
```diff
@@ -0,0 +1,282 @@
...
+     * @since Symphony 3.0.0
+     *  This function enforces the protected visibility.
+     *  This function has a new signature.
+     *
+     * @param Throwable $e
+     *  The Throwable object
+     * @return string
+     *  The HTML of the formatted error message.
+     */
+    protected static function renderHtml($e)
```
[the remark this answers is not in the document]
nitriques: It was my initial plan, but it felt too big of a change ;)

nitriques:
> I do wonder about final though, do we really want to prevent extensions from modifying this behaviour?

IMHO, there should be one and only one Handler. On class that register itself. The inheritance for this class was useless.

symphony/lib/core/class.exceptionhandler.php
nitriques: What do you mean ?

symphony/lib/core/class.exceptionhandler.php
nitriques: The goal is to remove all static methods, thanks for spotting that I can now do it !

symphony/lib/core/class.exceptionhandler.php
nitriques: Yup removed.

symphony/lib/core/class.exceptionhandler.php
nitriques: Yup

symphony/lib/core/class.exceptionhandler.php
nitriques: Done in d4569e1

Issue #2828, opened 2018-04-24
 Notice: Use of undefined constant rawurlencode $path = array_map(rawurlencode, $path);


wdebusschere:
Affected Symphony version(s) :
PHP version(s) :
OS(es) :

 Notice: Use of undefined constant rawurlencode $path = array_map(rawurlencode, $path);

https://github.com/symphonycms/symphony-2/blob/master/symphony/lib/core/class.session.php#L115

Please follow the guideline for security bug disclosure https://github.com/symphonycms/symphony-2/wiki/Security-Bug-Disclosure

michael-e:
I think that the line should be:

```php
$path = array_map('rawurlencode', $path);
```

instead of:

```php
$path = array_map(rawurlencode, $path);
```

nitriques:
@michael-e indeed. But still, it works ;) I'll fix it in 3.0.0

michael-e:
We generate warnings in the logs, so it would be cool to fix this in the LTS version, too.

wdebusschere:
I'm using the code below in manifest.config, thanks @nitriques 
but this gets overwritten in each symphony update

 ###### LOG ######
        'log' => array(
            'archive' => '1',
            'maxsize' => '10240000',
            'filter' => E_ALL ^ E_DEPRECATED,
        ),
        ########

nitriques:
> We generate warnings in the logs, so it would be cool to fix this in the LTS version, too.

@michael-e even with the default filter ??

> but this gets overwritten in each symphony update

With 24575 ? (this is the same value!)

nitriques:
```
$ php -r "echo E_ALL ^ E_DEPRECATED;"
24575
```

michael-e:
> @michael-e even with the default filter ??

@wdebusschere will know — he told me about it.

wdebusschere:
can't follow

nitriques:
This warning will only show up in the logs if

1. The log filter allows it (the default filter should)...
1. AND if the php settings are configured with with e_notice on (as @wdebusschere told me on the chat)

We turn off E_NOTICE, so I do not think we need to do anything on lts.

nilshoerrmann:
I'm just recently updated a few Symphony install to 2.7.6 and I got this on every install. Symphony throws an error pages: 

> Symphony Warning: Use of undefined constant rawurlencode - assumed 'rawurlencode' (this will throw an Error in a future version of PHP)

This is on PHP 7.2.1.
[...]

PR #2851, opened 2018-06-13
Make sure array_map does not create warnings

michael-e:
Re: #2828

This has already been fixed in 3.0.x.


nitriques:
I'll use 4d722a0 which fixes the issue a 2 places. But thanks @michael-e

PR #2859, opened 2018-08-09
Fix 2858

### LINKED ISSUE #2858: Anti-spam software suggests modifications to the email core
Affected Symphony version(s) : 2.7, 3.x
PHP version(s) : all
OS(es) : all

Anti-Spam software (e.g. [SpamAssassin](http://spamassassin.apache.org)) uses hundreds of criteria in order to build a spam score for email messages. Generally, emails built by the Symphony core are pretty "good", so in conjunction with a properly configured mail server your emails should pass Anti-Spam software without issues.

However, SpamAssassin is not alone. [Rspamd](https://rspamd.com), for example, is considered a more modern alternative by some people. When I integrated Rspamd into one of my mail servers, I noticed that it has two special rules that conflict with emails sent by Symphony:

* XM\_UA\_NO\_VERSION — *X-Mailer/User-Agent has no version*

	Symphony adds an "X-Mailer: Symphony Email Module" header to emails. Obviously Rspamd thinks that this is not really serious without a software version number. I wouldn't like to add the Symphony version here, because we shouldn't expose the exact Symphony version to the outer world without need. Do we need to advertise Symphony in email headers at all? I suggest to remove the "X-Mailer" header completely.

* SUBJ\_EXCESS\_QP — *Subect (sic!) is unnecessarily encoded in quoted-printable*

	Yes, the Symphony core always encodes the mail subject, even if it contains nothing but 7-Bit-ASCII. It's not difficult to fix this.

@nitriques: I can send a PR for both issues, if you agree with my above proposals. Would you consider this a bugfix (which would mean 2.7.x)?

michael-e:
Fixes #2858

### DIFF
--- a/symphony/lib/toolkit/class.emailhelper.php
+++ b/symphony/lib/toolkit/class.emailhelper.php
@@ -64,6 +64,11 @@ abstract class EmailHelper
             return $input;
         }
 
+        // Don't encode if all code points are in ASCII range 0-127
+        if (!preg_match('/[\\x80-\\xff]+/', $input)) {
+            return $input;
+        }
+
         $qpHexDigits  = '0123456789ABCDEF';
         $input_length = strlen($input);
         // Substract delimiters, character set and encoding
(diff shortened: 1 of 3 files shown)


nitriques:
Thanks @michael-e really good find ! Sorry for the long delay.

michael-e:
Yep, sure. Works fine.

PR #2862, opened 2018-09-14
Remove (forward) slashes (/) in htaccess

### LINKED ISSUE #2855: (Forward) slashes shouldn't be escaped in htaccess regex expressions
Affected Symphony version(s) : all
PHP version(s) : all
OS(es) : all

In `.htaccess` regex expressions, some (but not all) forward slashes are escaped. This is useless and might even be considered bad practice, because a slash has no special meaning in regex.

The lines are:

* https://github.com/symphonycms/symphony-2/blob/2.7.7/install/includes/htaccess.txt#L37
* https://github.com/symphonycms/symphony-2/blob/2.7.7/install/includes/htaccess.txt#L41
* https://github.com/symphonycms/symphony-2/blob/2.7.7/install/includes/htaccess.txt#L46

nitriques:
/ shouldn't be escaped in htaccess regex expressions

Fixes #2855

### DIFF
--- a/install/includes/htaccess.txt
+++ b/install/includes/htaccess.txt
@@ -35,16 +35,16 @@ Options +SymLinksIfOwnerMatch -Indexes
     RewriteRule ^(symphony/)?index.php(/.*/?) $1$2 [NC]
 
     ### ADMIN REWRITE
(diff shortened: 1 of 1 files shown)


nitriques:
@michael-e Good catch! I've rebased and amended the commit. Will merge as soon as CI turns green.

PR #2866, opened 2018-09-19
Remove forward slashes (/) and unicode escapes in env json

nitriques:
The json_encode function follows the standard at https://www.json.org/
which specifies to espace forward slashes. But doing so, makes all url
looks to be broken, because they are escaped.

In this particular case, since the json is outputed in a html page,
which makes it only consumable via javascript, it is safe to not espace
forward slashes, and allow reverse proxy to rewrite the urls, fixing a
bunch of problems!

### DIFF
--- a/symphony/lib/toolkit/class.administrationpage.php
+++ b/symphony/lib/toolkit/class.administrationpage.php
@@ -399,7 +399,7 @@ class AdministrationPage extends HTMLPage
         );
 
         $this->addElementToHead(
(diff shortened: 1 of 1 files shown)


nitriques:
@michael-e I second your proposed change! Should I open a new PR or update this one ?

michael-e:
Please add to this PR.

nitriques:
Done in 3f8ff2f. A compat check for PHP 5.3 was also needed.

PR #2867, opened 2018-09-20
Set a default log filter value when updating

nitriques:
The update process never took care to set the default log filter value
which ignores depreciation warninngs. This commit also adds a post
update documentation to tell the user about this change.

Re #2762

cc @nilshoerrmann

### DIFF
--- a/install/migrations/2.7.8.php
+++ b/install/migrations/2.7.8.php
@@ -14,6 +14,14 @@
 
         public static function upgrade()
         {
+            // Add the log filter value
+            // This was added in 2.7.1, but the update process never took care of it
+            // Re: #2762
+            $filter = Symphony::Configuration()->get('filter', 'log');
+            if ($filter === null) {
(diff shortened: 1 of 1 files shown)


install/migrations/2.7.8.php
```diff
@@ -14,6 +14,14 @@ public static function getReleaseNotes()
 
         public static function upgrade()
         {
+            // Add the log filter value
+            // This was added in 2.7.1, but the update process never took care of it
+            // Re: #2762
+            $filter = Symphony::Configuration()->get('filter', 'log');
+            if ($filter === null) {
+                Symphony::Configuration()->set('filter', E_ALL ^ E_DEPRECATED, 'log');
```
brendo: I have a feeling this will get stringified or turned into an integer on saving. That's ok, I believe it'll still have the same effect, it just won't be saved _exactly_ as it's seen here.

nitriques:
@michael-e I do not think so. Will be done only once and that's ok IMHO.

Issue #2864, opened 2018-09-14
Error reporting is set too late


michael-e:
Affected Symphony version(s) : 2.7.x
PHP version(s) : 7.2
OS(es) : all

Officially Symphony 2.7 is not supported on PHP 7.2. But, given the fact that it's an LTS release, I nevertheless suggest to keep it working if possible.

PHP 7.2 has an interesting bug. Accoording to [the docs](http://php.net/manual/en/function.idn-to-utf8.php), the `variant` parameter's default value for the `idn_to_utf8` function is `INTL_IDNA_VARIANT_2003` — but this value has been deprecated. (No, I am not kidding.) So calling the function without the parameter will raise a deprecation warning. The official bug report can be seen here: <https://bugs.php.net/bug.php?id=75609>.

**If PHP's default error reporting is set to `E_ALL`**, this PHP bug will cause the deprecation warning to be thrown on every page:

	Deprecated: idn_to_utf8(): INTL_IDNA_VARIANT_2003 is deprecated in /var/www/foo/symphony/lib/boot/defines.php on line 218

Why? Symphony sets its own (more forgiving/suppressing) error reporting too late. If you look at `index.php`, you see:

```php
// Include autoloader:
require_once DOCROOT . '/vendor/autoload.php';

// Include the boot script:
require_once DOCROOT . '/symphony/lib/boot/bundle.php';
```

Currently Symphony sets the error reporting in `bundle.php`. But the autoloader will load `defines.php` before! In this file (line 218) you will see:

```php
define_safe('HTTP_HOST', function_exists('idn_to_utf8') ? idn_to_utf8($http_host) : $http_host);
```

There are two possible (simple) fixes for this:

1. Move the line that defines Symphony's default error reporting from `bundle.php` to `index.php` (before the autoloader). This would mean that Symphony's error handling is applied to **everything** that "happens" in the autoloaded files.
2. Fix **only the line** that causes the issue, by calling the function like so: `idn_to_utf8($http_host, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46)`.

@nitriques: I am open to both solutions, and I can send a PR if you tell me which route to take.

nitriques:
@michael-e Moving the default error reporting up seems like a breaking change, so for LTS, it leaves us with only solution 2. Are we sure that `INTL_IDNA_VARIANT_UTS46` exists in php 5.3 (it looks like it) ? Also, do they produce the same result for the same input ?

[...]

PR #2868, opened 2018-09-23
Prevent warning if PHP's error reporting is E_ALL

michael-e:
Re: #2864

### DIFF
--- a/symphony/lib/boot/defines.php
+++ b/symphony/lib/boot/defines.php
@@ -211,11 +211,32 @@ define_safe('TWO_WEEKS', (60*60*24*14));
 define_safe('HTTPS', server_safe('HTTPS'));
 
 /**
- * Returns the current host, ie. google.com
+ * Returns the current host, e.g. google.com
  * @var string
  */
 $http_host = server_safe('HTTP_HOST');
-define_safe('HTTP_HOST', function_exists('idn_to_utf8') ? idn_to_utf8($http_host) : $http_host);
+if (function_exists('idn_to_utf8')) {
+    // In PHP 7.2, `idn_to_utf8` should not be called with default parameters,
+    // because the default for `variant` has been deprecated. However, the
+    // alternative variant `INTL_IDNA_VARIANT_UTS46` was not introduced before
+    // PHP 5.4, so we must be careful.
+    // https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003
+    // https://bugs.php.net/bug.php?id=75609
+    // @deprecated: This 'hack' can be removed later; when dropping PHP < 5.4,
+    // `idn_to_utf8($http_host, 0, INTL_IDNA_VARIANT_UTS46)` can be used
+    // exclusively.
+    if (defined('INTL_IDNA_VARIANT_UTS46')) {
+        $host_utf8 = idn_to_utf8($http_host, 0, INTL_IDNA_VARIANT_UTS46);
+    } else {
+        $host_utf8 = idn_to_utf8($http_host);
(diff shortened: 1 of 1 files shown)


symphony/lib/boot/defines.php
```diff
@@ -211,11 +211,32 @@
 define_safe('HTTPS', server_safe('HTTPS'));
 
 /**
- * Returns the current host, ie. google.com
+ * Returns the current host, e.g. google.com
```
nitriques: Out of curiosity, why not `i.e.` ?

michael-e: AFAIK i.e. means "that is", e.g. means "for example". The host is not always google.com, it's just an example. (We might es well use example.com here.)

nitriques: Cool, thanks for educating me.

nitriques:
cc @froschdesign @fhamon

symphony/lib/boot/defines.php
```diff
@@ -211,11 +211,32 @@
...
  * @var string
  */
 $http_host = server_safe('HTTP_HOST');
-define_safe('HTTP_HOST', function_exists('idn_to_utf8') ? idn_to_utf8($http_host) : $http_host);
+if (function_exists('idn_to_utf8')) {
```
froschdesign: My suggestion is still: `if (extension_loaded('intl')) {`

* if nobody knows a library that can fill the void
* if you have no documentation for it
* if you do not have an unit test for it

Btw. the "Intl" extension can be checked during the installation process.

michael-e: I will be happy to add to the pull request as soon as there is a sort of "common opinion" on the matter. :-)

nitriques: @froschdesign Please do not take my reply the wrong way. I really want to learn and find what is best for the project. @michael-e knows how much I like to argue: but I never do it in a mean/bad way. I want to learn and keep getting better, that is all.

> If nobody knows a library that can fill the void

I feel like user _could_ provide their own version of idn_to_utf8. And this would not be possible.

>  if you have no documentation for it

We do not. Never do we require it. It is not a formal dependancy. Symphony works without it. (That's why is it not checked in the installer, but things like xslt are). Should we require it ? I would not mind to do that for 3.0.x.

> if you do not have an unit test for it

Indeed we do not. But do we really need to test php's code ? Parts of me says yes, part of me says YAGNI.

Finally, since this is LTS, I fear about changing this. There is so much voodoo out there. That's why I would keep it as it is.

Please, continue to weigh in.

froschdesign: @nitriques 
> But do we really need to test php's code ?

Oh no, you would not test PHP's code here. You would test the behaviour of the entire code block / workflow. And this with different input and output values (valid and invalid) and also with the user-definied function respectively the existing of the "Intl" extension - which is much easier to test. 😉 

> Finally, since this is LTS...

But LTS does not mean that you need to support dead PHP versions! 😄

froschdesign:
@michael-e 
Another suggestion: there should be a separate issue report which lists all the legacy code that should be removed.

michael-e:
What do you mean by legacy code? Do you mean code that can be removed if we drop certain PHP versions?

symphony/lib/boot/defines.php
nitriques: > But LTS does not mean that you need to support dead PHP versions!

Indeed, but that's not what was agreed upon. Anyhow, 2.7.8 will be the last bug fix release! Thanks for your input @froschdesign

Issue #2870, opened 2018-09-27
Fix PHP 7 warnings


michael-e:
Affected Symphony version(s) : 2.7
PHP version(s) : 7.x
OS(es) : all

Some time ago we invented a configuration setting in Symphony in order to aid code improvements by outputting all PHP warnings.

My hope was that some experienced programmer would use this option to fix all those nasty warnings, simply because these warnings may become severe errors as PHP evolves. And I would love to see Symphony's LTS version work fine on the latest and greatest PHP version for the coming years.

Who ever dared to set `error_reporting_all` to `yes` for a real-world Symphony website (on PHP 7)?. I did! Well, yes, the result is a bit shocking at first sight. :-)

I will send a PR that fixes a bunch of warnings in the Symphony backend. It is far from complete, and I have not cared for the frontend at all, nor will I probably do. It just doesn't make much sense, because my programming knowledge and experience ist too limited. In other words: It would be much simpler for someone who really knows programming.

So this is really just a start (and nevertheless I am sure that some of my "fixes" should be discussed and/or improved). I would be glad to see someone taking this over, sending more PRs!

nitriques:
:+1: !!! I also do it from time to time in order to find bugs ! I will try to make the unit test pass with error_reporting_all for 3.0.x !

nitriques:
`2.7.8` is out with the fixes

PR #2871, opened 2018-09-27
Fix PHP 7 warnings

michael-e:
Re: #2870

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -337,13 +337,25 @@ class contentBlueprintsDatasources extends ResourcesPage
         $fieldset->appendChild($p);
         $group = new XMLElement('div');
 
-        $label = Widget::Checkbox('fields[redirect_on_required]', $fields['redirect_on_required'], __('The required parameter is missing'));
+        $label = Widget::Checkbox(
+            'fields[redirect_on_required]',
+            isset($fields['redirect_on_required']) ? $fields['redirect_on_required'] : null,
+            __('The required parameter is missing')
+        );
         $group->appendChild($label);
 
-        $label = Widget::Checkbox('fields[redirect_on_forbidden]', $fields['redirect_on_forbidden'], __('The forbidden parameter is present'));
+        $label = Widget::Checkbox(
+            'fields[redirect_on_forbidden]',
+            isset($fields['redirect_on_forbidden']) ? $fields['redirect_on_forbidden'] : null,
+            __('The forbidden parameter is present')
+        );
         $group->appendChild($label);
 
-        $label = Widget::Checkbox('fields[redirect_on_empty]', $fields['redirect_on_empty'], __('No results are found'));
+        $label = Widget::Checkbox(
+            'fields[redirect_on_empty]',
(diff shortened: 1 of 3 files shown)


symphony/content/content.blueprintsdatasources.php
```diff
@@ -336,13 +337,25 @@ public function __form()
         $fieldset->appendChild($p);
         $group = new XMLElement('div');
 
-        $label = Widget::Checkbox('fields[redirect_on_required]', $fields['redirect_on_required'], __('The required parameter is missing'));
+        $label = Widget::Checkbox(
+            'fields[redirect_on_required]',
+            isset($fields['redirect_on_required']) ? $fields['redirect_on_required'] : null,
```
froschdesign: `isset` is wrong here. See: http://php.net/manual/en/function.array-key-exists.php#example-6170

nitriques: Well, that's ok ;) It will do the same thing.

nitriques:
@michael-e Please try not to change the indentation for 2.7.x as it makes rebasing 3.0.x a lot harder. Thanks!

symphony/content/content.blueprintsdatasources.php
```diff
@@ -510,23 +523,48 @@ public function __form()
...
+            isset($fields['filter']['author']['email']) ? $fields['filter']['author']['email'] : null,
+            !isset($fields['filter']['author']['email'])
+        );
+        $this->__appendAuthorFilter(
+            $ol,
+            __('User Type'),
+            'user_type',
+            isset($fields['filter']['author']['user_type']) ? $fields['filter']['author']['user_type'] : null,
+            !isset($fields['filter']['author']['user_type'])
+        );
```
nitriques: Hum I cannot agree with this change. I prefer the null array implementation.  Why did it make sense to you to get rid of it ?

symphony/content/content.blueprintsdatasources.php
froschdesign: @nitriques 
> It will do the same thing.

Only in this one case, but in others it is wrong. (See in the classes `Administration`,  `DateTimeObj`, `FieldManager`, `Symphony`, `Section`, etc.)
Always use `array_key_exists`, which makes your life easier – that should be the goal. 😃

nitriques: @froschdesign 
Well, we almost always treat non-set and set to null as being the same thing, so `isset()` does the job! But yeah, we might be wrong at a couple of places ;)

symphony/content/content.blueprintsdatasources.php
michael-e: As explained in the commit message:

> Properly prevent E_NOTICE when using filters
>
> The former solution only worked if no filters were set at all. As soon as one filter was used, the array existed (but was still missing the other keys).

michael-e:
@nitriques: Regarding indentation, you mean those "one-liner arrays"? Sure I can revert to these (and I see that will make rebasing much easier), but it should be noted that it will be nearly impossible to understand these lines, because they become even more complex with the fixes.

I will add a commit.

PR #2872, opened 2018-09-28
Make sorting entries deterministic

### LINKED ISSUE #2856: Backend publish index table unambiguous sorting
Affected Symphony version(s) : 2.7.7 (all?)
PHP version(s) : 7.2
OS(es) : All

Don't know how much it is related to #2498 - guess the same issue but this time concerning backend publish index - but:

**Use case**
Having paginated section entries in a publish backend index, sorted by a field (column) which has the same value across all or most of the entries, e.g. a `yes` value checkbox field.

**Issue**
Not all entries are shown when paginating the index table; entries are randomized or being repeated on subsequent pages.

**Expected**
To be able to browse through all entries in section.

nitriques:
MySQL does not offer deterministic order bys. This leads to problems
where the user could see records move in the list or see a record that
was displayed in the previous page, which it not nice UX.

It is up to the programmer to create a deterministic sort. When the
fields contains mostly random data, the problem is not visible. But when
there are multiple records with the same value, the problem becomes
apparent: our order by clause is not unique.

The simplest fix I could find was to add the entry id as the last order
by clause. This clause needs to follow the same sort direction, to make
sure the last item becomes the first one when the order is flipped.

Fixes #2856

cc @jurajkapsz

### DIFF
--- a/symphony/lib/toolkit/class.field.php
+++ b/symphony/lib/toolkit/class.field.php
@@ -1588,7 +1588,7 @@ class Field
             $sort = 'ORDER BY RAND()';
         } else {
             $joins .= "LEFT OUTER JOIN `tbl_entries_data_".$this->get('id')."` AS `ed` ON (`e`.`id` = `ed`.`entry_id`) ";
-            $sort = sprintf('ORDER BY `ed`.`value` %s', $order);
+            $sort = sprintf('ORDER BY `ed`.`value` %1$s, `e`.`id` %1$s', $order);
         }
     }
 
--- a/symphony/lib/toolkit/fields/field.author.php
+++ b/symphony/lib/toolkit/fields/field.author.php
@@ -548,7 +548,7 @@ class FieldAuthor extends Field implements ExportableField
                 LEFT OUTER JOIN `tbl_entries_data_".$this->get('id')."` AS `ed` ON (`e`.`id` = `ed`.`entry_id`)
                 LEFT OUTER JOIN `tbl_authors` AS `a` ON (ed.author_id = a.id)
             ";
-            $sort = sprintf('ORDER BY `a`.`first_name` %1$s, `a`.`last_name` %1$s', $order);
+            $sort = sprintf('ORDER BY `a`.`first_name` %1$s, `a`.`last_name` %1$s, `e`.`id` %1$s', $order);
         }
     }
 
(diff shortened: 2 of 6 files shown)


twiro:
Good one 👍 

And seems to come up just at the right time as I'm currently struggling with exactly this behaviour in a new project :)

brendo:
Does this impact the query analyzer in MySQL? Interested in some before/after data from `EXPLAIN` queries to see if there's an impact to performance.

nitriques:
@brendo 

Original

`SELECT * FROM sym_entries as e WHERE section_id = 58 ORDER BY (SELECT ed.value FROM sym_entries_data_216 as ed WHERE ed.id = e.id) ASC`

0.0007 sec. (after a couple of runs)

```
EXPLAIN SELECT * FROM sym_entries as e WHERE section_id = 58 ORDER BY (SELECT ed.value FROM sym_entries_data_216 as ed WHERE ed.id = e.id) ASC

1	PRIMARY	e	ref	section_id	section_id	4	const	8	Using where; Using filesort	
2	DEPENDENT SUBQUERY	ed	eq_ref	PRIMARY	PRIMARY	4	dev_sym7.e.id	1	NULL
```

With the new clause

`SELECT * FROM sym_entries as e WHERE section_id = 58 ORDER BY (SELECT ed.value FROM sym_entries_data_216 as ed WHERE ed.id = e.id) ASC, e.id ASC`

0.0006 sec.

```
EXPLAIN SELECT * FROM sym_entries as e WHERE section_id = 58 ORDER BY (SELECT ed.value FROM sym_entries_data_216 as ed WHERE ed.id = e.id) ASC, e.id ASC

1	PRIMARY	e	ref	section_id	section_id	4	const	8	Using where; Using filesort	
2	DEPENDENT SUBQUERY	ed	eq_ref	PRIMARY	PRIMARY	4	dev_sym7.e.id	1	NULL	
```

So it looks like the execution plan does not change. Nor the performance. It is logic: the second order by clause is only used on ties.

Those tests where done on a section with 8 entries. I get the same results with a section with ~100 entries. I always sorted a checkbox field.

With more entries, the penalty will always be the `LIMIT ... OFFSET ...` clause, which makes mysql loads all previous records, but discards them.

nitriques:
@twiro Lots of extensions have the exact same problem, so make sure to check which field are you sorting with!

brendo:
Thanks for the additional context @nitriques!

nitriques:
@brendo The pleasure's mine ;)

PR #2874, opened 2018-09-30
$provides array in Extension class should be protected, not private.

petertron:
In the `Extension` class, the `$provides` array is there to be loaded with provider information using a `registerProviders` function in the extension driver that extends `Extension`. At present this is prevented by `$provides` being private.

### DIFF
--- a/symphony/lib/toolkit/class.extension.php
+++ b/symphony/lib/toolkit/class.extension.php
@@ -64,7 +64,7 @@ abstract class Extension
      * @since Symphony 2.5.0
      * @var array
      */
-    private static $provides = array();
+    protected static $provides = array();
 
     /**
      * Default constructor for an Extension, at this time it does nothing


froschdesign:
@nitriques 
…and the `static` is wrong here.

This would also be consistent with the use in the extension manager, because the manager uses an extension as an object:

https://github.com/symphonycms/symphony-2/blob/8ae5cae1be0fa50a9c54cd8a45ac4442d23324d0/symphony/lib/toolkit/class.extensionmanager.php#L240-L247

nitriques:
@petertron Thanks!

@froschdesign You are right! But since this is a signature change, I will do it only for 3.0.x

nitriques:
@@froschdesign  see ddc55af

nitriques:
@michael-e @fhamon can I get a ✅on this ?

michael-e:
This change (from *private* to *protected*) breaks some extensions that declare the array as private, for example the Remote Datasource extension. PHP throws a fatal error:

> Fatal error: Access level to Extension_Remote_Datasource::$provides must be protected (as in class Extension) or weaker in /path/to/extension.driver.php on line 0

I can send PRs for the extensions that I noticed breaking. Changing the access level in the extensions to *protected* should be backwards-compatible, shouldn't it?

michael-e:
BTW, the error happens on PHP 5.6. Nothing like `7.future`. :-)

michael-e:
Done.

cylkee:
> BTW, the error happens on PHP 5.6. Nothing like `7.future`. :-)

@michael-e I'm getting this error on PHP 7.3.19 when trying to install via the /symphony/system/extensions/ UI.

PHP Fatal error: Access level to Extension_Section_Schemas::$provides must be protected (as in class Extension) or weaker in /path-redacted/extensions/section_schemas/extension.driver.php on line 0

Issue #2841, opened 2018-05-20
Server health/check 


brendo:
I just spent a few hours debugging why login wasn't working on a server.

Turns out it was because [`session_save_path`](http://php.net/manual/en/function.session-save-path.php) wasn't writable. 

It made me think, is there something we could do to help debug/check that a server is healthy?

nitriques:
We could do it in the installer. We could also add a page alert for this.

brendo:
Downside of the installer is that it's usually run on the development environment, not the production environment.

froschdesign:
@nitriques 
**Extension – not in the core!**

The installer can use the extension.

nitriques:
@froschdesign Well, no sadly: they are the last thing to be installed.

Issue #2017, opened 2014-04-06
Inputs showing previous values for a second or two


DavidOliver:
After changing input values and checkboxes in certain pages, the success/saved page-load often (but not always) shows the old values and options. This is the case for Symphony 2.3.6 and 2.4.

If I leave it for a second or two and hit refresh, the new values and options are correctly shown.

If, instead, I immediately hit refresh again two or three times within the space of a second or two, the old values and options are still shown, with the new ones only showing up after a second or two of hitting refresh.

I'm seeing this behaviour in both Firefox and Chrome.

Pages affected:
- Preferences
- Data source edit

Is anyone else seeing this?

PR #2876, opened 2018-10-02
Suppress opcache warnings when opcache_invalidate does not work

jurajkapsz:
Avoid Symphony Warning screens when `opcache_invalidate` function does not work.

When editing filesystem based preferences like `Data Sources`, `Events` and Symphony's `Preferences`, saving changes will result in a Symphony Warning error dump screen, if Zend OPcache is enabled but `opcache_invalidate` function does not work.

opcache

Related discussion: https://github.com/Solutions-Nitriques/anti_brute_force/issues/42#issuecomment-426083630

### DIFF
--- a/symphony/content/content.blueprintsdatasources.php
+++ b/symphony/content/content.blueprintsdatasources.php
@@ -1458,7 +1458,7 @@ class contentBlueprintsDatasources extends ResourcesPage
                 // Write successful
             } else {
                 if (function_exists('opcache_invalidate')) {
-                    opcache_invalidate($file, true);
+                    @opcache_invalidate($file, true);
                 }
 
                 // Attach this datasources to pages
--- a/symphony/content/content.blueprintsevents.php
+++ b/symphony/content/content.blueprintsevents.php
@@ -605,7 +605,7 @@ class contentBlueprintsEvents extends ResourcesPage
                 // Write successful
             } else {
                 if (function_exists('opcache_invalidate')) {
-                    opcache_invalidate($file, true);
+                    @opcache_invalidate($file, true);
                 }
 
                 // Attach this event to pages
--- a/symphony/content/content.systempreferences.php
+++ b/symphony/content/content.systempreferences.php
@@ -223,7 +223,7 @@ class contentSystemPreferences extends AdministrationPage
 
                 if (Symphony::Configuration()->write()) {
                     if (function_exists('opcache_invalidate')) {
-                        opcache_invalidate(CONFIG, true);
+                        @opcache_invalidate(CONFIG, true);
                     }
 
                     redirect(SYMPHONY_URL . '/system/preferences/success/');


jurajkapsz:
Hope I got the description kind of right :cold_sweat:

nitriques:
👍 Thanks @jurajkapsz 

cc @froschdesign

froschdesign:
@nitriques 
Hiding an error will never solve a problem!

What I'm missing here is the reason why the warning is triggered. Can we clarify this? This includes the configuration of [`opcache.restrict_api`](http://php.net/manual/en/opcache.configuration.php#ini.opcache.restrict-api).

brendo:
My understanding is that if the `restrict_api` configuration value is anything _other_ than `''`, certain opcache functions may not work. 

```
# The value of opcache.restrict_api is the path prefix of scripts allowed to
# call opcache API functions: http://php.net/manual/en/ref.opcache.php
# Set the value as needed or comment out the line to allow all scripts to use
# the opcache API functions.
```

[Zend reference](http://files.zend.com/help/Zend-Server/content/zendserverapi/zend_opcache-configuration_directives.htm#dir-opcache.restrict_api)

If we wanted to avoid suppressing the error, we'd have to check that Symphony is not included in that value (probably with a helper function?)

https://github.com/Yoast/wordpress-seo/pull/9883 decided muting was the fastest fix without overhead. 

Interestingly https://github.com/craftcms/cms/issues/2834 suggested that servers reset the value back to ``.

Heh 🤷‍♂️

froschdesign:
@jurajkapsz 
Can you check your configuration? Thanks!

@brendo 
I think the warning shows an important detail here: the current configuration of the cache does not work with the CMS. This should not be suppressed or hidden!

> we'd have to check that Symphony is not included in that value

Maybe during the installation process or a _system health_ extension.

jurajkapsz:
> Hiding an error will never solve a problem!

No, it does not. Actually I have to be aware of this, because after I save those [mentioned](https://github.com/symphonycms/symphony-2/pull/2876#issue-219738740) preferences, the backend page still returns the old values, so I have to refresh the screen if I want to continue to edit those preferences, because I would resave it again with old values. But at least I get to the backend screen "safe and sound" with the `@` warning suppression.
Refreshing or revisiting the backend page shows the latest values.

The way we handle it now, originates from [this issue](https://github.com/symphonycms/symphony-2/issues/2017).

> Can you check your configuration? Thanks!

My `opcache.restrict_api` setting at the hosting provider has a lovely value: `/data/web/nonexist`, which is something like `/dev/null` I guess. So I think that means no playing around with OPcache API. I think I cannot change that setting. But I have the option to turn off the opcache. Do I want to do that?

brendo:
> My `opcache.restrict_api` setting at the hosting provider has a lovely value: `/data/web/nonexist`, which is something like `/dev/null` I guess.

That's odd. What they are essentially saying is that no-one can use these functions:

> Allows calling OPcache API functions only from PHP scripts which path is started from specified string. 

> Refreshing or revisiting the backend page shows the latest values.

I believe that's why the change was made to prevent requiring this.

jurajkapsz:
> That's odd. What they are essentially saying is that no-one can use these functions:
>>Allows calling OPcache API functions only from PHP scripts which path is started from specified string.

Yes. I will actually write them about this, because they tend to care. Let's see.

> I believe that's why the change was made to prevent requiring this.

Yes.

froschdesign:
@jurajkapsz 
> No, it does not.

I'm sorry, but you describe a symptom of the problem, but not the cause. Yes, it is frustrating, but hiding the warning message does not solve the cause of the problem.

> Yes. I will actually write them about this, because they tend to care. Let's see.

Very good! 👍 

@nitriques @michael-e 
This brings us back to the discussion of why it's bad that Symphony sets its own error reporting. If @jurajkapsz can not change the configuration for the OPcache then he _could_ – in the last resort - suppress the warning message on his system.

There is even a name for it: a shot in the foot! 😉

michael-e:
@froschdesign You bring up very interesting points. I would be glad if you could provide a better solution to the issue.

A remark about the opcache: Most webhosting providers configure the opcache with a revalidate frequency of 2 seconds, which causes problems (if PHP files get written by a CMS). In most cases (i.e. unless you have a real high-traffic website) setting the revalidate frequency to 0 is not an issue and solves these problems. (Remember that this only means that the files' modification time will always be checked, it does **not** render the cache useless!) You can easily do this if you have your own server, and even on webhosting it is sometimes possible. For example, if you are allowed to have a `.phpcfg` file in your webroot, you can ad `opcache.revalidate_freq=0`.

froschdesign:
@michael-e 
> I would be glad if you could provide a better solution to the issue.

1. Remove the setting of the error reporting within Symphony. (This means fixing all problems in Symphony itself. :wink:)
2. Remove and never use the "STFU operator"!

This allows the user, like @jurajkapsz, to set error reporting if the hosting provider does not want to change the configuration.
(It will also save your life if you want to debug the code.)

To avoid the current problem, it may be helpful to check the configuration during the installation process or create an extension to check the general system health, which includes the configuration of the OPcache.

---

Before anything is changed, we should wait for the answer from @jurajkapsz hosting provider. If they change the configuration for `opcache.restrict_api` then close this pull request.  If not then move the `error_reporting` from `symphony/lib/boot/bundle.php` to `manifest/config.php` with a comment like "change on your own risk" – later, when all problems are resolved in Symphony, remove this line.

jurajkapsz:
Really thanks everyone guys for all your valuable insights on this.

So, got the answer, provider told me, that they want to make sure on their shared webhosting service, which I also use, that certain opcache functions are not available, because it would be possible for me to influence other clients data. Namely:

- opcache_compile_file ()
- opcache_get_configuration ()
- opcache_get_status ()
- opcache_invalidate ()
- opcache_reset ()

Speaking of `opcache.revalidate_freq`, I also have preset a value of `2`.

So, is this just me?

Just two side notes:

> No, it does not.

@froschdesign I wanted to confirm your initial statement `Hiding an error will never solve a problem`, probably did it incorrectly english-wise.

**Secondly**, just for the record, I was thinking if we could improve the `if (function_exists('opcache_invalidate'))` condition - if that would be a thing - by checking if our PHP script is in the path of a possible `opcache.restrict_api` setting, maybe by comparing `__FILE__` with `ini_get('opcache.restrict_api')`. Because it does not make sense to try to run the `opcache_invalidate` function then, cause the result would be the same.

froschdesign:
@jurajkapsz 
Thanks for the feedback!

> I wanted to confirm your initial statement…

👍 

> …if we could improve the `if (function_exists('opcache_invalidate'))` condition…

You can. Keep in mind to test for an empty string, `false` and that the path must start with a specified string. The optimal solution includes unit tests and tests with different PHP versions. (Travis is your friend here.)

**But** this can also hide a misconfiguration of OPcache. All users who can change the PHP configuration and the `opcache.restrict_api` setting contains an incorrect path, will never know that a problem exists.

froschdesign:
@brendo 
You have already made the same proposal: #2841 "Server health/check" 😃

nitriques:
Looks like I've missed the party, as always :( 

From a purist pov, I understand why one would like to have control over their error_reporting value.

But yeah, the only way that would be feasible would be to run ok with the E_ALL option, which is not the case. It also makes us responsible to help us to debug their nasty shared hosting: that's why we currently resort to setting error_reporting.

That being said, an extension could provide the feature. So I think I am gonna wack it from 3.0.x.

Anybody opposed to this removal ? (I would keep the error_reporting_all quick option tho)

nitriques:
As for 2.7.x, since we already do it, I would mute the shit out of this error.

nitriques:
Merging for 2.7.x, will be reverted in 3.0.x

Issue #2517, opened 2015-10-30
evaluate system date logic


nitriques:
This discussion started [there](https://github.com/symphonists/system_date_fields/issues/13) but is related to the core. Basically:

- saving the same timestamp in two different representations (GMT and localized) is useless and increases the probability of error
- reading a localized date and assuming a time offset is error-prone as well (see my example above, changing the time zone)

Proposed change:

- Store only the GMT value.
- Save the timezone offset in a separate field, to preserve the information if anybody needs this.
- Removed the localized date.

cc @michael-e

nitriques:
Breaking change, so bump.

PR #2879, opened 2018-10-19
Force sorting on a date columns we select

### LINKED ISSUE #2878: Mysql 5.7.23 Compatibility: ORDER BY clause is not in SELECT list
**Mysql 5.7.23**
Affected Symphony version(s) : 2.7.7
PHP version(s) : 7.1

Normal datasource, with order by system modification date cannot be executed because system modification date is not in select statement.

```
MySQL Error (3065): Expression #1 of ORDER BY clause is not in SELECT list, references column 'zzz.e.modification_date_gmt' which is not in SELECT list; this is incompatible with DISTINCT in query: SELECT SQL_CACHE DISTINCT `e`.`id`, `e`.section_id,
                `e`.`author_id`, `e`.`modification_author_id`,
                `e`.`creation_date` AS `creation_date`,
                `e`.`modification_date` AS `modification_date`
                
            FROM `sym_entries` AS `e`
            
                LEFT JOIN
                    `sym_entries_data_230` AS t230_1
                    ON (e.id = t230_1.entry_id)
            
            WHERE 1
            
            AND `e`.`section_id` = 2
            
                    AND (t230_1.value IN ('Yes'))
                
            ORDER BY `e`.`modification_date_gmt` DESC
            LIMIT 0, 24 on 964 of file /Users/xxx/Sites/www.yyy.com/symphony/lib/toolkit/class.mysql.php
```

nitriques:
This prevents error in default setups of MySQL 5.7.x

In the core, we select the non GMT version and show the non GMT version:
a user could reasonably expect the sort to be done on the shown value.

Fixes #2878

RE #2517

### DIFF
--- a/symphony/lib/toolkit/class.entrymanager.php
+++ b/symphony/lib/toolkit/class.entrymanager.php
@@ -414,11 +414,11 @@ class EntryManager
 
         // Handle Creation Date or the old Date sorting
         } elseif (self::$_fetchSortField === 'system:creation-date' || self::$_fetchSortField === 'date') {
-            $sort = sprintf('ORDER BY `e`.`creation_date_gmt` %s', self::$_fetchSortDirection);
+            $sort = sprintf('ORDER BY `e`.`creation_date` %s', self::$_fetchSortDirection);
 
         // Handle Modification Date sorting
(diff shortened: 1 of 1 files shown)


nitriques:
cc @wdebusschere

Issue #2884, opened 2019-02-18
PHP compatibility of LTS version


michael-e:
Affected Symphony version(s) : LTS

The current Symphony version is an "LTS" (Long Term Support) version. However, PHP compatibility is limited to PHP 7.1. Active support for PHP 7.1 has already ended, and security support will end on December 1st, nine months from now (see <http://php.net/supported-versions.php>). How bad.

My suggestion is to release a 2.8.0 LTS version and mark it compatible with PHP 7.3.

I already run one of my smaller web apps on PHP 7.3, and the changes required were not too difficult. I will also make my — legendary — "big system" run on PHP 7.3, and of course I am willing to share any findings and fixes.

Until now I discovered only two significant issues with PHP 7.3. (BTW, here is a nice overwview of the changes: <https://ayesh.me/Upgrade-PHP-7.3>.)

1. PHP's migration from PCRE to PCRE2 library requires to change some Regex expressions in the Symphony core. It's no big deal (and even backwards compatible).

2. The introduction of a `JsonException` class in PHP requires to rename Symphony's `JSONException` class in order to prevent a name clash. In pure theory renaming the class is a breaking change, but I really wouldn't care in this case — who would ever use this exception except the JSON class itself?

I can send corresponding PRs, and I suggest to open a 2.8.x branch for this.

nitriques:
While LTS ends soon, I do not oppose it.  Please submit both PR on `2.7.x` for now.

nitriques:
@michael-e What about doing `if class_exists('JSONException')` before we declare the custom implementation ? Both should work (albeit they won't have the same error messages).

michael-e:
I wouldn't oppose this solution either, but I prefer to rename the Symphony class. e.g. to `SymJSONException`. This, at least theoretically, allows custom error handling in Symphony.

michael-e:
OK, done. Please take a look and comment.

All my systems are on PHP 7.3 now, so chances are that I will notice any hickups if they still occur. :-)

PR #2887, opened 2019-02-24
Fixes for PHP 7.3

michael-e:
Re: #2884


symphony/lib/toolkit/class.json.php
```diff
@@ -5,13 +5,13 @@
...
 /**
- * The `JSONException` class extends the base `Exception` class. It's only
+ * The `SymJSONException` class extends the base `Exception` class. It's only
  * difference is that it will translate the `$code` to a human readable
  * error.
  *
  * @since Symphony 2.3
  */
-class JSONException extends Exception
+class SymJSONException extends Exception
```
brendo: I'm curious, would a better fix be to only create this exception if it didn't exist? What sort of error messages does the new native `JsonException` give? Are they are descriptive as ours?

michael-e: That was [sugested](https://github.com/symphonycms/symphony-2/issues/2884#issuecomment-466185413) by @nitriques as well... Honestly I haven't tested anything here — I simply needed to make it run on PHP 7.3. Unfortunatley, the PHP docs don't telll me much about how descriptive the exception will be. But if it is desired, I can change the code acoordingly.

nitriques: The code for the native JSONException is almost identical to our implementation: https://github.com/php/php-src/blob/623911f993f39ebbe75abe2771fc89faf6b15b9b/ext/json/json.c#L215

I would not mind dropping our version. We may loose the translation, but will always support all error codes. I do not mind loosing the translation (it is not meant to be user facing anyways).

I'll fix it with a class exists check.

symphony/lib/toolkit/class.general.php
```diff
@@ -344,7 +344,7 @@ public static function createHandle($string, $max_length = 255, $delim = '-', $u
         $string = preg_replace('/[\s]+/', $delim, $string);
 
         // Find all legal characters
-        preg_match_all('/[^<>?@:!-\/\[-`;‘’…]+/u', $string, $matches);
+        preg_match_all('/[^<>?@:!\-\/\[-`;‘’…]+/u', $string, $matches);
```
nitriques: This is backward compatible right ?

symphony/lib/toolkit/class.json.php
michael-e: Please go ahead! Thanks!

vendor/composer/ClassLoader.php
```diff
@@ -279,7 +279,7 @@ public function isClassMapAuthoritative()
      */
     public function setApcuPrefix($apcuPrefix)
     {
-        $this->apcuPrefix = function_exists('apcu_fetch') && ini_get('apc.enabled') ? $apcuPrefix : null;
+        $this->apcuPrefix = function_exists('apcu_fetch') && filter_var(ini_get('apc.enabled'), FILTER_VALIDATE_BOOLEAN) ? $apcuPrefix : null;
```
nitriques: Is this auto-generated ? Or custom fixes ?

symphony/lib/toolkit/class.general.php
michael-e: Yes, definitely.

vendor/composer/ClassLoader.php
michael-e: Yes, that was generated by Composer.

michael-e: It was:

> Composer version 1.8.0 2018-12-03 10:31:16

symphony/lib/toolkit/class.json.php
michael-e: Oh, I misunderstood. Should I add a commit to my PR?

michael-e:
I think that we don't need to run `composer dump-autoload` anymore, because `autoload_classmap.php` wouldn't change. (FYI, if I try, I get the same programmatic differences in `ClassLoader.php` as before.)

symphony/lib/toolkit/class.json.php
michael-e: Done.

michael-e:
@nitriques: I added a commit with the requirements change. This might cause some headache for extension developers. Unfortunately, we never officially added the PHP compatibility attributes to the extension meta XML schema. 

However, my big system (on PHP 7.3) uses a lot of extensions, and I have only found a few issues (and sent PRs). And the more important point is: Symphony LTS should officially run on the current PHP version. I'd rather risk a bug than make Symphony look "discontinued".

nitriques:
> Unfortunately, we never officially added the PHP compatibility attributes to the extension meta XML schema.

We did it ;) it is not released yet...

> Symphony LTS should officially run on the current PHP version. 

Yes, but we can't drop anything!

froschdesign:
@nitriques 
> Yes, but we can't drop anything!

Sure you can! Or does someone pay you to support the dead PHP versions?

And it is a wrong signal to the users of the CMS, because if you support the old PHP versions the user believes he does not have to do anything. But it is a security risk!

As a (open source) project without charge, it should not be responsible for problems with unsupported PHP versions.

froschdesign:
Look also at semantic versioning: "[What should I do if I update my own dependencies without changing the public API?](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api)"

PR #2888, opened 2019-02-27
Add hash (#) to transliteration

michael-e:
Surprisingly the hash character was still missing. I noticed this when a user started an article's title with a hashtag — the handle-based URL was not working as expected, of course.


michael-e:
Will you pull this? (There is more in the works, so I'd like to rebase/start from here.)

michael-e:
You're welcome!

PR #2891, opened 2019-03-03
Fix transliterations for parentheses

### LINKED ISSUE #2890: Transliterations are broken for parentheses
Affected Symphony version(s) : 2.7+
PHP version(s) : all
OS(es) : all

It is almost unbelievable that we have not noticed this very old bug before. It was introduced by @brendo in 2011: https://github.com/symphonycms/symphony-2/commit/6983d12cf2d5f28935b1b0fd6bdeb00e234cbb9e

Transliterations don't work for parentheses — `(` and `)`.

If you, for example, create a section called `Hello (World)`, the handle will be `hello-(world)`.

I will send a PR.

michael-e:
Fixes #2890


michael-e:
Thanks!

PR #2899, opened 2019-04-17
Some minor changes to make codebase a little more flexible, especially when extending base classes.

pointybeard:
- Updated `Widget::Anchor(`), `Widget::Input()`, and `Widget::Textarea()` allow for 'string', 'int', 'double', 'float' in values.
- Made `General::ensureType()` more flexible by allowing the 'type' property in param arg to be an array of allowable types.
- Made `_fieldPool` in class.datasource.section.php protected instead of private to allow child classes access
- Changed all private methods and member variables to protected in `XMLElement` class to make extended easier
- Changed class property `FrontendPage::_page` to protected instead of private. Using `Datasource::execute()` instead of `grab()` in FrontendPage to avoid filling up logs with deprecated method messages.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4,10 +4,10 @@
 
 [![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/symphonycms/symphony-2/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/symphonycms/symphony-2/?branch=master)
 
-- Version: 2.7.9
-- Date: 26th March 2019
-- [Release notes](https://www.getsymphony.com/download/releases/version/2.7.9/)
-- [Github repository](https://github.com/symphonycms/symphony-2/tree/2.7.9)
+- Version: 2.7.10
+- Date: 8th April 2019
+- [Release notes](https://www.getsymphony.com/download/releases/version/2.7.10/)
+- [Github repository](https://github.com/symphonycms/symphony-2/tree/2.7.10)
 - [MIT Licence](https://github.com/symphonycms/symphony-2/blob/master/LICENCE)
 
 ## Contents
@@ -26,7 +26,7 @@
 
 ## Overview
 
-Symphony is a `PHP` & `MySQL` based CMS that utilises `XML` and `XSLT` as its core technologies. This repository represents version `2.7.9` and is considered stable.
+Symphony is a `PHP` & `MySQL` based CMS that utilises `XML` and `XSLT` as its core technologies. This repository represents version `2.7.10` and is considered stable.
 
 Useful places:
 
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "symphonycms/symphony-2",
-    "version": "2.7.9",
+    "version": "2.7.10",
     "description": "Symphony is a PHP & MySQL based CMS that utilises XML and XSLT as its core technologies.",
     "homepage": "https://www.getsymphony.com",
     "license": "MIT",
--- a/install/index.php
+++ b/install/index.php
@@ -27,7 +27,7 @@
     define('DOCROOT', $clean_path);
 
     // Required boot components
-    define('VERSION', '2.7.9');
+    define('VERSION', '2.7.10');
     define('INSTALL', DOCROOT . '/install');
 
     // Include autoloader:
--- a/install/migrations/2.7.10.php
+++ b/install/migrations/2.7.10.php
@@ -1,0 +1,14 @@
+<?php
+
+    class migration_2710 extends Migration
+    {
+        public static function getVersion()
+        {
+            return '2.7.10';
+        }
+
+        public static function getReleaseNotes()
+        {
+            return 'https://www.getsymphony.com/download/releases/version/2.7.10/';
+        }
+    }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "symphonycms",
-  "version": "2.7.9",
+  "version": "2.7.10",
   "lockfileVersion": 1,
   "dependencies": {
     "abbrev": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "symphonycms",
-  "version": "2.7.9",
+  "version": "2.7.10",
   "homepage": "https://www.getsymphony.com",
   "repository": {
     "type": "git",
--- a/symphony/assets/css/devkit.min.css
+++ b/symphony/assets/css/devkit.min.css
@@ -1,6 +1,6 @@
 /*!
- * symphonycms v2.7.9
- * commit 1bf7f0d - 2019-03-26
+ * symphonycms v2.7.10
+ * commit 06ee00e - 2019-04-08
  * https://www.getsymphony.com
  * Copyright (c) 2019
  * License MIT
--- a/symphony/assets/css/installer.min.css
+++ b/symphony/assets/css/installer.min.css
@@ -1,6 +1,6 @@
 /*!
- * symphonycms v2.7.9
- * commit 1bf7f0d - 2019-03-26
+ * symphonycms v2.7.10
+ * commit 06ee00e - 2019-04-08
  * https://www.getsymphony.com
  * Copyright (c) 2019
  * License MIT
--- a/symphony/assets/css/symphony.min.css
+++ b/symphony/assets/css/symphony.min.css
@@ -1,6 +1,6 @@
 /*!
- * symphonycms v2.7.9
- * commit 1bf7f0d - 2019-03-26
+ * symphonycms v2.7.10
+ * commit 06ee00e - 2019-04-08
  * https://www.getsymphony.com
  * Copyright (c) 2019
  * License MIT
--- a/symphony/assets/js/symphony.min.js
+++ b/symphony/assets/js/symphony.min.js
@@ -1,6 +1,6 @@
 /*!
- * symphonycms v2.7.9
- * commit 1bf7f0d - 2019-03-26
+ * symphonycms v2.7.10
+ * commit 06ee00e - 2019-04-08
  * https://www.getsymphony.com
  * Copyright (c) 2019
  * License MIT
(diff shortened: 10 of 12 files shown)


froschdesign:
@pointybeard 
> …to make extended easier…

…and maintaining the classes becomes more difficult, because changing the internals of the class without affecting userland code is problematic.

You do not need an inheritance everywhere. Use the getter and setter methods or patterns like composite, decorator, etc.

"mark it private unless you have a good reason not to" – and at the moment we have no examples for this.

pointybeard:
Sure, that's fine. Symphony core code is missing a lot of helpful features so I often extend classes for projects to improve its usefulness. The things I have flagged were show-stoppers in terms of extensibility and helped me to create some very useful libraries. I was just throwing them back at the main repo since I consistently hit these few changes as issues.

> You do not need an inheritance everywhere. Use the getter and setter methods or patterns like composite, decorator, etc.

Uh huh. Sure. I'm not suggesting inheritance everywhere, but, I take your point. Symphony's code base is a horrendous mess of wonky inheritance and an insanely huge vocabulary of tightly coupled, yet poorly typed helper functions. I know, because I wrote it. ;)

> Changed all private methods and member variables to protected in XMLElement class to make extended easier

Perhaps I will submit my changes to XMLElement instead.

> Made _fieldPool in class.datasource.section.php protected instead of private to allow child classes access

Currently, there are no setters or getters for this. I will look at implementing some to allow manipulation of the `$_fieldPool` then.

> Changed class property FrontendPage::_page to protected instead of private.

I'll split this commit up and look at FrontendPage to see if I can work in a setter for `$_page`

I'll resubmit a PR with the commits that have nothing to do with inheritance and one for those that do.

froschdesign:
@pointybeard 
> Symphony's code base is a horrendous mess of wonky inheritance and an insanely huge vocabulary of tightly coupled, yet poorly typed helper functions.

I do not want to contradict that. 😉 

> Currently, there are no setters or getters for this. I will look at implementing some to allow manipulation of the `$_fieldPool` then.

I didn't say getter and setter are the solution here, but it needs good reason to open the API. That decides the individual case. The class `SectionDatasource` is declared as abstract and this can be a reason to change the visibility to protected.

> Perhaps I will submit my changes to XMLElement instead.

I don't know if the master branch is the correct target. Maybe [version 3](https://github.com/symphonycms/symphony-2/tree/3.0.x)?

symphony/lib/toolkit/class.widget.php
```diff
@@ -79,7 +79,7 @@ public static function Input($name, $value = null, $type = 'text', array $attrib
     {
         General::ensureType(array(
             'name' => array('var' => $name, 'type' => 'string'),
-            'value' => array('var' => $value, 'type' => 'string', 'optional' => true),
+            'value' => array('var' => $value, 'type' => ['string', 'int', 'double', 'float'], 'optional' => true),
```
brendo: `double` is an alias of `float`, was there a benefit to including both instead of just `float`?

symphony/lib/toolkit/class.general.php
```diff
@@ -1666,7 +1666,9 @@ public static function hash($input, $algorithm = 'sha1')
      *  $key = the name of the variable
      *  $value = Array[
      *      'var' => the variable to check
-     *      'type' => enforced type. Must match the XXX part from an `is_XXX()` function
+     *      'type' => enforced type. Must match the XXX part from an `is_XXX()`
+     *                function. Can be and array of types. Validation passes if
```
brendo: Typo `and` => `an`

symphony/lib/toolkit/class.frontendpage.php
```diff
@@ -892,7 +892,7 @@ public function processDatasources($datasources, XMLElement &$wrapper, array $pa
             // if the XML is still null, an extension has not run the data source, so run normally
             // This is deprecated and will be replaced by execute in Symphony 3.0.0
             if (is_null($xml)) {
-                $xml = $ds->grab($this->_env['pool']);
+                $xml = $ds->execute($this->_env['pool']);
```
brendo: I can't believe we still have calls to `grab`! 😆

froschdesign:
@brendo 
> Over the years I've started to feel that `private` should be for extreme cases.

The other way around: it‘s an extreme position to break encapsulation of a class without any reason. Instead, you should think about it to mark classes as final.

> `protected` goes most of the way and helps with extending classes in ways you couldn't of possibly imagined.

Why do you need flexibility in first place?

Inheritance creates chains of subclasses but does not solve problems. Use interfaces and keep the public API of an object small, this allows own implementations without breaking the encapsulation of a class and both are maintainable, the implementation of the CMS and your own.

brendo:
> Why do you need flexibility in first place?

Symphony allows extensions to manipulate almost all concepts it provides so I see it as a hat tip to extension authors. Locking things down pushes authors into awkward positions:

- Duplicating code from a "core" class. This leads to maintenance nightmares and increases the boilerplate for authors.
- Raise an issue/PR on the project to provide appropriate accessors and getters.
- Internalise extensions and modifications and don't share them back to the community.
- Don't do any changes at all and walk to another product.

It's not without tradeoffs, it does increase the maintenance cost for the core team as they now need to be mindful of changing "internals".

> Inheritance creates chains of subclasses but does not solve problems. Use interfaces and keep the public API of an object small, this allows own implementations without breaking the encapsulation of a class and both are maintainable, the implementation of the CMS and your own.

I agree with you, but let's be realistic. Symphony was built in PHP4 days (by @pointybeard and friends no less) and for the most part, we have dragged into into a PHP7 world without really revisiting the core architecture and structure. We're not working with a system where things are perfectly abstracted in the first place so I think we'd do best to be mindful of this and apply two streams of thought to it:

1. How can we modernise the codebase and apply things like composition, or create interfaces, without causing heartache for extension developers (and therefore the wider community who uses Symphony as a tool, but it not necessary developer savvy).
1. How can we allow the community to continue to make a living using Symphony.

I think @pointybeard and I both agree with you in principle, but in the context of this PR it doesn't feel productive.

froschdesign:
@brendo 
> Symphony was built in PHP4 days and for the most part, we have dragged into into a PHP7 world without really revisiting the core architecture and structure. We're not working with a system where things are perfectly abstracted in the first place…

You are right with this, but this pull requests pushes the same _old_ and _not perfectly_ things. It goes in the wrong direction, hence my objection.

pointybeard:
Loving the conversation! @brendo is correct in that I do agree with you @froschdesign. 

> You are right with this, but this pull requests pushes the same _old_ and _not perfectly_ things. It goes in the wrong direction, hence my objection.

To be fair, I wrote these changes in the context of the current Symphony code base. Two points @brendo brings up are specifically applicable in my case:

> - Duplicating code from a "core" class. This leads to maintenance nightmares and increases the boilerplate for authors.
> - Internalise extensions and modifications and don't share them back to the community.

I have a lot of internalised extensions and modifications that I haven't been able to share back with the community. This PR was an attempt to start to bridge the gap so I could start doing that. I have my own build of Symphony with a handful of minor, but really helpful, modifications. This was just a couple of those changes that, if introduced to the code base, help with the "Duplicating code" problem. I don't really like running my own fork with custom code since it's tricky to keep it up to date.

You might have noticed some of the extensions I have been contributing, e.g. [Symphony Class Mapper](https://github.com/pointybeard/symphony-classmapper) and [Symphony PDO](https://github.com/pointybeard/symphony-pdo), and more recently [Symphony Section Builder](https://github.com/pointybeard/symphony-section-builder), which sidestep a lot of the core behaviour/systems and have given me a much cleaner platform for continuing to develop on Symphony.

There are two ways to go here I think: 
1. Make the code base more "flexible", as this PR tries to do. It requires minimal changes and offers opportunities for external developers to use the core codebase without resorting to custom changes.
2. Ask that, instead of these smaller PRs that modify the API, the actual improvement that they seek to enable is offered as a PR. e.g. The changes to XMLElement allowed me to write a larger set of changes that vastly improve XMLElement by extending it into a subclass. I could offer those changes back to the core.

(1) is quicker and easier. e.g. asking `private` changes to `protected`
(2) keeps the API the same while improving the core code overall. However, it is more time consuming to prepare and submit a PR for big changes. Also, not all changes are suitable to be contributed back which means the developer is at an impasse and we're back at the "internalised extensions and modifications" problem again.

In a perfect world, the entire core would be rewritten on PHP7 with modern coding techniques and such. I am truly astounded by how long I've been happily using the Symphony code base though. Perhaps it's "flexibility" is one of its strengths. 🤷‍♀️

symphony/lib/toolkit/class.widget.php
pointybeard: Yup, good point! I didn't review that array after I wrote it the first time.

froschdesign:
@pointybeard 
_Sorry for the very late response._

> In a perfect world, the entire core would be rewritten on PHP7 with modern coding techniques and such.

You forgot one very important point: the change of the entire PHP ecosystem and the available toolsets.
At the moment the codebase of the CMS is written by yourself and this without any tests like unit, integration or end-to-end.

So the question that arises with this case is: what is the task of the project? Creating a database abstraction layer or a XML library or another logger? I do not think so, because that already exists – complete with all tests!
The focus should be on the functions of the CMS and the benefit for the end user. This requires a solid base of tools, time to develop and implement ideas.

If wee look in Symphony's configuration file for Composer then we will see in the `master` branch:

https://github.com/symphonycms/symphony-2/blob/68f44f0c36ad3345068676bfb8a61c2e6a2e51f4/composer.json#L18-L20

and in the `3.0.x` branch:

https://github.com/symphonycms/symphony-2/blob/26a6ff33ae23d1787c61e5a42f02e0bb53fae495/composer.json#L37-L39

https://github.com/symphonycms/symphony-2/blob/26a6ff33ae23d1787c61e5a42f02e0bb53fae495/composer.json#L18-L22

The `require` section is _empty_. Compare now with a competitor of your choice.
(Then replace [JIT Image Manipulation](http://symphonyextensions.com/extensions/jit_image_manipulation/) with [Glide](https://glide.thephpleague.com/). 😄)

And this brings me back to my previous statement: this pull request goes in the wrong direction – _riding the same old horse_.

--- 

Btw. this is the right direction:

* https://github.com/symphonycms/symphony-2/tree/3.0.x/tests
* https://github.com/symphonycms/symphony-2/blob/3.0.x/tests/phpstan.neon (Yes, level 0 is the loosest but the usage of static analysis tool is a very good decision.)

@nitriques 👍

PR #2906, opened 2019-10-29
Added old behaviour for filtering system:id with 0 value

beaubbe:
Fixes #

or

Describe new feature here


nitriques:
Merged as aa811da6aa

